PR #8, opened 2018-01-24
[BPK-1266] Run gulp task on CI and fail if git status isn't porcelain

matteo-hertel:


.travis.yml
```diff
@@ -12,7 +12,9 @@ install:
   - npm install
 before_script: greenkeeper-lockfile-update
 script:
+  - set -e
   - npm run build
+  - (! git status --porcelain | grep -q .) || (echo "The generated values from `npm run gulp` are different from the values committed, please run `npm run build` and commit the artifact" && false)
```
matthewdavidson: I think that this script should be in a separate file because we'll want to use it before we release too. ()Plus it's cleaner from a travis file POV.

matteo-hertel:
👍 for moving it to a separate file, about greenkeeper we can explicitly exclude the package lock from the porcelain check, wdyt?

matthewdavidson:
@matteo-hertel yeah sure, how would that look? also do we need to do that in this PR or can it wait until the next greenkeeper update where we can fix then.

matteo-hertel:
I'll have to move the script anyway so I can see how easy is to implement

scripts/check-pristine-state.js
```diff
@@ -0,0 +1,44 @@
...
+} = require("util");
+const {
+    exec
+} = require('child_process');
+const execAsync = promisify(exec);
+
+const command = 'git status --porcelain | awk \'{print $2}\'';
+const [node, path, ...filesToIgnore] = process.argv;
+
+init();
```
matthewdavidson: Matteo I know your a big fan of hoisting but `eslint-config-skyscanner` doesn't allow it. When we integrate it (soon) these will have to change anyways.

.travis.yml
```diff
@@ -12,7 +12,9 @@ install:
   - npm install
 before_script: greenkeeper-lockfile-update
 script:
+  - set -e
```
matthewdavidson: this isn't required anymore is it?

scripts/check-pristine-state.js
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
```
matthewdavidson: you calling this with node, therefore the hashbang isn't required? Otherwise we should name it like a true cli tool `check-pristine-state` and use like so `./check-pristine-state package-lock.json`

scripts/check-pristine-state.js
matteo-hertel: 😢

.travis.yml
matteo-hertel: that is for travis not stopping the build as soon as it gets a non 0 exit code, I don't want the build to go through if one of the scripts fail, I'll try remove it and see if the problem is gone now

matteo-hertel: After testing with the set -e if the node scripts exits with 1 it'll stop the build there, otherwise itll go through to the gradle bit and the CI will fail, fail fast it's much better in this case

PR #11, opened 2018-02-02
[BPK-1261] Add eslint config to backpack-android

georgegillams:


.idea/modules.xml
```diff
@@ -5,6 +5,7 @@
       <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
       <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
       <module fileurl="file://$PROJECT_DIR$/backpack-android.iml" filepath="$PROJECT_DIR$/backpack-android.iml" />
+      <module fileurl="file://$PROJECT_DIR$/backpack-android.iml" filepath="$PROJECT_DIR$/backpack-android.iml" />
```
matteo-hertel: any idea where this is from? is not different from the one above

.prettierignore
```diff
@@ -0,0 +1 @@
+
```
matteo-hertel: is an empty file needed?

package.json
```diff
@@ -7,11 +7,19 @@
...
   "scripts": {
+    "precommit": "lint-staged",
     "preinstall": "node -e \"$(curl -fsSL https://raw.githubusercontent.com/Skyscanner/ensure-node-env/master/dist/index.js)\"",
-    "build": "gulp",
     "prerelease:env": ": ${BINTRAY_USER?Need to set BINTRAY_USER} ${BINTRAY_KEY?Need to set BINTRAY_KEY}",
     "prerelease": "npm run prerelease:env",
-    "release": "node release.js"
+    "release": "node release.js",
+    "test": "eslint . --ext .js,.jsx",
+    "build": "npm run test && gulp",
```
matteo-hertel: can we use npm's command hooks?
```
"prebuild": "npm run test",
"build": "gulp",
```

.idea/modules.xml
shaundon: Yeah, no idea! I'll just remove it.

.prettierignore
shaundon: Yeah fair enough, will remove!

package.json
shaundon: Sure - I didn't know about those

.idea/modules.xml
shaundon: Actually shouldn't `.idea` just be gitignored?

package.json
shaundon: Done

.prettierignore
shaundon: Done

.idea/modules.xml
matteo-hertel: I tought so too, but the default android app has just some subfolders in the gitignore, not really sure what the deal it

PR #16, opened 2018-02-07
[BPK-1244] Add support for Backpack text styles

k0nserv:
Adds the Backpack text styles

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -3,7 +3,10 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
+      <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
       <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
+      <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
+      <module fileurl="file://$PROJECT_DIR$/backpack-android.iml" filepath="$PROJECT_DIR$/backpack-android.iml" />
       <module fileurl="file://$PROJECT_DIR$/backpack-android.iml" filepath="$PROJECT_DIR$/backpack-android.iml" />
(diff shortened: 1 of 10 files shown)


k0nserv:
Any idea about this https://travis-ci.org/Skyscanner/backpack-android/builds/338560287?utm_source=github_status&utm_medium=notification @matteo-hertel? Per the docs `getComplexUnit` should be available since API level 22 and we seem to be using API level 27 on Travis

.idea/modules.xml
```diff
@@ -3,8 +3,11 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
+      <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
+      <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
```
matteo-hertel: these two are still a mistery to me, are the exact same as the others, why do they get added?

app/src/androidTest/java/net/skyscanner/backpack/BpkResources.kt
```diff
@@ -3,6 +3,8 @@ package net.skyscanner.backpack
 import android.content.res.Resources
 import android.support.test.InstrumentationRegistry
 import android.support.test.runner.AndroidJUnit4
+import android.util.Log
```
matteo-hertel: this is probaby not needed anymore

package.json
```diff
@@ -44,6 +44,7 @@
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-prettier": "^2.6.0",
     "eslint-plugin-react": "^7.6.1",
+    "eslint_d": "^5.2.2",
```
matteo-hertel: what this one does?

.idea/modules.xml
k0nserv: I have no idea

app/src/androidTest/java/net/skyscanner/backpack/BpkResources.kt
k0nserv: Yup

package.json
k0nserv: Make eslint work with my setup and ensures we all use the same version of eslint_d

PR #18, opened 2018-02-08
[BPK-1243] expose radii on Android

matteo-hertel:
I've also added a coming soon message to the main app as part of the PR

screenshot_1518101001

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -3,10 +3,7 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
-      <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
       <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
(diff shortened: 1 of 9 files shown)


.idea/modules.xml
```diff
@@ -3,11 +3,8 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
-      <module fileurl="file://$PROJECT_DIR$/Backpack/Backpack.iml" filepath="$PROJECT_DIR$/Backpack/Backpack.iml" />
```
matthewdavidson: this file should be excluded from source control, right?

matteo-hertel: There is a default gitignore when creating and android app, I think we're using that one and it's blacklisting a bunch of files from the `.idea` folder but not all of it
https://github.com/Skyscanner/backpack-android/blob/master/.gitignore#L39

matthewdavidson: Ah that would imply that some of them should be in source control ... 🤔

matteo-hertel: THat would be me guess but I might be completely wrong

matteo-hertel: Also this https://stackoverflow.com/questions/26336709/android-studio-should-the-entire-idea-directory-be-in-git-ignore

matthewdavidson: Wanna try removing it in a separate PR?

matteo-hertel: I'll check if our gitignore is different from a default android one and change it accordgly 👍

PR #33, opened 2018-08-15
Bpk 1808 port native code

amolgupta:


app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailActivity.kt
```diff
@@ -0,0 +1,65 @@
...
+ */
+class ComponentDetailActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_component_detail)
+        val toolbar = findViewById<View>(R.id.detail_toolbar) as Toolbar
+        setSupportActionBar(toolbar)
+        val image = findViewById<ImageView>(R.id.img)
+        Picasso.get().load(R.drawable.header).resize(1024, 800)
```
tiagoengel: I'm not sure about this header anymore. We want to take screenshots of these components to show in the docs site and this header will kinda get in the way.

amolgupta: Will remove this header and add toggle for RTL once we add support, screenshot tests and docs.

shaundon: How does an RTL toggle work - can you programmatically trigger that with an Android API? As far as I know on iOS we're just gonna have to set it manually in settings.app

amolgupta: This can be done programmatically.

shaundon: Good old Android, not being all locked down!

.gitignore
```diff
@@ -42,12 +43,15 @@ captures/
...
 .idea/tasks.xml
 .idea/gradle.xml
+.idea/modules.xml
 .idea/libraries
 .idea/assetWizardSettings.xml
 .idea/dictionaries/
 .idea/caches/
 .idea/codeStyles/
 .idea/copyright/
+.idea/caches/build_file_checksums.ser
```
matteo-hertel: this should be part of line 50, why is it not catching it? 🤔

Backpack/build.gradle
```diff
@@ -10,107 +9,93 @@ repositories {
   mavenCentral()
 }
 android {
-    compileSdkVersion 27
-    buildToolsVersion "27.0.3"
+  compileSdkVersion rootProject.ext.compileSdkVersion
```
matteo-hertel: 🎖

app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailActivity.kt
matteo-hertel: @shaundon join the dark side, we have RTL toggles

app/src/main/res/values/dimes.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <dimen name="item_width">200dp</dimen>
+  <dimen name="app_bar_height">200dp</dimen>
```
matteo-hertel: here's an interesting one, is it possible to do math with the tokens? like spacingXXL is 40dp can we do spacingXXL * 5?

app/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,11 @@
 <resources>
-  <string name="app_name">Backpack Android</string>
   <string name="main_button">Show me more!</string>
-
+  <string name="app_name">Backpack</string>
+  <string name="stub" translatable="false">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
```
matteo-hertel: while we're at it some proper copy can be added here, like Backpack is awesome etc

badge/src/main/java/net/skyscanner/backpack/badge/BpkBadge.kt
```diff
@@ -0,0 +1,142 @@
...
+      this.gravity = Gravity.CENTER
+      this.layoutParams = params
+    }
+
+  enum class Type constructor(internal var id: Int,
+                              internal var type: String,
+                              internal var bgColor: Int,
+                              internal var textColor: Int) {
+    /**
+     * Style for badges with positive messages
```
matteo-hertel: I think for consistency sake we should apply the same rules as the monorepo: comments should be avoided unless strictly necessary, if the code is not clear perhaps can be written in a more explicit way

It's kind of annoying but consistency is key

core/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
matteo-hertel: is this file needed?

badge/src/main/java/net/skyscanner/backpack/badge/BpkBadge.kt
amolgupta: These are javadoc comments. The developers can hover on the function/variable in the IDE to get help on the topic. It's a handy feature and in my opinion we should have it.

app/src/main/res/values/dimes.xml
amolgupta: Unfortunately no. https://stackoverflow.com/questions/8101178/how-do-i-apply-mathematical-operations-to-android-dimensions

Also, the ones here are from generated for the demo app. In actual components, the dimensions would be set in kotlin files and would follow the suggestion.

app/src/main/res/values/strings.xml
amolgupta: @shaundon can I borrow one of your poems here?

app/src/main/res/values/dimes.xml
matteo-hertel: worth a shot 😄

georgegillams:
Squash it first though! 🚀

PR #36, opened 2018-08-23
[No-JIRA] Publish task for modules

amolgupta:
Adding publishing logic for jitpack. The Backpack dependecies can now be included using 

`implementation 'com.github.Skyscanner:backpack-android:no-jira-publish-tasks-SNAPSHOT' `

After release 
`implementation 'com.github.Skyscanner:backpack-android:0.2.2' `

or as individual packages:
    
`implementation 'com.github.Skyscanner.backpack-android:badge:0.2.2' `

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -5,9 +5,6 @@
       "release": true
     }
   },
-  "src": {
-    "afterReleaseCommand": "./gradlew bintrayUpload"
-  },
   "dry-run": false,
   "npm": {
     "publish": false
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
   dependencies {
     classpath 'com.android.tools.build:gradle:3.1.3'
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-    classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
+    classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
     classpath "org.jetbrains.dokka:dokka-android-gradle-plugin:$dokka_version"
     // NOTE: Do not place your application dependencies here; they belong
     // in the individual module build.gradle files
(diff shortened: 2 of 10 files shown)


.release-it.json
```diff
@@ -5,9 +5,6 @@
       "release": true
     }
   },
-  "src": {
```
k0nserv: So what does this mean? Are we no longer using release-it? If we are shouldn't we still have a step here that performs the actual release?

amolgupta: Everything remains the same. we run the `release-it` task and it commits and tags the release, just does not upload to bintray.

k0nserv: So a human has to manually remember to do the upload after `release-it` is done?

amolgupta: No. There is no upload task anymore. Once the release commit is made,  jitpack will build and distribute. see https://jitpack.io/#Skyscanner/backpack-android

k0nserv: Okay so jitpack monitors our commits and builds/publishes when it sees a tagged commit on master?

amolgupta: On the first request to that particular tag/commit, it builds and caches the artifacts.

k0nserv: Hmm cool. Do we have something to ensure we have a changelog entry for each release?

amolgupta: Will need to setup Unreleased and changelog similar to the backpack monorepo.

build.gradle
```diff
@@ -32,7 +32,7 @@ task clean(type: Delete) {
   delete rootProject.buildDir
 }
 
-def BpkVersion = '0.2.1'
+def BpkVersion = '0.2.2'
```
matteo-hertel: is the version bump done on purpose?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "backpack-android",
-  "version": "0.2.1",
+  "version": "0.2.2",
```
matteo-hertel: same as above about the version bump

build.gradle
amolgupta: This was done by the `release-it.js` task. Will reverse it for now.

PR #39, opened 2018-08-24
RTL testing support

amolgupta:
This PR adds ability to test the components for RTL support. The storybook has a toggle switch to enable and disable RTL on the viewgroup that holds the sample components. This may change in the future as:

1. When screenshot tests are added, we might need to test for RTL and LTR

2. When we add documentation with screenshots, we might need to remove the toggle switch to something less visual.

Currently this is a testing support mechanism and it is upto the developer to visually ensure that component does not break for RTL languages.

device-2018-08-28-094536

device-2018-08-28-094552

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,8 +36,5 @@ dependencies {
   testImplementation 'junit:junit:4.12'
   androidTestImplementation 'com.android.support.test:runner:1.0.1'
   androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.1'
-  implementation ('com.squareup.picasso:picasso:2.71828'){
-    exclude group: 'com.android.support'
-  }
   implementation project(':Backpack')
 }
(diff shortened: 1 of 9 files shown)


matteo-hertel:
There is an argument about how/where we display the RTL toggle, the main concern are future screenshot testing, my humble opinion is to merge this, when we come down to implementing screenshots we'll worry about it, can I please have your opinions on this @tiagoengel @matthewdavidson @shaundon @k0nserv

tiagoengel:
I'm ok with it but I think taking screenshots (for docs, not tests) is something we will have to do very soon. So I'm fine with merging but we need to keep in mind that we will need an option to take a screenshot with a "empty" screen.

k0nserv:
Can't we have a dev menu trigger via shake for this kind of stuff? I bet there will be more stuff like this

matthewdavidson:
My feeling is YAGNI. No point in having code that we _might_ need. Plus those screenshots above don;t look great ... the badges are cut off etc (not sure if thats related to the RTL toggle or not).

amolgupta:
On popular opinion, I will close this PR for now.

As per the discussion with @matteo-hertel , I will be adding a product flavor for RTL testing. The screenshots and docs will not have any navbar or toggle. If the developer wants to test RTL, they can use the new build variant from android studio.
Will keep this branch for working on the enhancement.

PR #40, opened 2018-08-26
[BPK-1845] add text component

matteo-hertel:
This adds the TextComponent

- [x] Add the component 
- [x] Update Demo app 
- [x] Add tests 

here's what it looks like 

screenshot_1535303416

I'm happy with everything but the tests, I spent way too much time trying to figure out how to get the font family from a text view and I failed miserably

### DIFF
--- a/text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
+++ b/text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
@@ -1,0 +1,82 @@
+package net.skyscanner.backpack.text
+
+import android.content.Context
+import android.content.res.TypedArray
+import android.support.annotation.IntDef
+import android.support.v4.widget.TextViewCompat
+import android.support.v7.widget.AppCompatTextView
+import android.util.AttributeSet
+
+
+open class BpkText(
+  context: Context,
+  attrs: AttributeSet?,
+  defStyleAttr: Int) : AppCompatTextView(context, attrs, defStyleAttr) {
+
+  @IntDef(XS.toLong(), SM.toLong(), BASE.toLong(), LG.toLong(), XL.toLong(), XXL.toLong())
+
+  annotation class Styles
+
+  companion object {
+    const val XS = 0
+    const val SM = 1
+    const val BASE = 2
+    const val LG = 3
+    const val XL = 4
+    const val XXL = 5
+  }
+
+  @Styles
+  var textStyle: Int = BASE
+    set(value) {
+      field = value
+      this.setup()
+    }
+
+  var emphasize: Boolean = false
+    set(value) {
+      field = value
+      this.setup()
+    }
+
+  private val styleMapping = mapOf(
+    XS to Pair(R.style.bpkTextXsEmphasized, R.style.bpkTextXs),
+    SM to Pair(R.style.bpkTextSmEmphasized, R.style.bpkTextSm),
+    BASE to Pair(R.style.bpkTextBaseEmphasized, R.style.bpkTextBase),
+    LG to Pair(R.style.bpkTextLgEmphasized, R.style.bpkTextLg),
+    XL to Pair(R.style.bpkTextXlEmphasized, R.style.bpkTextXl),
+    XXL to Pair(R.style.bpkTextXxlEmphasized, R.style.bpkTextXxl)
+  )
+
+  constructor(context: Context) : this(context, null)
+  constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, R.style.bpkTextBase)
+
+  init {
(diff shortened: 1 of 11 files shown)


shaundon:
Probably worth aligning on what the example text says, I think the iOS one might say something different. I like yours so I reckon the iOS one should change (I can do that in another PR).

text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -0,0 +1,60 @@
...
+    setup()
+  }
+
+  private fun initialize(context: Context, attrs: AttributeSet?, defStyleAttr: Int) {
+
+    val a: TypedArray = context.theme.obtainStyledAttributes(
+      attrs,
+      R.styleable.text,
+      0, 0)
+
```
amolgupta: need to add null checks here

text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -0,0 +1,60 @@
...
+  companion object {
+    val XS = 0
+    val SM = 1
+    val BASE = 2
+    val LG = 3
+    val XL = 4
+    val XXL = 5
+  }
+
+  var textStyle: Int = BASE;
```
tiagoengel: Since those properties are public and mutable we need to react to updates here (possibly by calling setup again?). e.g: 
```
var textStyle: Int = BASE;
  set(value) {
     field = value
     setup()
  }
```

k0nserv: Shouldn't this be an enum type instead of just `Int`? It seems with this solution it's valid to do something like `text.textStyle = 287319` which is nonsensical. With an enum the compiler would prevent that right?

amolgupta: There is a performance oveahead in android for enums and it is recommended to use constants for such cases. This pattern can be found everywhere in android codebase. The concern of having wrong value set to the type is addressed using the annotation `@IntDef`. For example see the linear layout implementation here https://github.com/aosp-mirror/platform_frameworks_base/blob/faac4ad153f55b1d224e2e800c24d0b9eb0ae0e5/core/java/android/widget/LinearLayout.java#L91-L96

amolgupta:
are we supporting auto-resizing text? https://developer.android.com/guide/topics/ui/look-and-feel/autosizing-textview

k0nserv:
> are we supporting auto-resizing text? https://developer.android.com/guide/topics/ui/look-and-feel/autosizing-textview

Yes we should

text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -0,0 +1,60 @@
...
+      0, 0)
+
+    textStyle = a.getInt(R.styleable.text_textStyle, BASE);
+    emphasize = a.getBoolean(R.styleable.text_emphasize, false);
+
+    a.recycle()
+  }
+
+  private fun setup() {
+    when (textStyle) {
```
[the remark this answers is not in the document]
matteo-hertel: Neat, I like it, thanks!

text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
matteo-hertel: I've implemented that under Amol's guidance so I'm sure is the best approach 😄

matteo-hertel:
@amolgupta about `auto-resizing` given this is a wrapper of TextView, the consumer can pass `android:autoSizeTextType="uniform"` to it an it'll be applied correct? If I understood correctly there is no need for us to explicitly add it in

text/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    textStyle = a.getInt(R.styleable.text_textStyle, BASE)
+    emphasize = a.getBoolean(R.styleable.text_emphasize, false)
+
+    a.recycle()
+  }
+
+  private fun setup() {
+
+    val styleProps = styleMapping[textStyle]
```
tiagoengel: As far as I can tell this will never be null because the annotation above is constraining the types that can be set to `textStyle` but I still think we should probably use `getOrElse` here instead of `!!` bellow. e.g. `styleMapping.getOrElse(textStyle, BASE)`

matteo-hertel: the compiler complains if you don't check for null case, but as you said it will never be null

I don't mind change it, but is it worth it?

PR #44, opened 2018-08-28
[nojira] Upgrade `eslint-config-skyscanner-with-prettier`

matthewdavidson:


package.json
```diff
@@ -40,10 +40,9 @@
   },
   "homepage": "https://github.com/Skyscanner/backpack-android#readme",
   "devDependencies": {
-    "babel-eslint": "^8.2.1",
```
matthewdavidson: This dependency  is encapsulated inside `eslint-config-skyscanner-with-prettier`

PR #46, opened 2018-08-29
No jira general cleanup

amolgupta:
This PR is for lint issues

1. Fixed the ones based on lint suggestions. (Remove unused resources mainly)
2. Disabled a few lint checks with description.


panel/src/main/java/net/skyscanner/backpack/panel/BpkPanel.kt
```diff
@@ -2,18 +2,19 @@ package net.skyscanner.backpack.panel
...
 import android.content.Context
 import android.support.annotation.Dimension
+import android.support.annotation.Nullable
 import android.support.v7.widget.LinearLayoutCompat
 import android.util.AttributeSet
 import android.widget.LinearLayout
 
 
-class BpkPanel(
+open class BpkPanel(
```
matteo-hertel: open might be needed in the text component as well

amolgupta: It was missing only here :)

PR #48, opened 2018-08-30
[BPK-1836] Config API docs

tiagoengel:
It looks like this:
screen shot 2018-08-30 at 16 02 46

The little `source` link at the top links to the source code on github \o/

It can be generated locally with `npm run docs` and viewed with `chrome build/docs/backpack-android/index.html`

To configure the outputFolder `npm run docs -- -PdokkaOutput=/path/to/folder`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,12 @@ Once the component it generated, add it to the `settings.gradle` file to start u
 
 Run `npm run release` and follow the process through, you'll be asked which semantic version the release.Once released verify the artifacts on [`jitpack`][3]
 
+## Docs
+
+Run `npm run docs` to genere API docs. Docs will be generated in the `build/docs` folder by default. To provide a different output folder use `npm run docs -- -PdokkaOutput=/path/to/folder`
+
+Run `npm run docs:clean` to remove generated docs.
+
 [1]: https://github.com/creationix/nvm
 [2]: https://github.com/audreyr/cookiecutter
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -29,6 +29,12 @@ Once the component it generated, add it to the `settings.gradle` file to start u
 
 Run `npm run release` and follow the process through, you'll be asked which semantic version the release.Once released verify the artifacts on [`jitpack`][3]
 
+## Docs
+
+Run `npm docs` to genere API docs. Docs will be generated in the `build/docs` folder by default. To provide a different output folder use `npm run docs -- -PdokkaOutput=/path/to/folder`
```
matthewdavidson: should be `npm run docs`

CONTRIBUTING.md
```diff
@@ -29,6 +29,12 @@ Once the component it generated, add it to the `settings.gradle` file to start u
...
+
+Run `npm docs:clean` to remove generated docs.
```
matthewdavidson: should be `npm run docs:clean`

k0nserv:
> Side note: we could, for now, push it to the gh-pages branch as part of the build so docs even if not pretty, are available if people start using Backpack, wdyr?

That's a pretty neat idea Matteo 👍

tiagoengel:
That's a great idea, but I guess we will be also adding it to the doc site build next too, so may be better to do just that?

PR #50, opened 2018-08-31
[BPK-1827] add readmes to components

matteo-hertel:
This adds readmes to the different components and improves consistency with the backpack monorepo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
-# Change Log
+# Backpack Android changelog
 
-All notable changes to this project will be documented in this file.
+[Unreleased changes](./UNRELEASED.md).
 
-## Master
+## 0.2.1
 
-_Nothing Yet_
+Introducing BpkBadge and BpkPanel component.
 
 ## 0.2.0
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-# Backpack
+# Backpack Android
+
+> Backpack is a collection of design resources, reusable components and guidelines for creating Skyscanner's products.
 
 [![CI Status](http://img.shields.io/travis/Skyscanner/backpack-android.svg?style=flat)](https://travis-ci.org/Skyscanner/backpack-android)
 [![Greenkeeper badge](https://badges.greenkeeper.io/Skyscanner/backpack-android.svg)](https://greenkeeper.io/)
@@ -14,19 +16,19 @@ all of it, add the following line to your `build.gradle` (in your app module) in
 ```gradle
 implementation 'com.github.Skyscanner:backpack-android:0.2.1'
 ```
-To install a single component, add the following line to your `build.gradle` (in your app module) in the `dependencies` block:
 
-```gradle
-implementation 'com.github.Skyscanner.backpack-android:badge:0.2.1'
-```
-
-Backpack is available in Jitpack so if your app resolves dependencies through Jitpack you're all set, if not add in your root `build.gradle`
+If your app resolves dependencies through Jitpack you're all set, if not add in your root `build.gradle`
 
 ```gradle
 repositories {
     maven { url 'https://jitpack.io' }
 }
 ```
+## Components
+* [Badge][docs/Badge/README.md]
+* [Panel][docs/Panel/README.md]
+* [Text][docs/Text/README.md]
+
 ## Usage
 
 ### Radii
--- a/decisions/README.md
+++ b/decisions/README.md
@@ -1,0 +1,8 @@
+# Backpack decisions
+
+This directory is a collection of decisions that the squad has made on various topics.
+
+## Decisions
+
+<!-- Please keep this in alphabetical order. -->
+
--- a/decisions/TEMPLATE.md
+++ b/decisions/TEMPLATE.md
@@ -1,0 +1,16 @@
+<!-- Copy this to make your own decisions. -->
+
+# Title <!-- e.g. "Squad colour coordination" -->
+
+## TL;DR
+<!-- Only if long enough to merit one. -->
+
+## Decision
+<!-- e.g. "On Wednesdays we wear pink" -->
+
+## Thinking
+<!-- e.g. "Because a study showed that pink improves coding ability on Wednesdays" -->
+
+## Anything else
+<!-- Only if necessary. -->
+
--- a/docs/Badge/README.md
+++ b/docs/Badge/README.md
@@ -1,0 +1,31 @@
+# Badge
+
+## Installation
+
+Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/backpack-android). Check the main [Readme](../../README.md#installation) for a complete installation guide.
+
+## Usage
+
+The Badge component can be used in both XML and Kotlin
+
+Example of a success type badge in XML
+
+```xml
+<net.skyscanner.backpack.badge.BpkBadge
+  android:layout_width="wrap_content"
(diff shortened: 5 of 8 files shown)


panel/README.md
```diff
@@ -0,0 +1,29 @@
+# Panel
+
```
tiagoengel: Can we add installation instructions here?

panel/README.md
```diff
@@ -0,0 +1,29 @@
...
+The Panel component can be used in both XML and Kotlin
+
+Example of a padded panel in XML
+
+```xml
+<net.skyscanner.backpack.panel.bpkpanel
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    app:bpk_padding="true">
```
tiagoengel: is it still `bpk_padding` or just `padding`?

panel/README.md
```diff
@@ -0,0 +1,29 @@
...
+        android:text="panel message" />
+</net.skyscanner.backpack.panel.bpkpanel>
+```
+
+Example of a padded panel in Kotlin
+
+```Kotlin
+import net.skyscanner.backpack.panel
+
+val panel = BpkPanel(context)
```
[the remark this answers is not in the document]
matteo-hertel: I don't have anything against it, I just don't know why should we do one or the other, is it best practice or something like that?

decisions/TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+# Backpack decisions
```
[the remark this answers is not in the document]
matteo-hertel: To drive consistency with the monorepo and the iOS codebase if https://github.com/Skyscanner/backpack-ios/pull/41 get merged

I think the usage of `apply` could be a good candidate for a first entry

panel/README.md
matteo-hertel: The original idea was to put it there but there is a PR to restructure the codebase that will make it redundant so I purposefully left it out

good catch though!

panel/README.md
matteo-hertel: 🤔 if memory serves me well I stole the snippet from the demo app, I'll double check, thanks!

CHANGELOG.md
```diff
@@ -1,10 +1,9 @@
...
+# Backpack Android changelog
 
-All notable changes to this project will be documented in this file.
+[Unreleased changes](./UNRELEASED.md).
 
-## Master
-
-_Nothing Yet_
+## 0.2.1
+Test Release
```
matthewdavidson: What is dis?

decisions/TEMPLATE.md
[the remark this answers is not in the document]
tiagoengel: I mean the template, because the decision `README` looks exactly the same 🤷‍♂️

panel/README.md
tiagoengel: I remember it was like this initially because we were not sure if just `padding` would clash with something. But the last thing I remember is that we were not using the bpk prefix anymore. @amolgupta would know better

panel/README.md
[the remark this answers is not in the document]
tiagoengel: It has the same effect, but since this is an `initialisation` block it looks much nicer to have it inside the apply block. Plus avoid the rewriting the panel variable, so you could just add

```
BpkPanel(context).apply {
   padding = true
}
```

This is actually the only instance where apply is explicitly preferred in Kotlin. E.g,  when you have a function that returns a new object, you don't need to declare a new variable for this.

decisions/TEMPLATE.md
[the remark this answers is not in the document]
matteo-hertel: 😱 copy-pasta, thanks for spotting it 😅

CHANGELOG.md
matteo-hertel: I think the description is missing and that should be a placeholder, I'll update it shortly 😅

panel/README.md
amolgupta: yes. We need to drop the `bpk_` prefix. A PR after moving things to single modules would be good.

matteo-hertel: so just to be sure, at the minute the `bpk` is required, correct?

amolgupta: I assume this PR would be merge only after moving to single module. A git rebase and rename variable would do the job.

docs/BpkText/README.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+Example of a text view with large font in Kotlin
+
+```Kotlin
+import net.skyscanner.backpack.text
+
+BpkText(context).apply {
+   text = 'Flights to Edinburgh',
+   textStyle='xxl'
```
tiagoengel: nit: space between the equals sign

matteo-hertel:
All done @tiagoengel

README.md
```diff
@@ -27,6 +29,11 @@ repositories {
     maven { url 'https://jitpack.io' }
 }
 ```
+## Components
+* [Badge][docs/BpkBadge/README.md]
```
tiagoengel: still BpkBadge 🙈

tiagoengel:
You've ignored the `installation` section question though? 😄

docs/Text/README.md
```diff
@@ -0,0 +1,26 @@
...
+  app:textStyle="xxl" />
+```
+
+Example of a text view with large font in Kotlin
+
+```Kotlin
+import net.skyscanner.backpack.text
+
+BpkText(context).apply {
+   text = 'Flights to Edinburgh',
```
tiagoengel: sorry, I didn't see this earlier but there shouldn't be a comma here

docs/Badge/README.md
```diff
@@ -0,0 +1,27 @@
...
+  app:type="success"/>
+```
+
+Example of a success type badge in Kotlin
+
+```Kotlin
+import net.skyscanner.backpack.panel
+
+BpkBadge(context).apply {
+   message = 'Message',
```
tiagoengel: sorry, I didn't see this earlier but there shouldn't be a comma here

matteo-hertel:
@tiagoengel as per the installation, there is no installation step after Amol's changes are going throug, what would you suggest we put in?

tiagoengel:
By installation I mean something like explaining how to add jitpack and to add the gradle dependency. I know it will be only one dependency for all components but someone looking at a component for the first time on our docs site needs to see that. Even if it is the same installation process for every component (and not required if they have it installed already) I still think it is important to have it there. It can even be a link to the main readme where we have the installation process.

matteo-hertel:
I've added a link to the main readme, this should be good to go now

CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
...
+# Backpack Android changelog
 
-All notable changes to this project will be documented in this file.
+[Unreleased changes](./UNRELEASED.md).
 
-## Master
+## 0.2.1
 
-_Nothing Yet_
+Introducing BpkBadge and BkpPanel component.
```
shaundon: `BkpPanel` is misspelled

decisions/README.md
```diff
@@ -0,0 +1,8 @@
+# Backpack decisions
+
+This directory is a collections of decisions that the squad has made on various topics.
```
shaundon: `collections` should be `collection`

CHANGELOG.md
matteo-hertel: No, Bkapack is the new name of the design system 😄

PR #51, opened 2018-08-31
Move files from modules to packages

amolgupta:
This PR changes the project structure from multi module to single module. Reason for doing this change:

- CI timings increased by 3-5 minutes on adding each component
- There was no requirement for adding individual components as dependencies for projects.

This PR  
- moves files from modules to their corresponding packages 
- fixes the imports.
- changes build settings for CI to accommodate the change


georgegillams:
The size of this lad....

And wow - 10 mins off the build time already! Nice work

img_3195

Backpack/src/main/res/values/attrs.xml
```diff
@@ -15,4 +15,18 @@
...
+      <enum name="sm" value="1" />
+      <enum name="base" value="2" />
+      <enum name="lg" value="3" />
+      <enum name="xl" value="4" />
+      <enum name="xxl" value="5" />
+    </attr>
+    <attr name="emphasize" format="boolean" />
+  </declare-styleable>
+  <declare-styleable name="panel">
+    <attr name="bpk_padding" format="boolean" />
```
matteo-hertel: remember to follow up with a PR to remove this

tiagoengel:
Are api still working? Now that we have only one project we probably don't need all that code to generate it, the plugin should work with the default config.

PR #52, opened 2018-09-03
testing more component

matteo-hertel:
Testing CI


matteo-hertel:
The single module approach is more performant than the current one

PR #64, opened 2018-09-11
[No-JIRA] Update node

amolgupta:
Update to latest node version 
read https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#2018-09-11-version-8120-carbon-lts-mylesborins


k0nserv:
We should update the node version to the LTS release that bundles that npm version.  See https://github.com/Skyscanner/backpack/pull/1048

PR #70, opened 2018-09-19
Remove header image and Image loading library

amolgupta:
Removing the header image and collapsing-toolbar from stories

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailFragment.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailFragment.kt
@@ -1,9 +1,7 @@
 package net.skyscanner.backpack.demo
 
 import android.os.Bundle
-import android.support.design.widget.CollapsingToolbarLayout
 import android.support.v4.app.Fragment
-import net.skyscanner.backpack.demo.data.ComponentRegistry
 
 /**
  * A fragment representing a single Component detail screen.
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailFragment.kt
```diff
@@ -17,25 +15,21 @@ import net.skyscanner.backpack.demo.data.ComponentRegistry
...
-        if (arguments!!.containsKey(ARG_ITEM_ID)) {
-
-            val activity = this.activity
-            val appBarLayout = activity!!.findViewById<CollapsingToolbarLayout>(R.id.toolbar_layout)
-            if (appBarLayout != null) {
-                appBarLayout.title = arguments!!.getString(ARG_ITEM_ID)
-            }
-        }
+  override fun onCreate(savedInstanceState: Bundle?) {
+    if (arguments!!.containsKey(ARG_ITEM_ID)) {
```
tiagoengel: can we change this to `arguments?.containsKey`?

app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailFragment.kt
```diff
@@ -17,25 +15,21 @@ import net.skyscanner.backpack.demo.data.ComponentRegistry
...
+      val activity = this.activity as ComponentDetailActivity
+      activity.title = arguments!!.get(ARG_ITEM_ID) as CharSequence?
```
tiagoengel: why is the cast to CharSequence necessary? doesn't it have a `getString/Text` function?

PR #71, opened 2018-09-19
Bpk 1935 icon docs

amolgupta:
Adding stories for icons 

screenshot_1537353354

screenshot_1537353361


georgegillams:
Looking really really good.
Can we just make the "black" icons be `bpk-gray-900` like they are for [RN](https://backpack.github.io/components/icon/?platform=native)?

We don't like [vantablack](https://www.surreynanosystems.com/vantablack)!

amolgupta:
@georgegillams I am adding "All icons" as default without any color modification. The color logic is in the main "Icons" story. Icons can be colorized  by adding a tint color, keeping them black ensures that uses would not have to get into any color code maths while setting the tint.

georgegillams:
@amolgupta that makes sense, but could we add `android:tint="@color/bpkGray900"` to the small and large icons shown in the story - just so that we aren't showcasing them being used in a non-backpack color.
I'm not suggesting we should change the color of the resources themselves

amolgupta:
@georgegillams Cool. Will add them to the "icons" story but not the "All Icons" story :)

PR #72, opened 2018-09-20
Bpkbutton

antalindisguise:
image

# How to use

```xml 
    <net.skyscanner.backpack.button.BpkButton
      android:text="With icon"
      android:layout_marginStart="@dimen/bpkSpacingMd"
      app:buttonIconStart="@drawable/bpk_weather"
      app:buttonType="primary"/>
```

# Gotchas

- Needed to rename the BpkBadge generic attribute name `type` to `badgeType` and use `buttonType` otherwise there would be a namespace clash under `app:type`
- I'm not sure how the pressed state should look like - I used what we do in the Skyscanner app at the moment

=======

Might need some minor adjustments with the text baseline when icon is used but I'm pushing this to get some feedback on it.


tevelee:
Random observation: when you have image and text as well, the text doesn't seem like vertically centered. 
On the selected state, you can try them out here: https://backpack.github.io/components/button/?platform=web (and retrieve the properties from css)

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -0,0 +1,282 @@
...
+        for (f in values()) {
+          if (f.id == id) return f
+        }
+        throw IllegalArgumentException()
+      }
+    }
+  }
+
+  enum class Size(internal var id: Int) {
+    Basic(0),
```
tevelee: on iOS we have 'Default' and 'Large'. Not sure which one to go with

antalindisguise: `default` as an attribute in XML is not allowed by Android, therefore I chose another word for this. I wrote this in the Gotchas section of the PR. 

image

PR #73, opened 2018-09-21
[no-jira] Icons bugfix for API 19

amolgupta:
Fixes a crash on API-19 while loading vector icons. This fix enables using vector resources using compatibility library.
see https://stackoverflow.com/a/42956161

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/BackpackDemoApplication.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/BackpackDemoApplication.kt
@@ -1,0 +1,14 @@
+package net.skyscanner.backpack.demo
+
+import android.support.multidex.MultiDexApplication
+import android.support.v7.app.AppCompatDelegate
+
+/**
+ * Application class registered in AndroidManifest.xml
+ */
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/skyscanner/backpack/demo/BackpackDemoApplication.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.support.multidex.MultiDexApplication
+import android.support.v7.app.AppCompatDelegate
+
+/**
+ * Application class registered in AndroidManifest.xml
+ */
+class BackpackDemoApplication : MultiDexApplication() {
+  override fun onCreate() {
+    super.onCreate()
+    AppCompatDelegate.setCompatVectorFromResourcesEnabled(true)
```
tiagoengel: wasn't this `vectorDrawables.useSupportLibrary = true` supposed to make it work with the api 19?

amolgupta: Surprisingly its both. Adding SO link to PR for reference.

PR #75, opened 2018-09-24
[BPK-1838] deploy docs on tagged builds

georgegillams:


scripts/build-docs-ci
```diff
@@ -0,0 +1,102 @@
...
+} = require("util");
+const {
+    exec
+} = require('child_process');
+const execAsync = promisify(exec);
+
+const [node, path, releaseTag, ...rest] = process.argv;
+
+const buildDocs = () => {
+  fs.mkdirSync(`./dist`, function(err) {
```
tiagoengel: Indentation is still inconsistent here. Also, considering the error handling is exactly the same everywhere in this file can we create a function and use it? Something like this

```javascript
const exitOnError = err => {
  if (err) {
    console.error(err):
    process.exit(1);
  }
})

fs.mkdirSync(`./dist`, exitOnError);
```

georgegillams: Good point!

PR #78, opened 2018-09-26
[BPK-1949] Update API for button

amolgupta:
Updates the api to use 
`icon` & `iconPosition` 
instead of `iconStart` and `iconEnd`


Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -25,8 +26,26 @@ open class BpkButton @JvmOverloads constructor(
...
 ) : AppCompatButton(context, attrs, defStyleAttr) {
 
+  @IntDef(START, END, ICON_ONLY)
+
+  annotation class IconPosition
+
+  companion object {
+    const val START = 0
+    const val END = 1
+    const val ICON_ONLY = 2
```
k0nserv: Why do we need this? Can't we just ignore `START` or `END` if a text isn't set? That's what we do on iOS

amolgupta: There were a few problems with that:
1. Its possible to set both start and end at the same time with no priority to either.
2. In case of icon only, the padding values are different to make it look round. It was depending on checking the text length. With 'iconOnly` any extra text is force removed

k0nserv: How can you set start and end at the same time?

amolgupta: It was possible by setting both `iconStart` and `iconEnd` properties for the same button. With this PR it would not be allowed anymore. This also aligns with the button api on RN

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -25,8 +26,26 @@ open class BpkButton @JvmOverloads constructor(
...
+  var iconPosition: Int = BpkButton.END
+    set(value) {
+      field = value
+      this.setup()
+    }
+
+
   private val roundedButtonCorner by lazy(LazyThreadSafetyMode.NONE) { dpToPx(24, context).toFloat() }
-  private val strokeWidth by lazy(LazyThreadSafetyMode.NONE) { 2 }
+  private val strokeWidth by lazy(LazyThreadSafetyMode.NONE) { 4 }
```
tiagoengel: Are we 100% sure this will never be shared by more than one thread? I'm not sure about how android works internally but if there's any chance this will be shared by more than one thread we shouldn't do this. Also, if we do I think we should add a decision about this, so every lazy field inside a component uses no synchroniser.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -91,43 +97,36 @@ open class BpkButton @JvmOverloads constructor(
 
   @Suppress("DEPRECATION")
```
tiagoengel: didn't see this before but I don't think we should suppress any deprecation, if we need it then let's keep the warning there so are aware of it.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -91,43 +97,36 @@ open class BpkButton @JvmOverloads constructor(
...
+    }
 
+    // Text is 12dp and icon is 16dp. if icon is present,
+    // padding needs to be reduced by 2 dp on both sides
     this.setPadding(
-      dpToPx(12, context),
-      dpToPx(if (this.iconStart != null || this.iconEnd != null) 8 else 12, context),
-      dpToPx(12, context),
-      dpToPx(if (this.iconStart != null || this.iconEnd != null) 8 else 12, context)
+      dpToPx(if (iconPosition == ICON_ONLY) 10 else 12, context),
```
tiagoengel: shouldn't we be using tokens here?

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
antalindisguise: The default lazy initialization is synchronized which may cause accidental deadlocks. None uses no locks, and therefore it is faster. As we are simply returning primitive values and no complex object it is completely safe and faster to use this than the synchronized one.

tiagoengel: I understand what you mean but the whole point of lazy is to delay heavy operations and make sure you do them once (because they are heavy). To be honest looking at it again there's no point in having 4 as a lazy value... Maybe we should just remove from roundedButtonCorner as well, as it's not an expensive operation. 

I'm still not sure about the deadlock part as it is supposed to be safe even with the lock, the only diference is that without the lock it _may_  be initialised twice in case this instance is shared by two (or more) threads. 

I think we should just remove the lazy initialisation in this case.

antalindisguise: I can partially agree with what you say, but if we use tokens here (or at least on roundedButtonCorner we should keep them - as you don't want to call and load values from resources until they are not used/called). On the ones where we have simple values we can remove lazy.

PR #79, opened 2018-09-27
[BpkButton] should be possible to set isEnabled from code

antalindisguise:
**Issue:** with the current API if you set isEnabled from code (f.x. disabling the button upon certain condition in code, and not from XML) then it had no effect

**Cause:** `setEnabled` is called on the AppCompatButton API and the `setup()` is not called.

**Solution:** Overriding the `setEnabled` API call and triggering a `setup()` call after the state change (similar to how we do it after any custom property change - icon, type)

button-isenabled-fix

I also added a hidden click listener on the icon button on each story so we can flip the state of the disabled button to show this functionality. *If you prefer, it can be removed.*

### DIFF
--- a/Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
+++ b/Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
@@ -58,13 +58,13 @@ class BpkButtonTest {
       type = BpkButton.Type.Primary
     }
     val newState = false
-    val disabledBackgroundColor = R.color.bpkGray100
+    val expectedBackgroundState = subjectUnderTest.disabledBackground
 
     // When
     subjectUnderTest.isEnabled = newState
 
     // Then
     Assert.assertEquals(subjectUnderTest.isEnabled, newState)
-    Assert.assertEquals(subjectUnderTest.drawingCacheBackgroundColor, disabledBackgroundColor)
+    Assert.assertEquals(subjectUnderTest.background, expectedBackgroundState)
   }
 }
--- a/Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
+++ b/Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
@@ -4,9 +4,6 @@ import android.content.Context
 import android.support.test.InstrumentationRegistry
 import android.support.test.runner.AndroidJUnit4
 import android.support.v7.content.res.AppCompatResources
-import com.nhaarman.mockito_kotlin.atLeastOnce
-import com.nhaarman.mockito_kotlin.spy
-import com.nhaarman.mockito_kotlin.verify
 import net.skyscanner.backpack.R
 import org.junit.Assert
 import org.junit.Before
@@ -67,7 +64,7 @@ class BpkButtonTest {
     subjectUnderTest.isEnabled = newState
 
     // Then
-    assert(subjectUnderTest.isEnabled == newState)
-    assert(subjectUnderTest.drawingCacheBackgroundColor == disabledBackgroundColor)
+    Assert.assertEquals(subjectUnderTest.isEnabled, newState)
+    Assert.assertEquals(subjectUnderTest.drawingCacheBackgroundColor, disabledBackgroundColor)
   }
 }
--- a/Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
+++ b/Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
@@ -51,20 +51,16 @@ class BpkButtonTest {
   }
 
   @Test
-  fun givenBpkButtonInAnyTypeAndFormWhenIsEnabledIsSetThenSetupIsCalled() {
-    // Given
-    val subjectUnderTest = BpkButton(context).apply {
+  fun test_enabled_state() {
+    val button = BpkButton(context).apply {
       isEnabled = true
       type = BpkButton.Type.Primary
     }
     val newState = false
-    val expectedBackgroundState = subjectUnderTest.disabledBackground
+    val expectedBackgroundState = button.disabledBackground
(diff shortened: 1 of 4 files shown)


Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -24,23 +35,39 @@ class BpkButtonTest {
...
     val button = BpkButton(context).apply {
       icon = trainIcon
       iconPosition = BpkButton.START
     }
-    Assert.assertEquals(button.compoundDrawables[0],trainIcon)
+    Assert.assertEquals(button.compoundDrawables[0], trainIcon)
+  }
+
+  @Test
+  fun givenBpkButtonInAnyTypeAndFormWhenIsEnabledIsSetThenSetupIsCalled() {
```
tiagoengel: can we change this to normal unit testing? We're not using BDD here.

Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -24,23 +35,39 @@ class BpkButtonTest {
...
+      type = BpkButton.Type.Primary
+    }
+    val newState = false
+    val disabledBackgroundColor = R.color.bpkGray100
+
+    // When
+    subjectUnderTest.isEnabled = newState
+
+    // Then
+    assert(subjectUnderTest.isEnabled == newState)
```
tiagoengel: Does this give a readable error? I suspect `Assert.assertEquals` will give an error with more information

Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -1,19 +1,30 @@
 package net.skyscanner.backpack.button
 
+import android.content.Context
 import android.support.test.InstrumentationRegistry
 import android.support.test.runner.AndroidJUnit4
 import android.support.v7.content.res.AppCompatResources
+import com.nhaarman.mockito_kotlin.atLeastOnce
```
tiagoengel: I think those are not required right? CI is failing because of it

Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
antalindisguise: context is needed if you want to inflate the view. Tried mocking it but Android framework doesn't allow for that.

tiagoengel:
Apparently CI is failing on the last test you added @antalindisguise

Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
antalindisguise: I follow the convention we use generally in Skyscanner for Android. I don't see any real value in the naming convention used above in the other tests but feel free to change my test name to whatever it would suit this project.

amolgupta: You may rename it to 'test_enabled_state'. Since this PR is from a fork, I won't be able to change it.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -12,6 +12,7 @@ import android.os.Build
 import android.support.annotation.ColorInt
 import android.support.annotation.ColorRes
 import android.support.annotation.IntDef
+import android.support.annotation.VisibleForTesting
```
amolgupta: Unused import?

app/src/main/res/layout/fragment_button.xml
```diff
@@ -18,7 +21,7 @@
     android:layout_margin="@dimen/bpkSpacingSm"
     android:layout_width="wrap_content"
     android:id="@+id/btn_disabled"
-    android:layout_toRightOf="@id/btn_default"
+    android:layout_toEndOf="@id/btn_default"
```
amolgupta: nice one

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -112,22 +143,19 @@ open class BpkButton @JvmOverloads constructor(
...
       compoundDrawablePadding = paddingWithIcon / 2
     }
 
-    this.background = getSelectorDrawable(
-      normalColor = ContextCompat.getColor(context, if (this.isEnabled) type.bgColor else R.color.bpkGray100),
-      cornerRadius = roundedButtonCorner,
-      strokeWidth = if (this.isEnabled) strokeWidth else 0,
-      strokeColor = if (this.isEnabled) ContextCompat.getColor(context,type.strokeColor) else ContextCompat.getColor(context,android.R.color.transparent)
-    )
+    this.background = if (this.isEnabled) enabledBackground else disabledBackground
```
amolgupta: this refactoring has caused a snapshot test to fail. please verify

Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -24,23 +32,39 @@ class BpkButtonTest {
...
       iconPosition = BpkButton.START
     }
-    Assert.assertEquals(button.compoundDrawables[0],trainIcon)
+    Assert.assertEquals(button.compoundDrawables[0], trainIcon)
+  }
+
+  @Test
+  fun givenBpkButtonInAnyTypeAndFormWhenIsEnabledIsSetThenSetupIsCalled() {
+    // Given
+    val subjectUnderTest = BpkButton(context).apply {
```
amolgupta: nit: same as the BDD comment above. Just calling it button should be good.

amolgupta:
Dismissing @tiagoengel 's review as the requested changes have been made and he is away for the week.

PR #80, opened 2018-09-28
[BPK-1959] Card component

tiagoengel:
Amol Gupta[08/10/2018]: Update on @tiagoengel 's description of the PR

As per my discussions with the users, they are not using Card this way. The approach we agreed upon is to have a CardView with recommended values of elevation, padding and supply the dividers as assets. This would allow them to put anything they required in the cardview. There are no cards with horizontal and vertical separations in current designs. 
screenshot_1539071956
screenshot_1539071959

--------------------------------------------------------------------------------
First of all a bit of context around why I decided to implement this API. It's not that simple to achieve ripple effect + elevation (but possible in API >= 21), on API 19 though it's very difficult, I've found myself reimplementing this class https://android.googlesource.com/platform/frameworks/support/+/lollipop-release/v7/cardview/eclair-mr1/android/support/v7/widget/RoundRectDrawableWithShadow.java from the CardView component. 

That wouldn't be good as we'd have to maintain that code. That's why we are using a CardLayout around the LinearLayout.

The inner LinearLayout is so we can provide a similar API to the RN version where users can customise how much space each view (main and stub) uses. By default is 50/50.

My first idea was to hide this extra LinearLayout inside our component and programatically move the views onto this layout. This also turned out to involve too much code to be worth it, which is why I decided for the API bellow:

```xml
<net.skyscanner.backpack.card.BpkCard>
    <net.skyscanner.backpack.card.BpkCardContent
      android:orientation="vertical">
        <AnotherView>
```

A big advantage of this approach is that we simply need to provide the LinearLayout attributes to `BpkCardContent` to control direction and how much space each view uses, no extra attr needed!

```xml
<net.skyscanner.backpack.card.BpkCard>
    <net.skyscanner.backpack.card.BpkCardContent
      android:orientation="horizontal">
        <AnotherView android:layout_weight="0.3" />
        <AnotherView android:layout_weight="0.7" />
```

The divider is shown if `BpkCardContent` has two views as children. I'm also only allowing it to have a max of two views.

The component is almost finished, pending only correct padding around the "divider" and reacting/exposing the component properties programatically.

TODO:
- [x] Fix divider padding
- [x] Allow properties to be set programatically
- [x] Snapshot tests
- [x] Unit tests
- [x] Docs

It looks like this ATM and works on API 19 too
screenshot_1538153874
screenshot_1538153882


Backpack/src/main/java/net/skyscanner/backpack/card/BpkCardContent.kt
```diff
@@ -0,0 +1,72 @@
...
+import android.widget.LinearLayout
+import net.skyscanner.backpack.R
+
+open class BpkCardContent(
+  context: Context,
+  attrs: AttributeSet?,
+  defStyleAttr: Int) : LinearLayoutCompat(context, attrs, defStyleAttr) {
+
+  constructor(context: Context) : this(context, null)
+  constructor(context: Context, @Nullable attrs: AttributeSet?) : this(context, attrs, 0)
```
antalindisguise: This boilerplate can be removed similarly how it is done in BpkButton

app/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,7 @@
         exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor
         in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur
         sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt laborum.</string>
+  <string name="stub_md" translatable="false">Lorem ipsum dolor sit amet, consectetur adipiscing elit.
```
antalindisguise: No need for this since `tools:text="@tools:sample/lorem"`: https://developer.android.com/studio/write/tool-attributes#toolssample_resources

amolgupta: Nice hack. Although this generates a new string each time and would make the documentation screenshots inconsistent.

Backpack/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
   implementation "com.android.support:support-annotations:$rootProject.ext.supportLibVersion"
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
   implementation "com.android.support:appcompat-v7:$rootProject.ext.supportLibVersion"
+  implementation "com.android.support:cardview-v7:$rootProject.ext.supportLibVersion"
```
antalindisguise: Should there be documentation on what 3rd party dependencies backpack will bring in for the sake of transparency? I know @k0nserv has mentioned to our squad that there should be no 3rd party dependency from backpack coming in (even though I don't consider these super helpful Android components 3rd party but rather framework components). Should it be somewhere described that we build upon the `cardview-v7` for the `BpkCard`?

Backpack/src/main/java/net/skyscanner/backpack/card/BpkCard.kt
```diff
@@ -0,0 +1,51 @@
+package net.skyscanner.backpack.card
+
+import android.content.Context
+import android.support.annotation.Dimension
+import android.support.v7.widget.CardView
+import android.util.AttributeSet
+import net.skyscanner.backpack.R
+
+
+open class BpkCard @JvmOverloads constructor(
```
antalindisguise: General comment about naming: Why don't we call these components as `*View` - thinking here of `BpkText`, `BpkCard`, etc.

Backpack/src/main/java/net/skyscanner/backpack/card/BpkCard.kt
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * @property padding
+   * padding for card
+   */
+  var padded: Boolean = false
+    set(value) {
+      field = value
+      @Dimension
```
antalindisguise: Is this necessary?

Backpack/src/main/java/net/skyscanner/backpack/card/BpkCard.kt
```diff
@@ -0,0 +1,51 @@
...
+   * focus state for card
+   */
+  var focused: Boolean = false
+    set(value) {
+      field = value
+      cardElevation = context.resources.getDimension(if (value) R.dimen.bpkElevationLg else R.dimen.bpkElevationXs)
+    }
+
+  private fun initialize(context: Context, attrs: AttributeSet?, defStyleAttr: Int) {
+    val a = context.obtainStyledAttributes(attrs, R.styleable.BpkCard, defStyleAttr, 0)
```
antalindisguise: Could be rewritten in a nicer API using Kotlin built in let or also language features as the `a` variable is a bit hard to read. Either that or maybe use `try {} finally {}` instead with more verbose variable name?

antalindisguise:
One more random thought: Shouldn't the elevation drop when pressed as Google defines/does in Material? 

oct-08-2018 11-01-30

Backpack/src/main/java/net/skyscanner/backpack/card/BpkCard.kt
antalindisguise: On iOS the `BpkText` is named `BpkTextView` but the rest is not - again inconsistency

image

Backpack/build.gradle
amolgupta: Will add extra deps wherever possible to readme and doc comments.

PR #88, opened 2018-10-10
[BPK-1986] Typeramp settings

amolgupta:
This is a work in progress.
CI will fail due to changes in text causing screenshots to be different.


UNRELEASED.md
```diff
@@ -1,3 +1,4 @@
 # Unreleased
 
-__Nothing yet__
+- Type ramp changes for line spacing and letter-spacing
+- Update build tools to `v28.0.3` and gradle plugin to `v3.20`
```
tiagoengel: This is not being done here so can be removed. Don't need to do it now but remember to remove when you move to the changelog.

PR #99, opened 2018-10-26
[BPK-1974] Switch component

tiagoengel:


Backpack/src/main/java/net/skyscanner/backpack/toogle/BpkSwitch.kt
```diff
@@ -0,0 +1,38 @@
+package net.skyscanner.backpack.toogle
```
shaundon:
```suggestion
package net.skyscanner.backpack.toggle
```

Will need to be changed across the package

tiagoengel: 🤦‍♂️

tiagoengel: Done

shaundon:
Can't get it running locally, not sure what this means. I can look into it further once I'm back.

screen shot 2018-10-26 at 14 51 43

tiagoengel:
Just checked and it works fine here. I can look into it with you

PR #102, opened 2018-10-30
Bpk 1946 hide toolbar and rtl toggle

amolgupta:
Adds developer shortcuts for toggle toolbar and layout direction

- Hide/Un-hide toolbar: Shift + T
- Toggle layout direction:  Shift + D

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailActivity.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailActivity.kt
@@ -38,12 +38,12 @@ class ComponentDetailActivity : AppCompatActivity() {
       detail_toolbar.title = itemId
 
       val createFragment = ComponentRegistry.getStoryCreator(itemId)
-      val fragment = createFragment?.createStory()
+      val fragment = createFragment.createStory()
 
-      val arguments = fragment?.arguments ?: Bundle()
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailActivity.kt
```diff
@@ -59,4 +56,26 @@ class ComponentDetailActivity : AppCompatActivity() {
     }
     return true
   }
+  override fun onKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
+    /*
+      Hide/Un-hide toolbar: Shift + T
```
tiagoengel: I think this comment is better outside the function.

app/src/main/java/net/skyscanner/backpack/demo/ComponentDetailActivity.kt
```diff
@@ -59,4 +56,26 @@ class ComponentDetailActivity : AppCompatActivity() {
...
+      toggle layout direction: Shift + D
+     */
+
+    if (keyCode == KeyEvent.KEYCODE_T && event!!.isShiftPressed) {
```
tiagoengel: Can you change `event!!.` to event?`?

PR #108, opened 2018-11-07
[BPK-1970] Add dialog view

tiagoengel:
TODO:
- [X] Test
- [x] Docs


Backpack/src/main/java/net/skyscanner/backpack/dialog/DialogVariables.kt
```diff
@@ -0,0 +1,21 @@
+package net.skyscanner.backpack.dialog
+
+import android.content.Context
+import net.skyscanner.backpack.R
+
+internal class DialogVariables(val context: Context) {
+
+  private val spacingSm = context.resources.getDimension(R.dimen.bpkSpacingSm)
+
+  val iconContainerSize = (spacingSm * 18).toInt()
```
amolgupta: use a constant directly. Changing spacing value will change the icons size which is not the expected behaviour

Backpack/src/main/java/net/skyscanner/backpack/dialog/DialogVariables.kt
```diff
@@ -0,0 +1,21 @@
+package net.skyscanner.backpack.dialog
+
+import android.content.Context
+import net.skyscanner.backpack.R
+
+internal class DialogVariables(val context: Context) {
```
amolgupta: Can it be DialogProerties instead of variables

app/src/main/java/net/skyscanner/backpack/demo/stories/DialogStory.kt
```diff
@@ -0,0 +1,142 @@
...
+import net.skyscanner.backpack.dialog.BpkDialog
+
+class DialogStory : Story() {
+
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
+    val dialogType = arguments?.getString(DialogStory.TYPE) ?: savedInstanceState?.getInt(DialogStory.TYPE)
+
+    val btn = view.findViewById<BpkButton>(R.id.open_dialog)
+    btn.setOnClickListener({
```
amolgupta: IDE warning, move lambda out of parentheses at multiple instances in this file

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialogIcon.kt
```diff
@@ -0,0 +1,78 @@
...
+import net.skyscanner.backpack.util.getColor
+
+internal class BpkDialogIcon @JvmOverloads constructor(
+  context: Context,
+  attrs: AttributeSet? = null,
+  defStyleAttr: Int = 0
+) : RelativeLayout(context, attrs, defStyleAttr) {
+
+  private val variables = DialogVariables(context)
+  private val imageView = ImageView(context)
```
amolgupta: use AppCompatImageView

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+
+  var icon: Icon? = null
+    set(value) {
+      field = value
+      viewHolder?.iconView?.icon = icon
+    }
+
+  fun addActionButton(button: BpkButton) {
+    button.layoutParams = LinearLayout.LayoutParams(
+      LinearLayout.LayoutParams.MATCH_PARENT,
```
amolgupta: use LinearLayoutCompat

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+  private fun getContentBackground(): GradientDrawable {
+    val shape = GradientDrawable()
+    shape.shape = GradientDrawable.RECTANGLE
+    val radii = variables.contentRadius
+    shape.cornerRadii = floatArrayOf(radii, radii, radii, radii, radii, radii, radii, radii)
+    shape.setColor(ResourcesCompat.getColor(context.resources, R.color.bpkWhite, context.theme))
+    return shape
+  }
+
+  private class ViewHolder(container: View) {
```
amolgupta: We should consider moving to kotlinx to avoid viewholders

Backpack/src/main/res/layout/bpk_dialog.xml
```diff
@@ -0,0 +1,41 @@
+<RelativeLayout
```
amolgupta: add `<?xml version="1.0" encoding="utf-8"?>`

Backpack/src/main/res/layout/bpk_dialog.xml
```diff
@@ -0,0 +1,41 @@
...
+  xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  android:layout_width="match_parent"
+  android:layout_height="match_parent"
+  android:background="@android:color/transparent" >
+
+  <LinearLayout
```
amolgupta: use LinearLayoutCompat

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    requestWindowFeature(Window.FEATURE_NO_TITLE)
+    setContentView(setUpContentView())
+    window?.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
+
+    val metrics = DisplayMetrics()
+    val windowManager = context.getSystemService(Context.WINDOW_SERVICE) as WindowManager
+    windowManager.defaultDisplay.getMetrics(metrics)
+    window?.setLayout((metrics.widthPixels * 0.85).toInt(), RelativeLayout.LayoutParams.WRAP_CONTENT)
```
amolgupta: what is `.85` ?

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private fun setUpContentView(): View {
+    val inflater = context.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
+    val view = inflater.inflate(R.layout.bpk_dialog, null)
+
+    viewHolder = ViewHolder(view).apply {
```
amolgupta: some of this stuff can be done via layout.xml in a easier more readable way.

Backpack/src/main/res/anim/dialog_slide_out_down.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate
+  xmlns:android="http://schemas.android.com/apk/res/android"
+  android:duration="200"
```
amolgupta: we should consider having animation duration tokens some time soon. @willhg

Backpack/src/main/java/net/skyscanner/backpack/dialog/DialogVariables.kt
tiagoengel: This is the container around the icon and the icon itself, and this is based on the IOS version that does the same thing. Also there is no constant size (token) for this size.

Backpack/src/main/java/net/skyscanner/backpack/dialog/DialogVariables.kt
tiagoengel: Can you explain why? In my mind the name variables is more declarative.

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
tiagoengel: 85% of the screen size. Again this is based on the IOS version.

Backpack/src/main/res/anim/dialog_slide_out_down.xml
tiagoengel: The animation is again based on the IOS version here and is trying to copy that. I checked the animation with @willhg (and the component in general). I agree tokens would be good but they should not be part of this PR, and also the IOS version does the same so we would need to update both.

app/src/main/java/net/skyscanner/backpack/demo/stories/DialogStory.kt
tiagoengel: Is not complaining here. I believe you don't have the ktlint config applied locally if it is complaining. I'll remove in any case but this is not supposed to be a problem.

Backpack/src/main/java/net/skyscanner/backpack/dialog/DialogVariables.kt
tiagoengel: I still don't understand how would this cause a problem. Again this is only for the circular container around the icon and not the icon itself, it will not affect the icon size/density. Also we do the same with react native in a few cases and I don't think it is a problem. Can you explain how this would cause a problem? I'm sure we will need to do this more times as it is quite common in the RN components.

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
tiagoengel: Again following what is done in IOS. I can add an upper limit though.

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
tiagoengel: Done

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
tiagoengel: How would that work?

Backpack/src/main/res/layout/bpk_dialog.xml
tiagoengel: done

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
```diff
@@ -0,0 +1,119 @@
...
+      window?.setWindowAnimations(R.style.Bpk_dialog_animation)
+    }
+
+    requestWindowFeature(Window.FEATURE_NO_TITLE)
+    setContentView(setUpContentView())
+    window?.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
+
+    val metrics = DisplayMetrics()
+    val windowManager = context.getSystemService(Context.WINDOW_SERVICE) as WindowManager
+    windowManager.defaultDisplay.getMetrics(metrics)
```
tiagoengel: I've removed the fixed size here and will let android dictate the size of the Dialog view. It prevents it from getting too big on tables or landscape.

Backpack/src/main/java/net/skyscanner/backpack/dialog/BpkDialog.kt
tiagoengel: I just removed this so it will use the default dialog size now.

Backpack/src/main/res/layout/bpk_dialog.xml
tiagoengel: Done

tiagoengel:
I've disabled tests in this PR as they are flaky on the CI (adding sleep to make it pass is not a good sign) and created this card to improve a bit the situation https://gojira.skyscanner.net/browse/BPK-2061

Backpack/src/main/res/anim/dialog_slide_out_down.xml
amolgupta: @willhg agreed to look into this and we will take it up later

Backpack/src/main/java/net/skyscanner/backpack/dialog/DialogVariables.kt
amolgupta: The problem is if we update the spacing value, we will need to verify at every place where such maths is done instead of just the spacing instances. An icon size can not be 18 times spacing value, it gives you the correct value but does not make any sense, these are 2 independent things. While updating the RN codebase to apply lint rules, there were instances where I used calculations where it was logical and ignored the rule where it did not. Eg, you can achieve the action bar size of `56dp` by multiplying `spacingXs` by `56` but is not a logical thing to do.

PR #109, opened 2018-11-08
[BPK-2044] Add heavy styles

shaundon:
**To do**
- [x] Make Gulp generate heavy styles
- [x] Add to example app
- [x] Expose in Backpack gradle package
- [x] Add tests for it on Android side

**Notes**
I refactored the Gulp task a fair bit. It seemed like some of the logic was redundant so I removed it. I'm pretty sure the new stuff I'm generating has all the old stuff (plus heavy styles) so I think we're good. The diff is a little hard to parse as the order of lines has changed.

The API should probably change. Right now `emphasize` is a boolean prop, but we should probably do something like have a `weight` prop that can be `normal`, `emphasized` or `heavy`. That what we did on RN too. I guess this might be breaking too? Although I imagine putting in a PR to the app to fix this wouldn't be too difficult.


docs/Text/README.md
```diff
@@ -8,23 +8,25 @@ Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/b
 
 The Text component can be used in both XML and Kotlin
 
-Example of a text view with large font in XML
+Example of a text view with large and ephasized font in XML
```
shaundon:
```suggestion
Example of a text view with large and emphasized font in XML
```

docs/Text/README.md
```diff
@@ -8,23 +8,25 @@ Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/b
...
 <net.skyscanner.backpack.text.BpkText
   android:layout_width="wrap_content"
   android:layout_height="wrap_content"
   android:text="Flights to Edinburgh"
+  app:weight="emphasized"
   app:textStyle="xxl" />
 ```
 
-Example of a text view with large font in Kotlin
+Example of a text view with large and ephasized font in Kotlin
```
shaundon:
```suggestion
Example of a text view with large and emphasized font in Kotlin
```

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -69,23 +91,29 @@ open class BpkText(
...
+
+    val textAppearance = styleProps[weight.ordinal]
+    textAppearance ?: throw IllegalStateException("Weight $weight is not supported for the current size")
 
     if (textStyle == CAPS) {
       isAllCaps = true
     }
 
-    if (styleProps != null) {
-      TextViewCompat.setTextAppearance(this, if (emphasize) styleProps.first else styleProps.second)
```
shaundon: Without this line, how does `emphasize` get applied? We still want it to take effect right, if `weight` isn't passed in. I might have just missed it.

tiagoengel: Is still being set on line 117, I just moved the null check logic up. `TextViewCompat.setTextAppearance(this, textAppearance)` this is what applies all the styles to the textview.

shaundon: Ah yeah - nice 😄

shaundon:
I've just run it locally and it works great. You've got some ktlint errors but otherwise looks good to me.

PR #111, opened 2018-11-12
[Bpk-1930] Add gradient story

amolgupta:
Add story and docs for gradient

screenshot_1542010136


app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -52,7 +53,7 @@ private class NodeData(
   val items: Map<String, RegistryItem> = emptyMap()
 ) {
 
-  constructor(creator: () -> Story): this({ _ -> creator() })
+  constructor(creator: () -> Story) : this({ _ -> creator() })
```
tiagoengel: Why this change? All other entries in this file don't have the space before.

tiagoengel:
Looks good I have only one small question. Also the IOS version is quite different. We should try to keep both example apps in sync. Do you think we should follow what is in the IOS app or just create a card to make the IOS app follow the android examples?

screen shot 2018-11-13 at 10 48 34

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
amolgupta: The auto-formatter does that only for this instance. Maybe it has a different rule for constructors.

amolgupta:
I talked to @georgegillams about this and the android approach seemed more explanatory in this case. Maybe iOS can modify their version of the story.

PR #113, opened 2018-11-14
[BPK-1976] Chip component

amolgupta:
Chip component.
Initial approach was to extend Chip from material components. Dropped it because:
- It enforces the use of Material theme by the consumer of the component
- The `elevation` and `ChipDrawable` properties can not be modified according to the backpack theme

Notes(only for `BpkChipDismissible`):
- Added another drawable `bpk_close_small` as resizing the vector drawable programatically is too hackey and involves converting it into a bitmap. We might look at having all the icons in small size as well, even though they are vectors.
- The snapshot tests do not display the shadow and close icon. Had the same issue with button tests.
- Removed the documentation for the dismissible chip as it is out of scope. Leave the implementation as is, if it comes up again.

screenshot_1542350921
screenshot_1542350925


Backpack/src/main/res/color/bpk_chip_text_color.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+
+  <item android:color="@color/bpkWhite" android:state_enabled="true" android:state_selected="true" />
```
tiagoengel: Just a matter of preference but I prefer to keep this in the code. Won't reject the pr because of this though 😇

build.gradle
```diff
@@ -61,4 +61,5 @@ ext {
   googlePlayServicesVersion = "15.0.1"
   androidMapsUtilsVersion = "0.5+"
   androidx = "1.0.0"
+  materialDesign = "1.0.0"
```
tiagoengel: Is this still required?

docs/Chip/README.md
```diff
@@ -0,0 +1,34 @@
...
+      app:selected="true" />
+```
+
+Example of a chip in Kotlin
+
+```Kotlin
+import net.skyscanner.backpack.chip.BpkChip
+
+BpkChip(context).apply {
+  isSelected = true
```
tiagoengel: We should maybe add the text here as it will look add without any text

tiagoengel:
I've tested it locally and found two problems. 
1. It's not working on API 19, see:
screen shot 2018-11-19 at 10 40 31
I think elevation does not work properly on 19 but we need at least the border.

2. The ripple effect is not quite right. 
- In both IOS and the RN version it goes dark grey when pressed, and goes blue and the user releases the chip. This version goes blue even before releasing it.
- Same goes for unselecting, it should go dark blue and then white (when released)
- The ripple effect is a bit weird, the RN version is using the `selectableitembackground` ripple effect.

Backpack/src/main/res/color/bpk_chip_text_color.xml
amolgupta: I try to do `xml` for design/config and code for logic wherever possible.

Backpack/src/androidTest/java/net/skyscanner/backpack/chip/BpkChipTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  @Test
+  fun test_disabled_state() {
+    val chip = BpkChip(context).apply {
+      disabled = true
+    }
+    Assert.assertEquals(ContextCompat.getColor(context, R.color.bpkGray300), chip.currentTextColor)
+  }
+
+  fun test_select_event() {
```
tiagoengel: Those need the @Test annotation

Backpack/src/main/java/net/skyscanner/backpack/chip/BpkDismissibleChip.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.core.widget.TextViewCompat
+import net.skyscanner.backpack.R
+
+open class BpkDismissibleChip @JvmOverloads constructor(
+  context: Context,
+  attrs: AttributeSet? = null,
+  defStyleAttr: Int = R.style.Bpk_chip
+) : BpkChip(context, attrs, defStyleAttr) {
+
+  private var onDismissListener: OnDismissListener? = null
```
tiagoengel: This doesn't seem to be set anywhere? Why are we using this?

amolgupta: The user will need to set this. Its a wrapper around onClickListener.

tiagoengel: but it's always null here, how is this supposed to be used? Unless I'm missing something this or not adding the click listener will result in the same, so we can remove all this.

willhg:
Yep all good to go now aligned to same visual RN behaviour

PR #118, opened 2018-11-16
[BPK-2046] Update elevation tokens

shaundon:


georgegillams:
LGTM, but I think someone with more Android expertise should review 👍

PR #121, opened 2018-11-20
update snapshot tests

amolgupta:
Adding a wrapper view and a background color to snapshot tests.
- This makes is easier to view the snapshots and understand the component
- Changes to elevation and borders are more visible


amolgupta:
This is not adding any value. 
Depends on a bug reported in the snapshot testing tool at https://github.com/facebook/screenshot-tests-for-android/issues/104

PR #124, opened 2018-11-21
[BPK-2061] Flaky tests

tiagoengel:
It turned out android already has its own way of doing this so I didn't create any other class for this.


.travis.yml
```diff
@@ -73,11 +73,11 @@ script:
   # Run linters
   - ./gradlew check
   # Run espresso tests in the Backpack module ( this excludes snapshot tests)
-  - ./gradlew Backpack:connectedCheck
+  - ./gradlew -Pandroid.testInstrumentationRunnerArguments.notAnnotation=androidx.test.filters.FlakyTest Backpack:connectedCheck
```
amolgupta: these arguments can be put in `build.gradle` under `defaultConfig` as 
```testInstrumentationRunnerArgument "notAnnotation", "androidx.test.filters.FlakyTest"```

tiagoengel: But that would change every execution, we only want to ignore those tests on CI

PR #129, opened 2018-11-23
[NO-JIRA] Run npm ci before release

tiagoengel:
We (I) recently released a version of backpack-android with outdated tokens. This happened for two reasons:

1. My local version of `bpk-tokens` was outdated and the release process generates a bunch of things based on that.
2. Our snapshot tests don't capture elevation (and those where elevation tokens)

Point 2 is really hard to fix so I added `npm ci` command before the release process to make sure we do it with a clean and updated environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "preemulator": "npm run emulator:boot",
     "emulator": "adb wait-for-device && sleep 5 && npm run emulator:build && npm run emulator:test",
     "postemulator": "npm run emulator:kill",
-    "prerelease": "npm run emulator",
+    "prerelease": "npm ci && npm run emulator",
     "release": "node release.js",
     "test": "eslint . --ext .js,.jsx",
     "prebuild": "npm test",


package.json
```diff
@@ -16,7 +16,7 @@
     "preemulator": "npm run emulator:boot",
     "emulator": "adb wait-for-device && sleep 5 && npm run emulator:build && npm run emulator:test",
     "postemulator": "npm run emulator:kill",
-    "prerelease": "npm run emulator",
+    "prerelease": "npm ci && npm run emulator",
```
matthewdavidson: Why `npm ci` over `npm install` ?

tiagoengel: Mainly because it doesn't update the package lock. But also because it removes the node_modules first which I think gives us even more assurance that we a clean and updated environment before the release.

matthewdavidson: Ah interesting, i had no idea it removed `node_modules` first ...

matthewdavidson:
We should do the same on iOS too (if not already there). cc @shaundon, @georgegillams, @k0nserv

PR #130, opened 2018-11-26
[BPK-2080] Debug markers

amolgupta:
##Spike
Trying to show markers over all `bpk` views to debug clients and measure usage.
screenshot_1543310830


amolgupta:
This approach does not scale.

PR #132, opened 2018-11-27
[Bpk-2080] Add debug markers 

amolgupta:
Adds debug markers to all Backpack views.
Installation instructions:

Add the interceptor inr the Application class's `onCreate` as follows:

```
    ViewPump.init(ViewPump.builder()
      .addInterceptor(BpkInterceptor())
      .build())
```

Add the contextWrapper to the base activity as:
```
  override fun attachBaseContext(newBase: Context) {
    super.attachBaseContext(ViewPumpContextWrapper.wrap(newBase))
  }
```
Caveats:
- Introduces the dependecy on `inflaterX` in the app. This can be moved to test dependency if required
- Adds another function in the all the base class activities.

### DIFF
--- a/Backpack/build.gradle
+++ b/Backpack/build.gradle
@@ -35,6 +35,7 @@ dependencies {
   androidTestImplementation 'androidx.test:rules:1.1.1'
   androidTestImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.1.0'
   androidTestImplementation 'org.mockito:mockito-android:2.23.4'
+  api 'io.github.inflationx:viewpump:1.0.0'
   testImplementation 'junit:junit:4.12'
   // NOTE: This is here and not in the root file because snyk is not ignoring it otherwise.
   // This is a test dependency and should be ignored (as stated in snyk docs)
--- a/Backpack/src/main/java/net/skyscanner/backpack/util/BpkInterceptor.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/util/BpkInterceptor.kt
@@ -6,14 +6,14 @@ import androidx.annotation.RequiresApi
 import io.github.inflationx.viewpump.InflateResult
 import io.github.inflationx.viewpump.Interceptor
 
-class BpkInterceptor(private val enable: Boolean = false) : Interceptor {
+class BpkInterceptor : Interceptor {
 
   @RequiresApi(Build.VERSION_CODES.M)
   override fun intercept(chain: Interceptor.Chain): InflateResult {
     val result = chain.proceed(chain.request())
     val view = result.view()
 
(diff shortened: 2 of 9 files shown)


Backpack/build.gradle
```diff
@@ -35,6 +35,7 @@ dependencies {
   androidTestImplementation 'androidx.test:rules:1.1.1'
   androidTestImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.1.0'
   androidTestImplementation 'org.mockito:mockito-android:2.23.4'
+  api 'io.github.inflationx:viewpump:1.0.0'
```
tiagoengel: does this really need to be `api`? Don't remember where I read it but supposedly `api` should be avoided if possible

Backpack/src/main/java/net/skyscanner/backpack/util/BpkInterceptor.kt
```diff
@@ -0,0 +1,37 @@
...
+
+  @RequiresApi(Build.VERSION_CODES.M)
+  override fun intercept(chain: Interceptor.Chain): InflateResult {
+    val result = chain.proceed(chain.request())
+    val view = result.view()
+
+    if (view != null && isBpkView(view.javaClass)) {
+      if (view.background == null) {
+
+        view.background = ColorDrawable(0x77ac650c)
```
tiagoengel: nit: Would be good to move `0x77ac650c` to a constant with a descriptive name. E.g. `COLOR_BROWN` or just a comment here

app/src/main/java/net/skyscanner/backpack/demo/BackpackDemoApplication.kt
```diff
@@ -1,16 +1,51 @@
...
+      val componentName = intent!!.component
+      val mainIntent = Intent.makeRestartActivityTask(componentName)
+      context.startActivity(mainIntent)
+      Runtime.getRuntime().exit(0)
+    }
+  }
+
   override fun onCreate() {
     super.onCreate()
+    instance = (applicationContext as BackpackDemoApplication?)!!
```
tiagoengel: nit: this could be better written as
```
applicationContext?.let {
  // all code here
}
```

or 
```
instance = applicationContext!! as BackpackDemoApplication
```

Backpack/build.gradle
amolgupta: using `api` will pass down the implementation of the lib to the consumer apps. This is required in our case.

Backpack/src/main/java/net/skyscanner/backpack/util/BpkInterceptor.kt
amolgupta: I will add as comment, as colors file is generated atm.

PR #133, opened 2018-11-30
[BPK-2103] Rename white buttons to outline

shaundon:
The entry in unreleased.md makes it sound like it's completely new because we forgot to add the entry the first time.


app/src/main/java/net/skyscanner/backpack/demo/stories/ButtonStory.kt
```diff
@@ -29,8 +29,8 @@ class ButtonStory : Story() {
     view.findViewById<BpkButton>(R.id.btn_start_icon).type = buttonType
     view.findViewById<BpkButton>(R.id.btn_icon).type = buttonType
 
-    if (type == "white") {
-      view.setBackgroundColor(ContextCompat.getColor(context!!, R.color.bpkBlue500))
+    if (type == "outline") {
+      view.setBackgroundColor(ContextCompat.getColor(context!!, R.color.bpkBlue400))
```
shaundon: Changed to blue 400 to match iOS.

PR #152, opened 2018-12-14
[Bpk-2088] Button size bugfix

amolgupta:
screenshot_1544803809


willhg:
button_measures

Has the padding shifted with this change? it looks like the icon and text lock up is slightly off centre? The measurements are different at either end.

PR #153, opened 2018-12-14
Bpk-2119 update minsdk

amolgupta:
This PR removes support for Android Kitkat. The `minSdk` required will now be `21`

- Moves resources under `v21` folders to default folders
- Removes Multidex support [see](https://developer.android.com/studio/build/multidex)
- Removes Vector Drawables support [see](https://developer.android.com/guide/topics/graphics/vector-drawable-resources)
- Updates CI and gulp files


Backpack/src/main/res/values/backpack.text.xml
```diff
@@ -25,96 +25,115 @@
...
   <style name="bpkTextSm">
     <item name="android:fontFamily">sans-serif</item>
     <item name="android:textSize">14sp</item>
+    <item name="android:letterSpacing">0</item>
   </style>
 
   <style name="bpkTextXl">
     <item name="android:fontFamily">sans-serif</item>
     <item name="android:textSize">24sp</item>
+    <item name="android:letterSpacing">-0.024999999999999998</item>
```
antalindisguise: Where are these numbers documented as the correct ones?

app/build.gradle
```diff
@@ -16,7 +16,6 @@ android {
     targetSdkVersion rootProject.ext.targetSdkVersion
     versionCode 1
     versionName "1.0"
-    multiDexEnabled true
     vectorDrawables.useSupportLibrary = true
```
antalindisguise: Can be dropped as well according to https://developer.android.com/guide/topics/graphics/vector-drawable-resources

templates/BackpackText.njk
```diff
@@ -25,6 +25,7 @@
   <style name="{{text.name}}">
     <item name="android:fontFamily">{{text.fontFamily}}</item>
     <item name="android:textSize">{{text.size}}sp</item>
+    <item name="android:letterSpacing">{{text.letterSpacing}}</item>
```
antalindisguise: Shouldn't this go in a separate PR?

Backpack/src/main/res/values/backpack.text.xml
amolgupta: those are type scale changes and are generated from tokens. I can confirm they are correct.

app/build.gradle
amolgupta: good catch. I removed it in the library already, removing from the demo app now.

templates/BackpackText.njk
amolgupta: This is the part that generates the type scale values. CI would fail if not done along with this PR.

PR #154, opened 2018-12-14
[No-JIRA] Run AndroidX and Appcompat migrations 

amolgupta:
Runs the `Android Studio` tools : 
- Migrate to AppCompat
- Migrate to AndroidX

### DIFF
--- a/Backpack/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
+++ b/Backpack/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
@@ -11,7 +11,7 @@ import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.filters.FlakyTest
 import androidx.test.rule.ActivityTestRule
-import androidx.test.runner.AndroidJUnit4
+import androidx.test.ext.junit.runners.AndroidJUnit4
 import net.skyscanner.backpack.R
 import net.skyscanner.backpack.button.BpkButton
 import net.skyscanner.backpack.util.TestActivity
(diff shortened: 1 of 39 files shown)


Backpack/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
```diff
@@ -25,7 +25,7 @@ import org.junit.runner.RunWith
 class BpkDialogTest {
 
   private lateinit var handler: Handler
-  private lateinit var mActivity: Activity
+  private lateinit var mActivity: AppCompatActivity
```
antalindisguise: Can we maybe drop the `m` prefix?

Backpack/src/main/res/values/backpack.button.xml
```diff
@@ -6,9 +6,9 @@
   </declare-styleable>
 
   <style name="bpkButtonBase">
-    <item name="android:fontFamily">sans-serif-light</item>
+    <item name="fontFamily">sans-serif-light</item>
```
antalindisguise: Are these changes intended?

app/build.gradle
```diff
@@ -45,13 +45,13 @@ dependencies {
   implementation "androidx.appcompat:appcompat:$rootProject.ext.androidx"
   implementation "androidx.legacy:legacy-support-v4:$rootProject.ext.androidx"
   implementation "com.google.android.material:material:$rootProject.ext.androidx"
-  implementation 'com.android.support.constraint:constraint-layout:1.1.2'
+  implementation 'androidx.constraintlayout:constraintlayout:1.1.2'
```
antalindisguise: Could you bump it to `1.1.3` maybe as the latest stable?

Backpack/src/main/res/values/backpack.button.xml
amolgupta: These are done by the migration tool.

Backpack/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
amolgupta: I can take this up as a separate PR and do it in the complete codebase.

Backpack/src/main/res/values/backpack.button.xml
Bence1001: Removing the `android:` prefix means we are overriding `fontFamily` in the AppCompat Theme. If this is what we want then it's correct.

PR #155, opened 2018-12-17
BpkCalendar component

antalindisguise:
https://confluence.skyscannertools.net/display/~antalmonori/Android+BpkCalendar+component

~Still missing:~
- ~Add story into example app~
- ~Update documentation~
- ~Fix weekday header view after it got rewritten with constraintlayout~

Signed-off-by: Antal Monori <antal.monori@skyscanner.net>


Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -0,0 +1,56 @@
...
+
+  abstract val minDate: Calendar
+
+  abstract val maxDate: Calendar
+
+  abstract val isRtl: Boolean
+
+  abstract val locale: Locale
+
+  abstract fun onRangeSelected(range: CalendarRange)
```
Bence1001: Isn't there too many abstract properties and methods here? This forces the client to override all of them and write too many boilerplate code (while they only actually use a subset of them). I think a better API design would be to provide sensible defaults, like `Date()` (today) for `minDate`, empty function for `onRangeSelected` and so on.

antalindisguise: Good idea! I refactored it a bit, so hopefully it's more clear now!

PR #168, opened 2019-01-11
[BPK-2139] Add unit tests for BpkCalendar

tiagoengel:
This adds unit tests for the classes that were not copied from the google lib. I'll do snapshots in a different PR as they are not working well :/

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
@@ -55,6 +55,9 @@ abstract class BpkCalendarController {
     onRangeSelected(selectedRange)
   }
 
+  // TODO: This will not respect differences in the order of the fields for each country.
+  // E.g. US = mm/dd/yyyy and UK = dd/mm/yyyy
+  // Shall we use DateFormat.getDateInstance instead?
   internal fun getLocalizedDate(date: Date, pattern: String): String = SimpleDateFormat(pattern, locale).format(date)
 
   private companion object {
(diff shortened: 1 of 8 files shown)


Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -55,6 +55,9 @@ abstract class BpkCalendarController {
     onRangeSelected(selectedRange)
   }
 
+  // TODO: This will not respect differences in the order of the fields for each country.
+  // E.g. US = mm/dd/yyyy and UK = dd/mm/yyyy
+  // Shall we use DateFormat.getDateInstance instead?
```
amolgupta: do we plan to fix this in this PR?

tiagoengel: Not really, it's really a problem as far as I can tell because just now it is only used to localise the weekdays, so it doesn't matter much. But we can improve it at some point

antalindisguise: Do you mean `SimpleDateFormat(pattern, locale)` is not enough? As they are injected when the Controller is subclassed, it should not be an issue. Locale in our app is generated from our custom set of rules and therefore it should respect the differences of the fields based on the locale that is passed here. Passing on `mm/dd/YYYY` but UK locale should correctly format the date

antalindisguise:
👍 💯

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
tiagoengel: What I meant is that for the UK dates should be `dd/mm/yyyy` and US `mm/dd/yyyy`. Doing `SimpleDateFormat("MM/dd/YYYY", Locale.UK).format(Calendar.getInstance().time)` returns `01/14/2019` which works in the sense that it formats the date but it's not the correct pattern. 

Now if the pattern is always correct then there is no problem. The consumer has to provide the correct pattern, which I believe is what the app does. But there are ways of letting java do it. 

In any case I don't want to change it now, I just had the TODO there so we can revisit at some point and improve it or just leave it as it is.

PR #170, opened 2019-01-14
[BPK-2139] Add snapshot tests for BpkCalendar

tiagoengel:
Had to mark them as flaky because they have the same problem as dialog tests... They will still run locally and before releasing but I need to look into this at some point.


Backpack/src/androidTest/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarControllerTest.kt
```diff
@@ -127,6 +127,19 @@ class BpkCalendarControllerTest {
...
+  fun test_isToday() {
+    val today = Calendar.getInstance()
+    val day = today.get(Calendar.DAY_OF_MONTH)
+    val month = today.get(Calendar.MONTH)
+    val year = today.get(Calendar.YEAR)
+
+    Assert.assertTrue(subject.isToday(year, month, day))
+    Assert.assertFalse(subject.isToday(year, month, day + 1))
+    Assert.assertFalse(subject.isToday(year, month + 1, day))
+    Assert.assertFalse(subject.isToday(year + 1, month, day))
```
antalindisguise: Why are these three `assertFalse` assertions needed? Usually it's more than enough to test the correct expected assertion and not over-specify it with alternative ones.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -60,6 +60,10 @@ abstract class BpkCalendarController {
   // Shall we use DateFormat.getDateInstance instead?
   internal fun getLocalizedDate(date: Date, pattern: String): String = SimpleDateFormat(pattern, locale).format(date)
 
+  open fun isToday(year: Int, month: Int, day: Int): Boolean {
```
antalindisguise: Why does this needs to be all the way up in the controller and specially be left `open`? Do we expect this behavior to be defined by the consumer of this library? Sounds more like a helper method to me 🤔

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
```diff
@@ -0,0 +1,180 @@
...
+  override val startDate: Calendar
+    get() = initialStartDate ?: super.startDate
+
+  override val endDate: Calendar
+    get() = initialEndDate ?: super.endDate
+
+  override fun onRangeSelected(range: CalendarRange) {}
+
+  override fun isToday(year: Int, month: Int, day: Int): Boolean {
+    return day == 2 && month == 0 && year == 2019
```
antalindisguise: What is month `0`?

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
```diff
@@ -0,0 +1,180 @@
...
+      Locale.UK,
+      getDate(2019, 0, 2),
+      getDate(2019, 11, 31))
+
+    calendar.setController(controller)
+    val wrapped = wrapWithBackground(calendar)
+
+    val asyncScreenshot = prepareForAsyncTest()
+
+    mActivity.runOnUiThread {
```
antalindisguise: Minor nit, but I prefer we don't use the old Android code style but rather follow the new Google Kotlin style guide. The Hungarian Notation is generally discouraged even by Google now.

btw I'm referring to `mActivity` here

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
tiagoengel: I need this for the snapshot testing. We need "today" to be always the same day during snapshots. Couldn't find a better way of doing it.

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarControllerTest.kt
tiagoengel: Because with only the first one the method could be just returning true and it would pass. Those assertions is to make sure all parameters are being used correctly.

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
tiagoengel: January, date apis generally start from 0 to 11 🤷‍♂️

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
tiagoengel: Agreed, I copied this and I forgot to change :). Will do

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarControllerTest.kt
antalindisguise: Gotcha. Makes sense now if I think about it. 👍

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
antalindisguise: Not in the new java.time.* library; there it has a constant value of 1. But you are right, what we are using here is java.util.Calendar which still has the old 0-based starting index. 

https://developer.android.com/reference/java/util/Calendar#JANUARY (available since API 26; added in JDK 1.8)
https://developer.android.com/reference/java/time/Month#JANUARY (available since API 1)

🤷‍♂️ 😅

antalindisguise: So all in all, you were right. My bad.

PR #171, opened 2019-01-16
No-JIRA update gradle version

amolgupta:


amolgupta:
Merging this is not possible till either `ktlint` starts supporting gradle 5 or we find an alternative.

PR #173, opened 2019-01-21
[BPK-2127] Changes to bridge the calendar

tiagoengel:


Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
```diff
@@ -31,18 +32,9 @@ import net.skyscanner.backpack.calendar.presenter.MonthAdapter
 /**
  * This displays a list of months in a calendar format with selectable days.
  */
-internal class CalendarView constructor(context: Context, attr: AttributeSet?) : ListView(context, attr),
-  OnScrollListener {
+internal class CalendarView constructor(context: Context, attr: AttributeSet?) : ListView(context, attr) {
 
-  override fun onScrollStateChanged(view: AbsListView?, scrollState: Int) {
```
tiagoengel: Removing the scroll logic is not necessary but is not doing anything...

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -64,6 +64,13 @@ abstract class BpkCalendarController {
     return CalendarDay(year, month, day).date == CalendarDay().date
   }
 
+  fun updateSelection(range: CalendarRange) {
```
antalindisguise: Is there any use-cases currently that require to programatically set the range/selection? I am just a bit doubtful if it is clear what the purpose of this is when now there is two implementation that's updating selection:
- `onDayOfMonthSelected(selectedDay: CalendarDay)`
- `updateSelection(range: CalendarRange)`

And one that the implementation needs to define: `onRangeSelected(range: CalendarRange)`

Would it make sense to at least update the `onDayOfMonthSelected(selectedDay: CalendarDay)` code to utilize this new `updateSelection` function too?

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
```diff
@@ -31,19 +31,21 @@ import net.skyscanner.backpack.calendar.presenter.MonthAdapter
...
+) : ListView(context, attr), AbsListView.OnScrollListener {
 
   private var scrollFriction = 1.0f
   private var selectedDay: CalendarDay = CalendarDay()
   private var previousScrollPosition: Long = 0
   private var previousScrollState = OnScrollListener.SCROLL_STATE_IDLE
   private var currentScrollState = OnScrollListener.SCROLL_STATE_IDLE
 
+  override fun onScrollStateChanged(view: AbsListView?, scrollState: Int) {
+    // TODO
```
antalindisguise: Why did this became a TODO? I don't think we expect any behavior on scrollStateChange, hence it contained a comment before. Are we planning to do anything here?

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
```diff
@@ -61,6 +63,21 @@ internal class CalendarView constructor(context: Context, attr: AttributeSet?) :
...
+      View.MeasureSpec.makeMeasureSpec(height, View.MeasureSpec.UNSPECIFIED))
+    layout(left, top, right, bottom)
+  }
+
+  override fun requestLayout() {
+    super.requestLayout()
+
+    // This is required for the bridged component to render correctly
+    // based on: https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPicker.java#L75
+    this.post(measureAndLayout)
```
antalindisguise: Is this native component going to be used in React Native code as well? Isn't there a React Native calendar implementation already?

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
```diff
@@ -81,9 +98,7 @@ internal class CalendarView constructor(context: Context, attr: AttributeSet?) :
     setBackgroundColor(ContextCompat.getColor(context, R.color.bpkWhite))
   }
 
-  /**
-   * Updates the title and selected month if the view has moved to a new month.
-   */
+  // TODO: Updates the title and selected month if the view has moved to a new month.
```
antalindisguise: Why did this turn into a TODO again?

UNRELEASED.md
```diff
@@ -1 +1,9 @@
 # Unreleased
+
+**Added:**
+
+- BpkCalendar
+  - Added abilitty to update the selected dates programmatically using `updateSelection` in BpkCalendarControler.
+  - BpkCalendar is can now be extended.
```
antalindisguise: spelling:
- `abilitty -> `ability`
- `BpkCalendarControler -> `BpkCalendarController`
- `is can` -> `can`

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
```diff
@@ -170,6 +171,21 @@ class BpkCalendarTest : BpkSnapshotTest() {
       }
   }
 
+  @Test
+  @FlakyTest
```
antalindisguise: Would it make sense to write some detail on why is this flaky as it is defined here: https://developer.android.com/reference/android/support/test/filters/FlakyTest ?

tiagoengel: It's flaky because it fails on CI, we have a bunch of those cases and we need to figure out what's causing it. But for now we run them only locally and not on ci. Exactly why it's flaky we don't, it just times out on ci.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
tiagoengel: Are we not planning to add the same functionality present on IOS? That shoes the current year at the top, over the calendar? That's why I changed to TODO. If we don't plan to do anything with this it can be just removed. But I believe we do plan to add this feature

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
tiagoengel: No, we don't have it for React Native, those changes are exactly for that purpose. We are bridging this component to react native here https://github.com/Skyscanner/backpack-react-native/pull/34/files

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
tiagoengel: Because it doesn't actually do this right now, and is something the IOS version does

UNRELEASED.md
tiagoengel: 👍 I'll fix this when I release and move this entry to the changelog.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
antalindisguise: Gotcha!

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
antalindisguise: I think we do -- I believe Stark is just about to start doing that. We can keep it, just didn't understand that change belonging to this PR.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
antalindisguise: Good to know!

PR #176, opened 2019-01-25
Add hockeyapp integ

amolgupta:


app/build.gradle
```diff
@@ -15,7 +15,7 @@ android {
     minSdkVersion rootProject.ext.minSdkVersion
     targetSdkVersion rootProject.ext.targetSdkVersion
     versionCode 1
```
tiagoengel: Shouldn't we change this too?

amolgupta: it takes only integer values and is mostly for play-store etc to track version. I don't see value in it atm.

PR #179, opened 2019-01-30
[BPK-2212] New social like icon

matthewdavidson:
screenshot_1548847587

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,1 +1,4 @@
 # Unreleased
+
+**Added:**
+- New `bpk_social_like` icon.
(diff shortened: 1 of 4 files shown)


UNRELEASED.md
```diff
@@ -1 +1,4 @@
 # Unreleased
+
+**Added:**
+- New `bpk_social_like` icon.
```
georgegillams: Shouldn't this appear under a component name?

matthewdavidson: Icons in Android don't belong to a component, they are _[resources](https://developer.android.com/reference/android/content/res/Resources)_.

PR #180, opened 2019-01-30
Removed year from month header

Szabadember:
As per Antal's suggestion, I removed the year from the month header because this functionality is replaced by the year pill now.


Szabadember:
@tiagoengel I couldn't reproduce this. The text 9999 is included in the XML but it is immediately overwritten by the current year on start.

Szabadember:
@tiagoengel Do we still need to add things to UNRELEASED if we release these changes today?

Szabadember:
@tiagoengel I see 9999 on the snapshot test, but this won't happen in real life. I guess the snapshot is taken faster than the UI updated but this transition is invisible. One fix could be to change "9999" to "" inside the xml

tiagoengel:
Cool, the unreleased file is just to make the life of whomever in the team is releasing easier.

PR #192, opened 2019-02-14
[BPK-2246] Add settings page and theme switcher

tiagoengel:
kapture 2019-02-15 at 11 33 32


app/src/main/res/values/styles.xml
```diff
@@ -22,4 +22,12 @@
     <item name="android:padding">@dimen/bpkSpacingBase</item>
   </style>
```
amolgupta: this should be in `themes.xml` as Android convention.

app/src/main/res/layout/component_settings_theme_option.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
amolgupta: can use a `CheckedTextView` instead of `text` + `ImageView`

PR #195, opened 2019-02-21
Improve BPK calendar date selection

tibor92:
BPK calendar date changing mechanism is updated with new logic (every selection after a range defined clears the previously selected range).


tiagoengel:
Code looks good, I'll just run it locally after it passes CI

PR #207, opened 2019-03-12
Colored calendar update mechanism with example

Dominguezzz:
- Use case: preserve state but update colors/calendar content.
- Solution: Introduced CalendarUpdateCallback so the Calendar can be updated via the controller


Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -32,6 +33,8 @@ abstract class BpkCalendarController {
 
   internal val selectedRange: CalendarRange = CalendarRange()
 
+  internal var updateCallbackContentCallback: CalendarUpdateCallback? = null
```
tiagoengel: Can we call this just `updateContentCallback`?

Dominguezzz: 🤦‍♂️

PR #210, opened 2019-03-15
[NO-JIRA] Add listeners decision

tiagoengel:
Based on the problems and decision we made on #209 

### DIFF
--- a/decisions/event-listeners-and-state.md
+++ b/decisions/event-listeners-and-state.md
@@ -68,7 +68,7 @@ class MyCustomChip {
 
 ```
 
-In the example above `setOnClickListener` will be called by the super class before the props and init block have been called in the child class. This can be very hard to identify, especially from a user perspective.
+In the example above `setOnClickListener` will be called by the super class before the props and `init block` have been called in the child class. This can be very hard to identify, especially from a user perspective.
 
 **Flexibility:**
 
--- a/decisions/event-listeners-and-state.md
+++ b/decisions/event-listeners-and-state.md
@@ -79,5 +79,5 @@ to add a custom listener they will unintentionally override the behaviour we had
 
 Documentation with examples should be provided to remind users a listener should be added. 
 
-In the ocasions where a listener is absolutelly required we should take precautions to avoid 
+In the occasions where a listener is absolutely required we should take precautions to avoid 
 the two problems above, one course of action is a different event users can register without affection the default behaviour. For example, BpkChip could provide a `setOnSelectedStateChangeListener` so users can react to changes without replacing the click listener.
--- a/decisions/event-listeners-and-state.md
+++ b/decisions/event-listeners-and-state.md
@@ -77,7 +77,7 @@ to add a custom listener they will unintentionally override the behaviour we had
 
 ## Anything else
 
-Documentation with examples should be provided to remind users a listener should be added. 
+Documentation with examples should be provided to remind users that a listener should be added. 
 
 In the occasions where a listener is absolutely required we should take precautions to avoid 
 the two problems above, one course of action is a different event users can register without affection the default behaviour. For example, BpkChip could provide a `setOnSelectedStateChangeListener` so users can react to changes without replacing the click listener.
--- a/decisions/event-listeners-and-state.md
+++ b/decisions/event-listeners-and-state.md
@@ -72,7 +72,7 @@ In the example above `setOnClickListener` will be called by the super class befo
 
 **Flexibility:**
 
-Because with most listeners only one can be added (setListener vs addListener) if users need
+Because with most listeners only one can be added (`setListener` vs `addListener`) if users need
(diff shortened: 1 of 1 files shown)


decisions/event-listeners-and-state.md
```diff
@@ -0,0 +1,83 @@
...
+**Flexibility:**
+
+Because with most listeners only one can be added (setListener vs addListener) if users need
+to add a custom listener they will unintentionally override the behaviour we had defined for it. This can be avoided if we override the listener function and always add the behaviour, but this will make it impossible for users to disable this behaviour in certain ocasions.
+
+## Anything else
+
+Documentation with examples should be provided to remind users a listener should be added. 
+
+In the ocasions where a listener is absolutelly required we should take precautions to avoid
```
georgegillams:
```suggestion
In the occasions where a listener is absolutely required we should take precautions to avoid 
```

decisions/event-listeners-and-state.md
```diff
@@ -0,0 +1,83 @@
...
+In the example above `setOnClickListener` will be called by the super class before the props and init block have been called in the child class. This can be very hard to identify, especially from a user perspective.
+
+**Flexibility:**
+
+Because with most listeners only one can be added (setListener vs addListener) if users need
+to add a custom listener they will unintentionally override the behaviour we had defined for it. This can be avoided if we override the listener function and always add the behaviour, but this will make it impossible for users to disable this behaviour in certain ocasions.
+
+## Anything else
+
+Documentation with examples should be provided to remind users a listener should be added.
```
georgegillams:
```suggestion
Documentation with examples should be provided to remind users that a listener should be added. 
```

decisions/event-listeners-and-state.md
```diff
@@ -0,0 +1,83 @@
...
+
+Because with most listeners only one can be added (setListener vs addListener) if users need
+to add a custom listener they will unintentionally override the behaviour we had defined for it. This can be avoided if we override the listener function and always add the behaviour, but this will make it impossible for users to disable this behaviour in certain ocasions.
+
+## Anything else
+
+Documentation with examples should be provided to remind users a listener should be added. 
+
+In the ocasions where a listener is absolutelly required we should take precautions to avoid 
+the two problems above, one course of action is a different event users can register without affection the default behaviour. For example, BpkChip could provide a `setOnSelectedStateChangeListener` so users can react to changes without replacing the click listener.
```
georgegillams:
```suggestion
the two problems above, one course of action is a different event that users can register without affecting the default behaviour. For example, BpkChip could provide a `setOnSelectedStateChangeListener` so users can react to changes without replacing the click listener.
```

decisions/event-listeners-and-state.md
```diff
@@ -0,0 +1,83 @@
...
+}
+
+```
+
+In the example above `setOnClickListener` will be called by the super class before the props and init block have been called in the child class. This can be very hard to identify, especially from a user perspective.
+
+**Flexibility:**
+
+Because with most listeners only one can be added (setListener vs addListener) if users need
+to add a custom listener they will unintentionally override the behaviour we had defined for it. This can be avoided if we override the listener function and always add the behaviour, but this will make it impossible for users to disable this behaviour in certain ocasions.
```
georgegillams:
```suggestion
to add a custom listener they will unintentionally override the behaviour we had defined for it. This can be avoided if we override the listener function and always add the behaviour, but this will make it impossible for users to disable this behaviour in certain occasions.
```

tiagoengel:
All addressed @georgegillams. Thanks for the suggestions :)

decisions/event-listeners-and-state.md
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+```
+
+In the example above `setOnClickListener` will be called by the super class before the props and init block have been called in the child class. This can be very hard to identify, especially from a user perspective.
+
+**Flexibility:**
+
+Because with most listeners only one can be added (setListener vs addListener) if users need
```
georgegillams:
```suggestion
Because with most listeners only one can be added (`setListener` vs `addListener`) if users need
```

georgegillams:
```suggestion
Because with most listeners only one can be added (`setListener` vs `addListener`) if users need
```

decisions/event-listeners-and-state.md
```diff
@@ -0,0 +1,83 @@
...
+ 
+  override fun setOnClickListener(l: Listener) {
+    // 2
+    super.setOnClickListener(WrapWithAnalitics(l, myAnalyticProp))
+  }
+}
+
+```
+
+In the example above `setOnClickListener` will be called by the super class before the props and init block have been called in the child class. This can be very hard to identify, especially from a user perspective.
```
georgegillams:
```suggestion
In the example above `setOnClickListener` will be called by the super class before the props and `init block` have been called in the child class. This can be very hard to identify, especially from a user perspective.
```

georgegillams:

`com.github.Skyscanner:backpack-android:a286a56ad145215968b6f7dac81fca39de44341c`

PR #211, opened 2019-03-18
No jira spelling

shaundon:


package.json
```diff
@@ -19,10 +19,12 @@
     "release": "node release.js",
     "lint": "eslint . --ext .js,.jsx",
     "prebuild": "npm run lint",
-    "build": "gulp",
+    "build": "gulp && npm run spellcheck",
```
georgegillams: Should this be part of `test` instead?

shaundon: Well, we don't actually have a `test` script (because it's all done on the iOS side) and it felt silly to add one that'll just call one command. Plus it might incorrectly give people the impression there are JS tests in here.

PR #212, opened 2019-03-18
BPK 2138 calendar single date

valermor:
Single day example calendar

This PR introduces single day support to `BpkCalendar`.
Notice that the example Default and Coloured calendars in the project will show a radio switcher on the top to select the selection type. 
This switcher is not visible by default.
Moreover, the selection type needs to be set programmatically in the controller in the real usage.


Backpack/src/main/java/net/skyscanner/backpack/calendar/BpkCalendar.kt
```diff
@@ -17,11 +23,35 @@ open class BpkCalendar @JvmOverloads constructor(
   defStyle: Int = 0
 ) : ConstraintLayout(context, attrs, defStyle), OnYearChangedListener {
 
+  private var calendarController: BpkCalendarController? = null
+
+  var isSelectionTypeSwitcherVisible: Boolean = false
```
tiagoengel: I don't think this should live here, this code is only used for testing purposes so we should move it to the calendar stories inside the example app.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -74,14 +91,26 @@ abstract class BpkCalendarController {
...
+  }
+
   fun updateContent() = updateContentCallback?.updateContent()
 }
 
 internal fun Calendar.toCalendarDay() = CalendarDay(year = get(Calendar.YEAR), month = get(Calendar.MONTH), day = get(Calendar.DAY_OF_MONTH))
+
+enum class SelectionType {
+  RANGE,
+  SINGLE_DAY
```
tiagoengel: nit: can we change this to just `SINGLE`

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -27,15 +27,32 @@ abstract class BpkCalendarController {
 
   abstract val locale: Locale
 
+  var selectionType: SelectionType = SelectionType.RANGE
+
   abstract fun onRangeSelected(range: CalendarRange)
 
+  abstract fun onSingleDaySelected(day: CalendarDay)
```
tiagoengel: I think a better and more scalable approach would be to make the selection abstract and keep only one callback. We can keep the name `onRangeSelected` to keep it backwards compatible and avoid a breaking change. 

My idea would be something like this:
```kotlin
abstract fun onRangeSelected(selection: CalendarSelection) {
    if (selection is CalendarRange) {
       // handles range
    } else if (selection is SingleDay) {
       // handle single selection
    }
}
```

This has three main benefits:
1. Doesn't need to be breaking (we just need to keep the name)
2. We don't need to keep adding more methods if a new selections are introduced (and introducing more breaking changes)
3. When subclassing this the user will be able to just care about the mode they are using. E.g. someone only using single selection can do this:

```kotlin
override fun onRangeSelected(range: CalendarRange) {
   // handle range
}
```

whereas someone using only single date can do this:

```kotlin
override fun onRangeSelected(day: CalendarDay) {
   // handle day
}
```

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -27,15 +27,32 @@ abstract class BpkCalendarController {
 
   abstract val locale: Locale
 
+  var selectionType: SelectionType = SelectionType.RANGE
```
tiagoengel: I think there's a potential bug here. We probably need to clear the selection if the selection type changes, otherwise it will keep a range selected even after the type was set to `SINGLE`. Maybe a better approach would be to make this immutable and a constructor argument and if users need a different selection they can instantiate and set a new controller (which will reset the selection)

Backpack/src/main/res/layout/bpk_calendar_selection_type.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tiagoengel: This should also live only in the example app

Backpack/src/main/res/layout/view_bpk_calendar.xml
```diff
@@ -4,13 +4,26 @@
   android:layout_width="match_parent"
   android:layout_height="match_parent">
 
+  <include layout="@layout/bpk_calendar_selection_type"
```
tiagoengel: This should also live in the example app

app/src/main/java/net/skyscanner/backpack/demo/stories/ColoredCalendarStory.kt
```diff
@@ -14,6 +14,7 @@ class ColoredCalendarStory : Story() {
 
     val controller = ExampleBpkCalendarController(requireContext())
     val bpkCalendar = view.findViewById<BpkCalendar>(R.id.bpkCalendar)
+    bpkCalendar.isSelectionTypeSwitcherVisible = true
```
tiagoengel: Instead of this line we can replace it with adding a switcher in the example page and here adding a click/change listener that changes the selection/recreate the controller

app/src/main/java/net/skyscanner/backpack/demo/stories/DefaultCalendarStory.kt
```diff
@@ -11,7 +11,9 @@ class DefaultCalendarStory : Story() {
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     super.onViewCreated(view, savedInstanceState)
 
-    view.findViewById<BpkCalendar>(R.id.bpkCalendar).setController(ExampleBpkCalendarController(requireContext()))
+    val calendar = view.findViewById<BpkCalendar>(R.id.bpkCalendar)
+    calendar.isSelectionTypeSwitcherVisible = true
```
tiagoengel: Same as above

docs/Calendar/README.md
```diff
@@ -6,9 +6,10 @@ Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/b
 
 ## Usage
 
-BpkCalendar is an building on a fork of `CalendarView` and `MonthView` from the Android Open Source Project.
+BpkCalendar is building on a fork of `CalendarView` and `MonthView` from the Android Open Source Project.
```
tiagoengel: nit: is based on `CalendarView` and ..

docs/Calendar/README.md
```diff
@@ -6,9 +6,10 @@ Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/b
...
 
 The Calendar component can be used in both XML and Kotlin, but it currently requires a `BpkCalendarController` to be sub-classed and set.
+Both single day and data range are supported.
```
tiagoengel: nit: Both single and range selection are supported

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
valermor: If the intent is to avoid breaking changes into current classes extending `BpkCalendarController`, this won't work as on overriding `onRangeSelected()` the signature of the method needs to be kept, i.e. you need to pass a `CalendarSelection`.

This makes sense for me but just notice that the example at point 3 won't be possible.

tiagoengel: You're right, it would be possible to avoid breaking changes *after* this but it will be breaking now.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
valermor: Not sure I am getting this.
`selectionType` gets initialised here as a `RANGE` as I won't need any change in already existing code. On normal usage, it will be initialised once only.
For the purpose of being of exposing a switcher in the sample calendars this is mutable.
If I make this immutable, no switcher would be possible. This would mean creating maybe 2 other `Story` with single day selection.
I am fine either ways. But as iOS has this sort of switcher I thought I would provide it.

valermor: Or you mean re-instatiating the controller every time the selection is made?

app/src/main/java/net/skyscanner/backpack/demo/stories/ColoredCalendarStory.kt
valermor: I got this now. 
I am fine with that.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
tiagoengel: Yes I meant re-instantiating the controller to change the selection type and making it immutable, so something like:
```kotlin
view.controller = Controller(context) // Default is range
// latter on if selection type changes
view.controller = Controller(context, selectionType)
````
The problem right now is that if the controller is instantiated with `RANGE` and then latter one is changed to `SINGLE` a range will still be selected, which may not be a big problem but to be consistent we should reset the selected dates when the selection type changes.

File: Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt (hunk 3 above)
Comment (valermor) [reply Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt]: 👍

File: Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt (hunk 4 above)
Comment (valermor) [reply Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt]: 👍

**Comment by valermor** [comment]
@tiagoengel hopefully all the comments have been addressed.
The only thing I see it would be missing is updating https://backpack.github.io/components/calendar?platform=android to document single day selection. Let me know if there is any other thing I am supposed to do.

File: Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt [hunk 11]
```diff
@@ -74,14 +91,26 @@ abstract class BpkCalendarController {
     return CalendarDay(year, month, day).date == CalendarDay.today().date
   }
 
-  fun updateSelection(range: CalendarRange) {
+  fun updateSelectionForRange(range: CalendarRange) {
```
tiagoengel: Can we keep just one function here too `updateSelection(selection: CalendarSelection)`? So it can be called like this `updateSelection(CalendarRange())` or `updateSelection(CalendarDay())`

app/src/main/java/net/skyscanner/backpack/demo/stories/ColoredCalendarStory.kt
```diff
@@ -2,26 +2,55 @@ package net.skyscanner.backpack.demo.stories
...
+    single.text = "Single"
+    range.text = "Range"
+    range.isChecked = true
+
+    selection_type.visibility = View.VISIBLE
+
+    (selection_type as? RadioGroup)?.setOnCheckedChangeListener { _, checkedId ->
+      when (checkedId) {
+        R.id.single -> {
+          controller = ExampleBpkCalendarController(requireContext(), SelectionType.SINGLE)
```
tiagoengel: I think you need to call `bpkCalendar.setController(controller)` here again no?

app/src/main/java/net/skyscanner/backpack/demo/stories/DefaultCalendarStory.kt
```diff
@@ -2,16 +2,47 @@ package net.skyscanner.backpack.demo.stories
...
+    single.text = "Single"
+    range.text = "Range"
+    range.isChecked = true
+
+    selection_type.visibility = View.VISIBLE
+
+    (selection_type as? RadioGroup)?.setOnCheckedChangeListener { _, checkedId ->
+      when (checkedId) {
+        R.id.single -> {
+          controller = ExampleBpkCalendarController(requireContext(), SelectionType.SINGLE)
```
tiagoengel: Same here

docs/Calendar/README.md
```diff
@@ -31,9 +32,16 @@ BpkCalendar(context)
...
 
 ```Kotlin
-class ExampleBpkCalendarController(private val context: Context) : BpkCalendarController() {
-  override fun onRangeSelected(range: CalendarRange) {
-    // Do something with the selected range
+class ExampleBpkCalendarController(
+  private val context: Context,
+  override val selectionType: SelectionType
+ ) : BpkCalendarController(selectionType) {
+  override fun onRangeSelected(range: CalendarSelection) {
```
tiagoengel: Thinking of it now I think we should probably change it `onSelected` but we don't need to do it now, we can deprecate this in the future and rename it.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
valermor: I am fine with the change. Notice these methods are only used for testing at the moment.

app/src/main/java/net/skyscanner/backpack/demo/stories/ColoredCalendarStory.kt
valermor: It is called at the bottom of the `when` block. Line 50.

app/src/main/java/net/skyscanner/backpack/demo/stories/DefaultCalendarStory.kt
valermor: Same comment as above. Line 44.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
tiagoengel: It's also used in the RN bridge we have, so changing the name will break it now.

valermor: But keeping the name while changing the signature will be breaking anyway.
I am going to change it no issue.

PR #215, opened 2019-03-20
[BPK-2340] Fix year pill for RN

tiagoengel:
Current version doesn't work with the RN bridge. This also fixes a bug when the start date is in the past/future.


UNRELEASED.md
```diff
@@ -1 +1,5 @@
 # Unreleased
+
+**Fixed:**
+
+- Fix calendar year pill the react native bridge
```
antalindisguise: Could you give some context what was fixed?

Backpack/src/main/java/net/skyscanner/backpack/calendar/BpkCalendar.kt
```diff
@@ -27,11 +27,17 @@ open class BpkCalendar @JvmOverloads constructor(
...
     calendar_view.listener = this
+
+    updateYearPill(controller.startDate.year)
   }
 
   override fun onYearChanged(year: Int) {
+    this.post { updateYearPill(year) }
+  }
+
+  private fun updateYearPill(year: Int) {
```
antalindisguise: Why is this delegated to a new thread? Does not seem to be heavy logic and it does update UI

UNRELEASED.md
tiagoengel: The year pill doesn't show in the react native otherwise. Why exactly is hard to tell, it probably has to do with how things work over the bridge (js code running in a different thread). We had to do a similar thing in `CaldendarView.requestLayout` and that was based in the react native source code (https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPicker.java#L75).

Backpack/src/main/java/net/skyscanner/backpack/calendar/BpkCalendar.kt
tiagoengel: Same reason as above, the logic to set the text was running on a new thread already, because it wasn't working otherwise, I just moved the whole thing to a thread (as it doesn't work otherwise in RN). I suspect that because this is called inside a scroll event we need to wait for it to finish to show the pill, but I'm not sure, might be something completely different.

PR #222, opened 2019-03-27
BPKCalendar wrap content height bugfix

reneargento:
This PR fixes a bug that exists when BPKCalendar is set to a wrap_content height.
Now it should work both with match_parent and wrap_content heights.
The PR also simplifies the layout for the calendar in the sample app. 


tiagoengel:
Thanks for the contribution @reneargento. Can you please add another example using "wrap_content"? Can be just a copy and paste of the default but using wrap_content. Can you also please add an entry to `UNRELEASED.md`? 😇

reneargento:
Thanks for the review @tiagoengel, the issue should be fixed on the latest commit.

PR #226, opened 2019-04-01
Replacing java.util.* with JSR-310 Date constructs

antalindisguise:


Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -82,13 +76,10 @@ abstract class BpkCalendarController(
...
-  // Shall we use DateFormat.getDateInstance instead?
-  internal fun getLocalizedDate(date: Date, pattern: String): String = SimpleDateFormat(pattern, locale).format(date)
+  internal fun getLocalizedDate(date: LocalDate, pattern: String): String = DateTimeFormatter.ofPattern(pattern, locale).format(date)
 
   open fun isToday(year: Int, month: Int, day: Int): Boolean {
-    return CalendarDay(year, month, day).date == CalendarDay.today().date
+    return LocalDate.of(year, month, day) == LocalDate.now()
   }
 
   fun updateSelection(selection: CalendarSelection) {
```
tiagoengel: Not sure why this is happening only now, but `selectedRange.end` should be null here too when the selection is a `SelectionDay`. This is causing the snapshots to fail.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/WeekdayHeaderView.kt
```diff
@@ -30,25 +31,21 @@ internal class WeekdayHeaderView @JvmOverloads constructor(
...
-        1 -> secondWeekdayView.text = calendarDayName
-        2 -> thirdWeekdayView.text = calendarDayName
-        3 -> fourthWeekdayView.text = calendarDayName
-        4 -> fifthWeekdayView.text = calendarDayName
-        5 -> sixthWeekdayView.text = calendarDayName
-        6 -> seventhWeekdayView.text = calendarDayName
-      }
-    }
+    val formatter = DateTimeFormatter.ofPattern("ccc", locale)
+    dayOfWeek(locale, formatter, 1)
```
tiagoengel: What is this line for?

Backpack/src/main/java/net/skyscanner/backpack/calendar/BpkCalendar.kt
```diff
@@ -36,8 +36,8 @@ open class BpkCalendar @JvmOverloads constructor(
   }
 
   private fun updateYearPill(year: Int) {
-    val currentYear = Calendar.getInstance().get(Calendar.YEAR)
-    year_pill_view.message = year.toString()
-    year_pill_view.visibility = if (currentYear == year) View.GONE else View.VISIBLE
+    val yearCalendar = LocalDate.of(year, 1, 1)
```
tiagoengel: This should be `LocalDate.now()`

app/src/main/java/net/skyscanner/backpack/demo/data/ExampleBpkCalendarController.kt
```diff
@@ -51,20 +49,17 @@ class ExampleBpkCalendarController(
...
-  val greenSet = mutableSetOf<CalendarDay>()
-  val greySet = mutableSetOf<CalendarDay>()
-  val emptySet = mutableSetOf<CalendarDay>()
+  val daysBetweenStartAndEnd = ChronoUnit.DAYS.between(startDate, endDate) + 1
+  val redSet = mutableSetOf<LocalDate>()
+  val yellowSet = mutableSetOf<LocalDate>()
+  val greenSet = mutableSetOf<LocalDate>()
+  val greySet = mutableSetOf<LocalDate>()
+  val emptySet = mutableSetOf<LocalDate>()
+  var dateIterator = LocalDate.now()
```
tiagoengel: This should be ` var dateIterator = LocalDate.of(startDate.year, startDate.month, startDate.dayOfMonth)
`

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
antalindisguise: I can't seem to reproduce this. I don't think it's related to this PR

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/WeekdayHeaderView.kt
antalindisguise: Should have been removed, just a leftover. Cleaning it up now.

Backpack/src/main/java/net/skyscanner/backpack/calendar/BpkCalendar.kt
antalindisguise: Actually, no need to keep track of month, day here so just going to use `YearMonth.now()` instead and simply assign the parameters `year: Int` as the message

app/src/main/java/net/skyscanner/backpack/demo/data/ExampleBpkCalendarController.kt
antalindisguise: Good catch. Fixed now

antalindisguise:
There was an issue where we did not calculate the correct offset on the first visible day in the current month (good spotting @tiagoengel). This is fixed now by correctly checking the number of non visible days and the first visible locale-aware day of the week and the offset needed.

PR #229, opened 2019-04-01
[NO-JIRA] Correctly run snapshots before release

tiagoengel:
This also fixes a small bug in the calendar (that was happening because we were not running snapshots before release)

This ensures we run all snapshot tests before release. We run them on CI too with the exception of `FlakyTests` which fail very often because of timeouts. Those we still want to check locally before releasing.

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
@@ -100,7 +100,7 @@ abstract class BpkCalendarController(
       }
       is SingleDay -> {
         selectedRange.start = selection.selectedDay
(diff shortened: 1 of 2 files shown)


georgegillams:

`com.github.Skyscanner:backpack-android:94a3a94e125af3405573025baf933115a01746f3`

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/BpkCalendarController.kt
```diff
@@ -100,7 +100,7 @@ abstract class BpkCalendarController(
       }
       is SingleDay -> {
         selectedRange.start = selection.selectedDay
-        selectedRange.end = selection.selectedDay
```
antalindisguise: 👍 good spot

PR #230, opened 2019-04-02
Add ThreeTenABP dependency to our calendar instead ThreeTenBP

antalindisguise:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,5 +67,5 @@ ext {
   androidMapsUtilsVersion = "0.5+"
   androidx = "1.0.0"
   constraintLayout = "1.1.3"
-  threetenbp = "1.3.8"
+  threeTenBp = "1.3.8"
 }
--- a/build.gradle
+++ b/build.gradle
@@ -67,5 +67,5 @@ ext {
   androidMapsUtilsVersion = "0.5+"
   androidx = "1.0.0"
   constraintLayout = "1.1.3"
-  threeTenBp = "1.3.8"
+  threeTenAbp = "1.1.2"
 }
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -67,5 +67,5 @@ ext {
   androidMapsUtilsVersion = "0.5+"
   androidx = "1.0.0"
   constraintLayout = "1.1.3"
-  threetenbp = "1.3.8"
+  threeTenBp = "1.3.8"
```
tiagoengel: Shouldn't we be 1.1.2?

docs/Calendar/README.md
```diff
@@ -4,9 +4,15 @@
 
 Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/backpack-android). Check the main [Readme](https://github.com/skyscanner/backpack-android#installation) for a complete installation guide.
 
+Additionally, if you want to handle the output of the calendar (the dates), you also need to rely on the Android adaptation of the JSR-310 backport: [ThreeTenABP](https://github.com/JakeWharton/ThreeTenABP). This is, because we decided to rely on `org.threetenbp.bp.*` constructs as our business entities inside the calendar.
```
tiagoengel: I suggest we remove this and don't make the lib a `compileOnly` dependency and mention on the usage part how to initialise the lib. Doesn't make much sense to me not to include a lib we expose as part of the `BpkCalendar` public api.

I would also add the code a snippet with `AndroidThreeTen.init(context)` there to showing how to it (plus the link to the full doc)

antalindisguise: Updated the documentation accordingly

build.gradle
antalindisguise: Well, this code is now outdated. It used to use the ThreeTenBP instead of the ThreeTenABP which was in a higher version. Should be fine now.

PR #234, opened 2019-04-04
[BPK-2333] Add theme to chip

tiagoengel:
Screenshot_1554455947
Screenshot_1554455971

### DIFF
--- a/app/src/main/res/values/themes.xml
+++ b/app/src/main/res/values/themes.xml
@@ -11,6 +11,7 @@
     <item name="bpkSwitchStyle">@style/LondonThemeBpkSwitch</item>
     <item name="bpkButtonPrimaryStyle">@style/LondonThemeButtonPrimary</item>
     <item name="bpkButtonSecondaryStyle">@style/LondonThemeButtonSecondary</item>
+    <item name="bpkChipStyle">@style/LondonThemeChip</item>
     <item name="bpkSpinnerPrimaryStyle">@style/LondonThemeSpinnerPrimary</item>
     <item name="bpkTextStyle">@style/LondonThemeText</item>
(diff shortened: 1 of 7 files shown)


georgegillams:

`com.github.Skyscanner:backpack-android:b10e55b5318efb59acb26cbdbb5a94407bfc8c2c`

app/src/main/res/values/themes.xml
```diff
@@ -73,7 +82,8 @@
   <style name="DohaThemeButtonPrimary">
     <item name="buttonBackgroundColor">#ffb300</item>
     <item name="buttonTextColor">@color/bpkWhite</item>
-  </style>
+  </style
```
amolgupta: does the lint plugin not catch this?

app/src/main/res/values/themes.xml
```diff
@@ -36,15 +37,22 @@
     <item name="buttonStrokeColor">#5E072C</item>
   </style>
 
+  <style name="LondonThemeChip">
+    <item name="chipSelectedBackgroundColor">?colorPrimary</item>
```
amolgupta: we can call it only `chipColor` as there are no other attributes

app/src/main/res/values/themes.xml
tiagoengel: AFAIL it doesn't lint xml

app/src/main/res/values/themes.xml
tiagoengel: We follow this pattern everywhere else so it's better we keep it consistent, see: https://github.com/Skyscanner/backpack/blob/master/decisions/theme-props-naming.md

PR #243, opened 2019-04-16
[BPK-2390] Disable ViewPump on Android Q

amolgupta:
Disable `ViewPump` on Android-Q for now. This means the highlights can still be used on emulators and devices running a stable version of Android upto API-28

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
@@ -1,0 +1,19 @@
+@file:JvmName("BpkViewPumpContextWrapper")
+
+package net.skyscanner.backpack.util
+
+import android.content.Context
+import android.content.ContextWrapper
+import io.github.inflationx.viewpump.ViewPumpContextWrapper
+
+class BpkViewPumpContextWrapper {
+  companion object {
+    fun wrap(context: Context): ContextWrapper {
+      return if (android.os.Build.VERSION.SDK_INT > android.os.Build.VERSION_CODES.P) {
+        ContextWrapper(context)
+      } else {
+        ViewPumpContextWrapper.wrap(context)
+      }
+    }
+  }
+}
--- a/Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
@@ -5,11 +5,15 @@ import android.content.ContextWrapper
 import io.github.inflationx.viewpump.ViewPumpContextWrapper
 
 object BpkViewPumpContextWrapper {
-    fun wrap(context: Context): ContextWrapper {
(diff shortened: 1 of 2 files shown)


Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
```diff
@@ -0,0 +1,19 @@
+@file:JvmName("BpkViewPumpContextWrapper")
```
antalindisguise: Why is this needed?

Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package net.skyscanner.backpack.util
+
+import android.content.Context
+import android.content.ContextWrapper
+import io.github.inflationx.viewpump.ViewPumpContextWrapper
+
+class BpkViewPumpContextWrapper {
+  companion object {
```
antalindisguise: Sound like you don't need any state in this class stored but only using it to call a static function -- why not use an `object` instead?

georgegillams:

`com.github.Skyscanner:backpack-android:62adce2ba99c0816d8dcf73c03a58e2c33397dba`

georgegillams:

`com.github.Skyscanner:backpack-android:11530260987b0a91f65877492e7a250f2bbcb3fb`

Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
```diff
@@ -0,0 +1,15 @@
+package net.skyscanner.backpack.util
+
+import android.content.Context
+import android.content.ContextWrapper
+import io.github.inflationx.viewpump.ViewPumpContextWrapper
+
+object BpkViewPumpContextWrapper {
+    fun wrap(context: Context): ContextWrapper {
+      return if (android.os.Build.VERSION.SDK_INT > android.os.Build.VERSION_CODES.P) {
+        ContextWrapper(context)
```
tiagoengel: You can change the return to be `Context`  and just return `context` here instead of wrapping. A `ContextWrapper` is also a context.

Backpack/src/main/java/net/skyscanner/backpack/util/BpkViewPumpContextWrapper.kt
```diff
@@ -0,0 +1,19 @@
...
+import io.github.inflationx.viewpump.ViewPumpContextWrapper
+
+object BpkViewPumpContextWrapper {
+  /**
+   * Disables Viewpump on Android Q and above.
+   * See https://github.com/InflationX/ViewPump/issues/21
+   */
+  fun wrap(context: Context): Context {
+    return if (android.os.Build.VERSION.SDK_INT > android.os.Build.VERSION_CODES.P) {
+      ContextWrapper(context)
```
tiagoengel: Sorry I meant returning the context directly here without wrapping, e.g. just `context` instead of `ContextWrapper`

PR #255, opened 2019-05-14
New one test fail ci

georgegillams:
Testing that CI will fail if test snapshots have changed...


tiagoengel:
Thanks for this! I hope it fails 😅

georgegillams:
This failed as expected 💯
https://www.reddit.com/r/brandnewsentence

PR #263, opened 2019-05-22
[NO-JIRA] Change switch theming approach

tiagoengel:
Previous approach causes problems/confusion when our components are extended and we have moved away from it in other components. Switch was the only component still using this approach.


georgegillams:

`com.github.Skyscanner:backpack-android:614f1f2f92100f1b76b7a630a8023f447143633b`

PR #284, opened 2019-06-22
[No Jira] Check if StateListAnimator is supported

pietmau:
Need to check that api is at least 22 (LOLLIPOP_MR1) ☹️


Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
```diff
@@ -0,0 +1,17 @@
+package net.skyscanner.backpack.util
+
+import android.animation.AnimatorInflater
+import android.os.Build
+import android.view.View
+import androidx.annotation.DrawableRes
+
+fun View.isStateListAnimatorSupported() =
```
bvitaliyg: I don't think we need to make it public.

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
```diff
@@ -0,0 +1,17 @@
...
+  if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.LOLLIPOP) {
+    false
+  } else isSpecificDeviceSupported()
+
+private fun isSpecificDeviceSupported() = Build.MODEL != "GT-I9505"
```
bvitaliyg: We need to check manufacturer as well

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
```diff
@@ -0,0 +1,17 @@
...
+
+fun View.loadStateListAnimator(@DrawableRes animator: Int) {
+  stateListAnimator = AnimatorInflater.loadStateListAnimator(context, animator)
```
bvitaliyg: It's not clear what stateListAnimator is. Please add this prefix.

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
```diff
@@ -0,0 +1,17 @@
...
+import androidx.annotation.DrawableRes
+
+fun View.isStateListAnimatorSupported() =
+  if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.LOLLIPOP) {
+    false
+  } else isSpecificDeviceSupported()
+
+private fun isSpecificDeviceSupported() = Build.MODEL != "GT-I9505"
+
+fun View.loadStateListAnimator(@DrawableRes animator: Int) {
```
bvitaliyg: Also internal. Plus it's worth consideration to move all these methods to BpkButton

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
pietmau: I agree, internal is better

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
pietmau: yeah

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
pietmau: yup

Backpack/src/main/java/net/skyscanner/backpack/util/extensions.kt
pietmau: as discussed

tiagoengel:
closed in favour of #285

PR #291, opened 2019-06-28
Bpk 2643 dev android add checkbox component

bvitaliyg:


Backpack/src/main/java/net/skyscanner/backpack/text/BpkPrimaryColorSpan.kt
```diff
@@ -0,0 +1,15 @@
+package net.skyscanner.backpack.text
+
+import android.content.Context
+import android.text.TextPaint
+import android.text.style.CharacterStyle
+import net.skyscanner.backpack.util.BpkTheme
+
+class BpkPrimaryColorSpan(context: Context) : CharacterStyle() {
```
tiagoengel: I'm wondering if we shouldn't extend the from the font span here? It seems to me we should enforce the bpk font with this span too. What do you think?

bvitaliyg: I'd prefer to make a single responsibility for each span for now. And far as I know, android spans follow the same principle. It's usually not a big issue to add extra span to the same part of the string. Besides that would make this class too heavy as it requires font params for creation.

tiagoengel: 👍

PR #300, opened 2019-07-11
Bpk 2707 dev android add user interaction listener to interactive star rating

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
@@ -38,7 +38,10 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
     val x = if (layoutDirection == View.LAYOUT_DIRECTION_RTL) width - ev.x else ev.x
     val itemWidth = width / maxRating
     val selectedItems = x / itemWidth
-    rating = Math.max(1f, selectedItems + 0.5f)
+    super.rating = Math.round(Math.max(1f, selectedItems + 0.5f)).toFloat()
+    if (ev.action == MotionEvent.ACTION_UP || ev.action == MotionEvent.ACTION_CANCEL) {
+      onRatingChangedListener?.invoke(rating, maxRating.toFloat())
+    }
     return true
   }
 
--- a/app/src/main/java/net/skyscanner/backpack/demo/stories/InteractiveStarRatingStory.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/stories/InteractiveStarRatingStory.kt
@@ -1,18 +1,25 @@
 package net.skyscanner.backpack.demo.stories
 
 import android.os.Bundle
-import android.util.Log
 import android.view.View
-import net.skyscanner.backpack.demo.R
+import android.view.ViewGroup
 import net.skyscanner.backpack.starrating.BpkInteractiveStarRating
+import net.skyscanner.backpack.toast.BpkToast
 
 class InteractiveStarRatingStory : Story() {
 
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     super.onViewCreated(view, savedInstanceState)
 
-    view.findViewById<BpkInteractiveStarRating>(R.id.star_rating_empty).onRatingChangedListener = { current, max ->
-      Log.d("InteractiveStarRating", "$current/$max")
+    view as ViewGroup
+
+    (0 until view.childCount).forEach { idx ->
+      val child = view.getChildAt(idx)
+      if (child is BpkInteractiveStarRating) {
+        child.onRatingChangedListener = { current, max ->
+          BpkToast.makeText(context, "$current/$max", BpkToast.LENGTH_SHORT).show()
+        }
+      }
     }
   }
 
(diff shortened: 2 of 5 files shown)


Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
```diff
@@ -38,7 +38,10 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
     val x = if (layoutDirection == View.LAYOUT_DIRECTION_RTL) width - ev.x else ev.x
     val itemWidth = width / maxRating
     val selectedItems = x / itemWidth
-    rating = Math.max(1f, selectedItems + 0.5f)
+    super.rating = Math.round(Math.max(1f, selectedItems + 0.5f)).toFloat()
```
tiagoengel: Why are we adding `Math.round` here?

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
```diff
@@ -38,7 +38,10 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
...
+    if (ev.action == MotionEvent.ACTION_UP || ev.action == MotionEvent.ACTION_CANCEL) {
```
tiagoengel: Why do we trigger for `ACTION_CANCEL` too? Seems like `ACTION_CANCEL` should cancel the interaction no?

app/src/main/java/net/skyscanner/backpack/demo/stories/InteractiveStarRatingStory.kt
```diff
@@ -0,0 +1,27 @@
...
+import android.util.Log
+import android.view.View
+import net.skyscanner.backpack.demo.R
+import net.skyscanner.backpack.starrating.BpkInteractiveStarRating
+
+class InteractiveStarRatingStory : Story() {
+
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
+
```
tiagoengel: Can we change to a toast and add to all of them? It's easier to see when your trying it out
```suggestion
    view as ViewGroup

    (0..(view.childCount -1)).forEach { idx ->
      val child = view.getChildAt(idx)
      if (child is BpkInteractiveStarRating) {
        child.onRatingChangedListener = { current, max ->
          BpkToast.makeText(context, "$current/$max", BpkToast.LENGTH_SHORT).show()
        }
      }
    }
```

app/src/main/java/net/skyscanner/backpack/demo/stories/InteractiveStarRatingStory.kt
```diff
@@ -0,0 +1,27 @@
...
+    view.findViewById<BpkInteractiveStarRating>(R.id.star_rating_empty).onRatingChangedListener = { current, max ->
```
tiagoengel:
```suggestion
    view.findViewById<BpkInteractiveStarRating>(R.id.star_rating_empty).onRatingChangedListener = { current, max ->
        view as ViewGroup

    (0..(view.childCount -1)).forEach { idx ->
      val child = view.getChildAt(idx)
      if (child is BpkInteractiveStarRating) {
        child.onRatingChangedListener = { current, max ->
          BpkToast.makeText(context, "$current/$max", BpkToast.LENGTH_SHORT).show()
        }
      }
    }
```

tiagoengel: Can we add it for all views and make it a toast instead of a log? It's much easier to test that way.

```
    view as ViewGroup

    (0..(view.childCount -1)).forEach { idx ->
      val child = view.getChildAt(idx)
      if (child is BpkInteractiveStarRating) {
        child.onRatingChangedListener = { current, max ->
          BpkToast.makeText(context, "$current/$max", BpkToast.LENGTH_SHORT).show()
        }
      }
    }
```

tiagoengel:
Tested local and it works well, just a few suggestions/questions

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
bvitaliyg: Interactive Star Rating supports only Integer values because that's what user sees on the screen. You cannot have half-star on interactive rating. Besides without rounding the callback will be receiving something like 3.33241, 3.7534 instead of 3 and 4.

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
bvitaliyg: Since the UI updates together with movement, on CANCEL it will remain the its last state. However callback won't be called unless you add this check.

app/src/main/java/net/skyscanner/backpack/demo/stories/InteractiveStarRatingStory.kt
bvitaliyg: Sounds like a good idea!

bvitaliyg: Done

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
tiagoengel: 👍

PR #301, opened 2019-07-11
[NO-JIRA] Fix white button text colour bug

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
@@ -3,7 +3,6 @@ package net.skyscanner.backpack.button.internal
 import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.Canvas
-import android.graphics.Color
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.VectorDrawable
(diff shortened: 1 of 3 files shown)


Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
```diff
@@ -95,8 +94,10 @@ abstract class BpkButtonBase internal constructor(
 
     this.context.theme.obtainStyledAttributes(attrs, R.styleable.BpkButton, defStyleAttr, 0)
       ?.use {
-        it.getColor(R.styleable.BpkButton_buttonTextColor, INVALID_RES).let { res ->
-          if (res != INVALID_RES) {
+        it.getColor(R.styleable.BpkButton_buttonTextColor, Color.TRANSPARENT).let { res ->
+          // We use Color.TRANSPARENT here because -1 (INVALID_RES) is also used to represent the white color,
+          // which should be a valid colour here.
```
bvitaliyg: Please add TODO to remove the hack in the future.

Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
```diff
@@ -95,8 +94,10 @@ abstract class BpkButtonBase internal constructor(
...
+          if (res != Color.TRANSPARENT) {
```
bvitaliyg: We can use getColorStateList() instead, which returns null if the attribute is not defined. From there we can call getDefaultColor().

tiagoengel: That's another hack though, I'd also need to add a comment explaining why I'm doing that. Can you elaborate why that is a better hack than this?

bvitaliyg: ColorInt is a set of colours covering all Integer range. There is no way to get a color without having any specific color broken. In your case I cannot set transparent color. The behaviour will be confusing as well: instead of transparent colour I'll see default one. I agree that is a rare case, but it's still there. The solution is not oblivious as well, and that's why you added a comment.

Approach with ColorStateList wraps all the values into an object, and therefore provides a way to cover all colors range without exceptions. It also uses standart API  as it was designed so you don't need to have a comment anymore.

tiagoengel: Transparent is not supported for the text, there is not use case where we want people to set the button colour to transparent, and the colour state list is still a hack. We are using sate list just to get the default colour, it's not clear why unless I add a comment explaining it. 
I'll change to that but are still changing one hack for another.

bvitaliyg: A lot of component are retrieving the color via ColorStateList. It's even preferable way to do it since it supports multiple states and adds ability to change color when pressed, for example. Default Color means we don't support the dynamic change yet, but there's nothing really platform unusual for such pattern. Transparent just means that we change one buggy colour to another which is very unlikely to be used, but the bug is still there.

PR #307, opened 2019-07-22
Bpk 2706 expose text appearances xml

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/androidTest/java/net/skyscanner/backpack/text/BpkTextTest.kt
```diff
@@ -182,8 +148,16 @@ class BpkTextTest {
...
 
-      Assert.assertTrue(message, font.isCustomFont)
-      Assert.assertEquals(message, ResourcesCompat.getFont(withCustomFont, net.skyscanner.backpack.test.R.font.shadows_into_light), font.typeface)
+      val expectedFont = when {
+        message.contains("Emphasized") ->
+          Typeface.create("cursive", Typeface.NORMAL)
+        message.contains("Heavy") ->
+          Typeface.create("casual", Typeface.NORMAL)
+        else ->
+          ResourcesCompat.getFont(withCustomFont, net.skyscanner.backpack.test.R.font.shadows_into_light)
```
bvitaliyg: Can we do full import here?

Backpack/src/androidTest/java/net/skyscanner/backpack/text/internal/FontFamilyResolverTest.kt
```diff
@@ -0,0 +1,61 @@
...
+
+  @Test
+  fun test_default_font() {
+    val subject = FontFamilyResolver(BpkTheme.wrapContextWithDefaults(context))
+
+    val serif = Typeface.create("sans-serif", Typeface.NORMAL)
+    val serifMedium = Typeface.create("sans-serif-medium", Typeface.NORMAL)
+    val serifBlack = Typeface.create("sans-serif-black", Typeface.NORMAL)
+
+    Assert.assertEquals(serif, subject.getForWeight(BpkText.Weight.NORMAL))
```
bvitaliyg: I'd spilled it into several tests.

Backpack/src/androidTest/java/net/skyscanner/backpack/text/internal/FontFamilyResolverTest.kt
```diff
@@ -0,0 +1,61 @@
...
+
+  @Test
+  fun test_custom_font() {
+    val subject = FontFamilyResolver(ContextThemeWrapper(context, R.style.TestTextCustomFont))
+
+    val shadows = FontCache[R.font.shadows_into_light, context]
+    val cursive = Typeface.create("cursive", Typeface.NORMAL)
+    val casual = Typeface.create("casual", Typeface.NORMAL)
+
+    Assert.assertEquals(shadows, subject.getForWeight(BpkText.Weight.NORMAL))
```
bvitaliyg: As well as this

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -105,55 +87,42 @@ open class BpkText @JvmOverloads constructor(
...
-    val weightArg = a.getInt(R.styleable.BpkText_weight, -1)
-    if (weightArg == -1) {
-      // if weight has not been set we still read the emphasize property
-      emphasize = a.getBoolean(R.styleable.BpkText_emphasize, false)
-    } else {
-      weight = Weight.values()[weightArg]
-    }
+      defStyleAttr, 0
+    ).use {
+      _textStyle = it.getInt(R.styleable.BpkText_textStyle, BASE)
```
bvitaliyg: Let's use a local variable instead of backing property. We can do something like:
```
val property = BASE
...use {
propery = it.getInt(styleable, property)
}
this.property = property;
```

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -105,55 +87,42 @@ open class BpkText @JvmOverloads constructor(
...
-
-    val drawableTint = a.getColorStateList(R.styleable.BpkText_drawableTint)
-    if (drawableTint != null) {
-      setDrawableTint(drawableTint.getColorForState(EMPTY_STATE_SET, Color.WHITE))
-    }
-
-    fontBase = a.getResourceId(R.styleable.BpkText_fontFamilyBase, -1)
-    fontEmphasized = a.getResourceId(R.styleable.BpkText_fontFamilyEmphasized, -1)
-    fontHeavy = a.getResourceId(R.styleable.BpkText_fontFamilyHeavy, -1)
+      if (this.layoutDirection == View.LAYOUT_DIRECTION_RTL) {
```
bvitaliyg: We probably need to move it to extension properties. Maybe not in the ticket

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -202,53 +161,34 @@ open class BpkText @JvmOverloads constructor(
...
   textStyleAttributes.recycle()
 
-  val fontFromTheme = internalGetFontFromTheme(context, weight)
-
-  if (fontFromTheme != null) {
-    return BpkText.FontDefinition(fontFromTheme, fontSize, null, isCustomFont = true)
-  }
-
-  return BpkText.FontDefinition(Typeface.create(fontFamily, Typeface.NORMAL), fontSize, letterSpacing)
+  return FontFamilyResolver(context).getForWeight(weight)?.let {
```
bvitaliyg: What will allocate a new object every time the BpkText is used. We need to make it static.

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.content.Context
+import android.graphics.Typeface
+import android.util.TypedValue
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.text.BpkText
+import net.skyscanner.backpack.text.FontCache
+
+class FontFamilyResolver(val context: Context) {
+
+  private val fontResources by lazy {
```
bvitaliyg: unsafe lazy.

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
```diff
@@ -0,0 +1,42 @@
+package net.skyscanner.backpack.text.internal
+
+import android.content.Context
+import android.graphics.Typeface
+import android.util.TypedValue
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.text.BpkText
+import net.skyscanner.backpack.text.FontCache
+
+class FontFamilyResolver(val context: Context) {
```
bvitaliyg: @UIThread

Backpack/src/androidTest/java/net/skyscanner/backpack/text/BpkTextTest.kt
tiagoengel: We already have another R import here, I'll add an alias

Backpack/src/androidTest/java/net/skyscanner/backpack/text/internal/FontFamilyResolverTest.kt
tiagoengel: I'll just add a message to the assert as this is really only testing one function, this will make it clear which on is which if the test fails

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
tiagoengel: This will trigger the `setup` function twice in the constructor, that's why I moved it

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
tiagoengel: Not sure it applies anymore, but why is that?

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -202,53 +161,34 @@ open class BpkText @JvmOverloads constructor(
...
 
-  val fontFromTheme = internalGetFontFromTheme(context, weight)
-
-  if (fontFromTheme != null) {
-    return BpkText.FontDefinition(fontFromTheme, fontSize, null, isCustomFont = true)
-  }
-
-  return BpkText.FontDefinition(Typeface.create(fontFamily, Typeface.NORMAL), fontSize, letterSpacing)
+  return FontFamilyResolver(context, weight)?.let {
+    BpkText.FontDefinition(it, fontSize, letterSpacing)
```
bvitaliyg: Can we avoid FontDefinition allocation here? Just to consider.

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.graphics.Typeface
+import android.util.TypedValue
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.text.BpkText
+import net.skyscanner.backpack.text.FontCache
+
+object FontFamilyResolver {
+
+  private fun getFontResources(context: Context): List<Typeface?> {
+    val fontAttributes = intArrayOf(
```
bvitaliyg: external val?

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
```diff
@@ -0,0 +1,42 @@
...
+      R.attr.bpkFontFamilyBase,
+      R.attr.bpkFontFamilyEmphasized,
+      R.attr.bpkFontFamilyHeavy
+    )
+
+    return fontAttributes.map {
+      val outValue = TypedValue()
```
bvitaliyg: External val?

Backpack/src/androidTest/java/net/skyscanner/backpack/text/BpkTextTest.kt
```diff
@@ -24,6 +26,7 @@ class BpkTextTest {
   @Before
   fun setUp() {
     context = InstrumentationRegistry.getInstrumentation().targetContext
+    BpkTheme.applyDefaultsToContext(context)
```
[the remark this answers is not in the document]
tiagoengel: This is how it works in the "real" environment (the app). All defaults are set before any widget is loaded as they are also available to use outside backpack components.

We should probably remove the `wrapWithDefaults` from all components and just make this call a requirement in the docs (which is the case already just not well documented)

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -202,53 +161,34 @@ open class BpkText @JvmOverloads constructor(
...
 @StyleRes
-private fun getStyleId(textStyle: Int, weight: BpkText.Weight): Int {
+private fun getStyleId(context: Context, textStyle: Int, weight: BpkText.Weight): Int {
   val styleProps = styleMapping[textStyle]
   styleProps ?: throw IllegalStateException("Invalid textStyle")
-  val textAppearance = styleProps[weight.ordinal]
-  textAppearance
+  val textAppearanceAttr = styleProps[weight.ordinal]
+
+  textAppearanceAttr
```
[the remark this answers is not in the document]
tiagoengel: Not really, the same function is used by the `getFont` function

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
bvitaliyg: It's an indication what the elements is not a thread-safe and should be used in UI thread only. Since we want to make it a singleton with unsafe static initializations it makes sense.

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
```diff
@@ -0,0 +1,42 @@
+package net.skyscanner.backpack.text.internal
+
+import android.content.Context
+import android.graphics.Typeface
+import android.util.TypedValue
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.text.BpkText
+import net.skyscanner.backpack.text.FontCache
+
+object FontFamilyResolver {
```
bvitaliyg: internal

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
tiagoengel: No this is returned in the `getFont` function. It's only used by that function now but it we need it to wrap all font props

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
tiagoengel: Should be internal 👍

Backpack/src/main/java/net/skyscanner/backpack/text/internal/FontFamilyResolver.kt
tiagoengel: 👍

PR #316, opened 2019-08-02
[BPK-2780] Removed forced RTl layouts.

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -146,8 +147,8 @@ object ComponentRegistry {
   private val TOKENS_MAP = mapOf(
     "All Icons" story NodeData({ children -> SubStory of children },
       mapOf(
-        "Default" story NodeData { IconsStory() },
-        "RTL" story NodeData { IconsStory of View.LAYOUT_DIRECTION_RTL }
+        "Default" story NodeData { IconsStory of R.layout.fragment_all_icons },
+        "RTL" story NodeData { IconsStory of R.layout.fragment_all_icons with Direction.RTL }
```
tiagoengel: That's not giving the same result, setting this to the whole view inverts the order of the items in the view so it's harder to compare this with the normal version. 

Can we keep this like it was before?

PR #324, opened 2019-08-22
[BPK-2817] Implemented the component, added samples.

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -117,6 +117,7 @@ object ComponentRegistry {
       )),
     "Horizontal Nav" story NodeData { HorizontalNavStory of R.layout.fragment_horizontal_nav_default },
     "Floating Action Button" story NodeData { Story of R.layout.fragment_fab },
+    "Page Title" story NodeData { PageTitleStory of R.layout.fragment_page_title },
```
tiagoengel: We need a RTL version too

Backpack/src/main/res/values/dimens.xml
```diff
@@ -10,6 +10,12 @@
   <dimen name="bpk_horizontal_nav_size_normal">36dp</dimen>
   <dimen name="bpk_rating_score_badge_size_small">36dp</dimen>
   <dimen name="bpk_rating_score_badge_size_large">48dp</dimen>
+  <dimen name="bpk_page_title_toolbar_height">48dp</dimen>
```
tiagoengel: Since we are only using this values in code I think we should stick to using tokens. This one is spacing-lg * 2 for example.

app/src/main/java/net/skyscanner/backpack/demo/components/LockableNestedScrollView.kt
```diff
@@ -0,0 +1,25 @@
+package net.skyscanner.backpack.demo.components
+
+import android.annotation.SuppressLint
+import android.content.Context
+import android.util.AttributeSet
+import android.view.MotionEvent
+import androidx.core.widget.NestedScrollView
+
+class LockableNestedScrollView @JvmOverloads constructor(
```
tiagoengel: Why do we need this, would something like this be required in the app?

docs/PageTitle/README.md
```diff
@@ -0,0 +1,44 @@
...
+The Page Title component can be used in both XML and Kotlin/Java
+
+Example of a Page Title in XML
+
+```xml
+<net.skyscanner.backpack.pagetitle.BpkPageTitle
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content"
+  app:pageTitleText="Page Title"
+  app:pageTitleExpandedTextColor="@color/bpkYellow500"/>
```
tiagoengel: I'm a bit confused why `app:pageTitleExpandedTextColor` is not under themed props and `pageTitleCollapsedTextColor` is. I think we should list `pageTitleCollapsedTextColor` (with the default value) and remove from Theme Props as this is part of the "normal" props for this component.

Another note, we list all attributes here (minus theme attributes) with the **default** value so it's easier for consumers to see what attributes are available.

tiagoengel:
Looks good locally, can you make sure we have a card to fix look at RTL behaviour again?

PR #366, opened 2019-11-07
[BPK-3145] Added Dark Mode support for the Checkbox.

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
```diff
@@ -30,27 +30,33 @@ open class BpkCheckbox @JvmOverloads constructor(
   }
 
   private fun initialize(attrs: AttributeSet?, defStyleAttr: Int) {
-    var defaultColor = ContextCompat.getColor(context, R.color.bpkTextSecondary)
-    var checkedColor = BpkTheme.getPrimaryColor(context)
-    var disabledColor = ContextCompat.getColor(context, R.color.bpkSkyGrayTint04)
+    var checkmarkDefaultColor = ContextCompat.getColor(context, R.color.bpkTextSecondary)
+    var checkmarkCheckedColor = BpkTheme.getPrimaryColor(context)
+    var checkmarkDisabledColor = ContextCompat.getColor(context, R.color.bpkSkyGrayTint04)
+    val textDisabledColor = ContextCompat.getColor(context, R.color.bpkTextTertiary)
```
tiagoengel: This seems a bit too dark for the text colour

Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
```diff
@@ -30,27 +30,33 @@ open class BpkCheckbox @JvmOverloads constructor(
   }
 
   private fun initialize(attrs: AttributeSet?, defStyleAttr: Int) {
-    var defaultColor = ContextCompat.getColor(context, R.color.bpkTextSecondary)
-    var checkedColor = BpkTheme.getPrimaryColor(context)
-    var disabledColor = ContextCompat.getColor(context, R.color.bpkSkyGrayTint04)
+    var checkmarkDefaultColor = ContextCompat.getColor(context, R.color.bpkTextSecondary)
+    var checkmarkCheckedColor = BpkTheme.getPrimaryColor(context)
+    var checkmarkDisabledColor = ContextCompat.getColor(context, R.color.bpkSkyGrayTint04)
```
tiagoengel: Shouldn't this be the same as the text colour?

UNRELEASED.md
```diff
@@ -2,6 +2,9 @@
 
 > Place your changes below this line.
 
+- **Added:**
+  - Dark Mode support for `Checkbox`
```
tiagoengel: We should say the text colour is now SkyGray and cannot be changed. I would consider this a breaking change

Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
bvitaliyg: See the screenshots, this seems to be the only reasonable replacement among the semantic colours.

Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
bvitaliyg: No, if you look on the previous screenshots you notice it was always darker.

UNRELEASED.md
bvitaliyg: Ok

Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
tiagoengel: Yes but just a little bit, and because we were not setting the text colour before. Now is a very visible difference. How does it look it in dark mode? Why can't we just use `bpkSkyGrayTint04` for the disabled text colour too?

Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
tiagoengel: I'm talking based on the screenshot, it doesn't look like a disabled text

tiagoengel: I think this should be `textQuaternary`. Is the same as tertiary now but that is wrong, see slack conversation for more.

tiagoengel: let's update this to `bpkSkyGrayTint04` for both dark and light mode, as this is what iOS is doing for link buttons for now

bvitaliyg: Ok, the text colour is bpkSkyGrayTint04 now.

Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
bvitaliyg: Done

PR #388, opened 2019-12-18
[BPK-3403] Implemented the flavours, updated the publishing script.

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/gradle-maven-push.gradle
+++ b/Backpack/gradle-maven-push.gradle
@@ -29,7 +29,7 @@ afterEvaluate { project ->
             maven(MavenPublication) {
                 groupId 'com.github.skyscanner'
                 artifactId 'backpack-android'
-                version version
+                version "$version-internal"
                 description description
 
               artifact("$buildDir/outputs/aar/Backpack-internal-release.aar")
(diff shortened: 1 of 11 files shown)


Backpack/src/internal/res/values/backpack.font.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <style name="BackpackFont" parent="BackpackFontBase">
+    <item name="bpkFontFamilyBase">monospace</item>
```
tiagoengel: Can we keep the normal sans serif font here for now? Monospace was just a test but we don't want people using this by mistake.

Backpack/gradle-maven-push.gradle
```diff
@@ -32,7 +32,7 @@ afterEvaluate { project ->
                 version version
```
tiagoengel: What do you think we add `-internal` to the version name as well to make it impossible to use it by mistake. e.g version would be `20.0.1-internal`

PR #401, opened 2020-02-11
[NO-JIRA] Automate screenshots

tiagoengel:
This is a quick overview of how the script works:

- Builds and install `app-internal` and `app-test-internal`
- Starts a python `http` server listening for screenshot requests
- Runs tests in the emulator specifying that only `GenerateScreenshots` test should run.
- `GenerateScreenshots` runs two tests for every `story`, one for light and one for dark mode.
- Tests will prepare the view for screenshots and call the python server, that will then run adb to take the screenshot and save to the docs folder.

# Todo:

- [x] Component that requires action (e.g. toast)
- [x] Component with animation (e.g. spinner)
- [x] Save files in the correct folder (and without the hash in the name)
- [x] Dark mode
- [x] Docs
- [x] Automate build and execution

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
+++ b/app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
@@ -1,0 +1,109 @@
+package net.skyscanner.backpack.docs
+
+import androidx.test.espresso.Espresso
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.matcher.ViewMatchers
+import net.skyscanner.backpack.calendar.BpkCalendar
+import net.skyscanner.backpack.calendar.model.CalendarRange
+import net.skyscanner.backpack.demo.R
+import org.threeten.bp.LocalDate
+
+object DocsRegistry {
+  val screenshots = listOf(
+    Screenshot("Badge", "all"),
+    Screenshot("Button - Primary", "primary"),
+    Screenshot("Button - Secondary", "secondary"),
+    Screenshot("Button - Featured", "featured"),
+    Screenshot("Button - Destructive", "destructive"),
+    Screenshot("Button - Outline", "outline"),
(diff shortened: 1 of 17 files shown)


app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -63,7 +67,7 @@ private infix fun String.story(story: NodeData): Pair<String, NodeItem> {
 object ComponentRegistry {
 
   private val COMPONENTS_TREE = mapOf(
-    "Badge" story NodeData { Story of R.layout.fragment_badge },
+    "Badge" story NodeData({ Story of R.layout.fragment_badge }, "all"),
```
tiagoengel: This might not be the best approach, I'm thinking about leaving all info about this in the test case

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -0,0 +1,109 @@
...
+    Screenshot("Spinner - Small", "small"),
+    // Leave toast last as it stays visible in the screen for a while
+    Screenshot("Toast", "default", ::setupToast)
+    ).map { it.toArgs() }
+}
+
+data class Screenshot(
+  val name: String,
+  val screenshotName: String,
+  val setup: (() -> Unit)? = null
```
bvitaliyg: You can replace nullable function with empty function like this:
() -> Unit = {}. Less nullability checks, more safe code :)

At your discretion.

tiagoengel: I think a nullable value conveys better that this is not a required step and only present for a few use cases. Plus nullables are safe in Kotlin, the compiler will force you to treat them accordingly.

PR #407, opened 2020-02-26
[BPK-3693] limit colour usage in coloured calendar

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar/model/CalendarColoring.kt
```diff
@@ -1,22 +1,85 @@
...
+    val negative = object : CalendarCellStyle() {
+      override fun color(context: Context) = ContextCompat.getColor(context, R.color.bpkHillier)
+      override fun textStyle(context: Context) = TextStyle.Light
+    }
+
+    /**
+     * A custom cell style.
+     */
+    @JvmStatic
+    fun custom(@ColorInt color: Int?, textStyle: TextStyle? = null) = object : CalendarCellStyle() {
```
bvitaliyg: Please change color to not-null int and use with default argument, e.g.
@ColorInt color: Int = Color.Transparent

Also, consider adding default text style value to make the default value more explicit.

Backpack/src/main/java/net/skyscanner/backpack/calendar/model/CalendarColoring.kt
```diff
@@ -1,22 +1,85 @@
...
+      override fun color(context: Context) = ContextCompat.getColor(context, R.color.bpkErfoud)
+    }
+
+    /**
+     * A negative cell style which is suitable to indicate for example
+     * a date which has a comparatively high price among the dates in
+     * the calendar.
+     */
+    @JvmStatic
+    val negative = object : CalendarCellStyle() {
```
bvitaliyg: I'd use the camel case for these values, to make it feel more similar to enum and highlight the fact it's static.

PR #409, opened 2020-03-04
Bpk 3722 dev android create backpack component for bottom nav

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/bottomnav/BpkBottomNav.kt
```diff
@@ -0,0 +1,122 @@
...
+        if (menu.getItem(i) == item) {
+          return i
+        }
+      }
+      return -1
+    }
+  }
+
+  // We have to create a custom span as a BpkFontSpan causes some animation glitches as it sets text size as well.
+  private class BottomNavSpan(private val font: BpkText.FontDefinition) : CharacterStyle() {
```
tiagoengel: Does that mean we are not using a valid Backpack text size?

bvitaliyg: You cannot customise the text size of the Material Bottom Nav -- but as far as I can see it uses one of our sizes.

PR #427, opened 2020-03-24
[RHN-1408] Add ScrollListener and selection option to BpkCalendar

gabornovak:
## What
Add `onScrollListener` and `setSelectionFromTop` to `BpkCalendar`.

## Why
We have a project where we have to keep two `BpkCalendar`'s scroll position in sync (two different `Fragment` instance in a `ViewPager2`).
For that we needed to have access to the scrollable content within `BpkCalendar`. Since the calendar itself is a `ConstraintLayout` we cannot add scroll listener to it.
As a workaround we checked the `BpkCalendar`'s source code, and we found the `ListView`. We access it from the skyscanner app with the following code:
```
private fun BpkCalendar.getListView() = this.findViewById<ListView>(R.id.calendar_view)
```
Since we overriding the `onScrollListener` we've lost the year pill functionality as well as a side effect. It's clearly not ideal for numerous reasons.

## How
In this PR I'm adding two methods to the `BpkCalendar` to help us. The public methods are the following:
- `fun addOnScrollListener(listener: BpkCalendarScrollListener)`, and `fun removeOnScrollListener(listener: BpkCalendarScrollListener)`
This will let us add and remove a `BpkCalendarScrollListener` scroll listener on the `BpkCalendar`, which internally set the listener on the `CalendarView`. The `BpkCalendar` clearly scrollable, this method will let others register for the scroll events.  Also this will fix the year pill issue, since now we keep the original scroll listener.
- `fun setSelectionFromTop(position: Int, y: Int = 0)`
This method let's us scroll to a specific position with a specific offset. It's an `AbsListView` method. This method is needed to keep the scroll state in sync (we call this method, when the other calendar scrolled).

[This is the existing solution in the Skyscanner app](https://github.skyscannertools.net/apps-tribe/skyscanner-app/blob/develop/android/src/features/tripplanning/src/main/java/net/skyscanner/tripplanning/presentation/fragment/ListViewScrollSynchronizer.kt)

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
```diff
@@ -94,8 +95,10 @@ internal class CalendarView constructor(
     previousScrollState = currentScrollState
 
     listener?.onYearChanged(child.getYear())
+    scrollListener?.onScroll(this, firstVisibleItem, visibleItemCount, totalItemCount)
```
tiagoengel: I think this would work better if we remove the current `OnYearChangedListener` and allow more than one `scrollListener` to be added. This new scroll listener should provide the current year as the last argument as well.

```kotlin
interface BpkCalendarScrollListener {
   fun onScroll(view: AbsListView, firstVisibleItem: Int, visibleItemCount: Int, totalItemCount: Int, year: Int)
}
```

```kotlin
fun addScrollListener(listener: BpkCalendarSscrollListener) {
  listeners.add(listener)
}
fun removeScrollListener(listener: BpkCalendarSscrollListener) {
   listeners.remove(listener)
}
```

And in that case remove the custom  `OnYearChangedListener`

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
```diff
@@ -417,6 +417,22 @@ class BpkCalendarTest : BpkSnapshotTest() {
     snap(wrapWithBackground(calendar))
   }
 
+  @Test
+  @FlakyTest
```
tiagoengel: Any reason this is flaky?

Backpack/src/main/java/net/skyscanner/backpack/calendar/BpkCalendar.kt
```diff
@@ -39,6 +40,14 @@ open class BpkCalendar @JvmOverloads constructor(
     updateYearPill(controller.startDate.year)
   }
 
+  fun setOnScrollListener(listener: AbsListView.OnScrollListener) {
```
tiagoengel: We we follow my suggestion bellow we can change this to `addScrollListener` and also add a new `removeScrollListener`

gabornovak:
@tiagoengel , thanks for the review. Can you help me with the screenshot test recording? I cannot run it locally, the emulator doesn't set up correctly.

UNRELEASED.md
```diff
@@ -24,6 +24,9 @@ See [`CHANGELOG.md`](CHANGELOG.md) for real-world examples of good changelog ent
 
 - `bpk-component-infinity-gauntlet`:
   - New `timeStone` prop for controlling time. See &lt;link to docs site&gt;.
+- `BpkCalendar`:
```
tiagoengel: This should be a new entry at the top of the file, those are just examples

CONTRIBUTING.md
```diff
@@ -57,7 +57,7 @@ $ANDROID_HOME/tools/emulator -avd bpk-droid-avd -sdcard sd.img &
 After adding new snapshot tests, run
 
 ```
-./gradlew app:recordDebugAndroidTestScreenshotTest
+./gradlew app:recordOssDebugAndroidTestScreenshotTest
```
tiagoengel: 👍

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
```diff
@@ -49,8 +49,7 @@ internal class CalendarView constructor(
       }
     }
 
-  var listener: OnYearChangedListener? = null
-  var scrollListener: OnScrollListener? = null
+  var calendarScrollListeners = mutableSetOf<BpkCalendarScrollListener>()
```
tiagoengel: Can you please make this private and add `add` and `remove` listener functions?

gabornovak: I updated the PR

tiagoengel: This seems to be still the same, can we make this private and add functions to add and remove listeners?

UNRELEASED.md
tiagoengel: This is still in the wrong location, should be a the top of the file.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/CalendarView.kt
gabornovak: Sorry, forgot to push!

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
```diff
@@ -417,6 +417,21 @@ class BpkCalendarTest : BpkSnapshotTest() {
     snap(wrapWithBackground(calendar))
   }
 
+  @Test
+  fun screenshotTestCalendarSetSelectionFromTop_withTheme() {
```
tiagoengel: One last thing, can you remove `withTheme` from the name here and also not use `createThemedContext`. There is not reason to use the themed version for this test.

gabornovak: Make sense, done!

PR #428, opened 2020-03-24
[BPK-3774] Enable calendar footer view

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/MonthAdapter.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/MonthAdapter.kt
@@ -59,14 +59,17 @@ internal class MonthAdapter(
 
   override fun hasStableIds() = true
 
+  override fun getViewTypeCount() = 2
+
+  override fun getItemViewType(position: Int) = positionMetadata[position].viewType
+
   @Suppress("UNCHECKED_CAST")
   override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
     val metadata = positionMetadata[position]
     val month = metadata.month
     val year = metadata.year
-    val viewType = metadata.viewType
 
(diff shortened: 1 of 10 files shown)


Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/MonthAdapter.kt
```diff
@@ -17,81 +17,134 @@
...
-        }
+  override fun hasStableIds() = true
 
-        view.reuse()
+  @Suppress("UNCHECKED_CAST")
+  override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
+    val metadata = positionMetadata[position]
+    val month = metadata.month
+    val year = metadata.year
+    val viewType = metadata.viewType
```
bvitaliyg: There's getItemViewType() method and getViewTypeCount(), designed for this specifically and used for internal optimisations as well. Let's use it instead.

Backpack/src/main/java/net/skyscanner/backpack/calendar/presenter/MonthAdapter.kt
```diff
@@ -17,81 +17,134 @@
...
+  private fun isSelectedDayInMonth(calendarDay: LocalDate, year: Int, month: Int) =
+    calendarDay.year == year && calendarDay.month.value == month
+
+  private fun onDayTapped(day: LocalDate) {
+    controller.onDayOfMonthSelected(day)
+    selectedDay = day
+  }
+
+  private fun getFooterView(convertView: View?, month: Int, year: Int): View {
+    val shouldBind = convertView != null && convertView !is MonthView
```
bvitaliyg: As far as I remember you don't need to do a type check when you use the viewTypes. Let's just check it and make sure it works okay without it.

PR #435, opened 2020-04-07
[BPK-3808] Add support for flare pointing upwards

tiagoengel:
Screenshot_1586255761

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
```diff
@@ -176,13 +211,13 @@ open class BpkFlare @JvmOverloads constructor(
     val height = height.toFloat()
     val radiusHeight = radiusMask.height
     val radiusWidth = radiusMask.width
-    val radiusHalfHeight = (radiusHeight / 2).toFloat()
-    val radiusHalfWidth = (radiusHeight / 2).toFloat()
+    val radiusHalfHeight = radiusHeight.toFloat() / 2
```
tiagoengel, line 214: This fixes a little issue that was causing it to render a line on the the right borders. It's visible on the docs site. https://backpack.github.io/components/flare?platform=android#rounded

Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
```diff
@@ -151,20 +175,31 @@ open class BpkFlare @JvmOverloads constructor(
...
 
-    clipRect.set(width, yStart, pointerXEnd, height)
+    clipRect.set(width, yStart, pointerXEnd, yEnd)
     canvas.clipOutRectCompat(clipRect)
 
     super.draw(canvas)
     canvas.restoreToCount(count)
 
-    canvas.drawBitmap(pointerMask, pointerXStart, yStart, paint)
+    val mask = if (pointerDirection == PointerDirection.DOWN) {
```
bvitaliyg, line 196: instead of allocating different masks you can just rotate here the existing one. E.g. something like
val count = canvas.save()
canvas.rotate(90)
canvas.drawBitmap()
canvas.restoreToCount(count)

Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
```diff
@@ -151,20 +175,31 @@ open class BpkFlare @JvmOverloads constructor(
       pointerPosition.offset
     }
 
-    val yStart = height - pointerMask.height
+    val (yEnd, yStart) = if (pointerDirection == PointerDirection.DOWN) {
```
bvitaliyg, line 178: Instead of using pairs you can just clip rect in this conditions. If at any point we'll be needed to draw it on left/right it'll be more scalable.

Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
```diff
@@ -176,13 +211,13 @@ open class BpkFlare @JvmOverloads constructor(
...
-    val radiusHalfHeight = (radiusHeight / 2).toFloat()
-    val radiusHalfWidth = (radiusHeight / 2).toFloat()
+    val radiusHalfHeight = radiusHeight.toFloat() / 2
+    val radiusHalfWidth = radiusHeight.toFloat() / 2
     val pointerHeight = pointerMask.height
 
     val count = canvas.saveCount
 
-    // bottom right corner
+    // bottom let corner
```
bvitaliyg, line 220: left?

Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
tiagoengel: Not sure I see the benefit, if we need different left/right logic we'll see what to do then, although that is very unlikely. This will duplicate the clip logic in both branches when the only thing that really changes is the y start/end.

tiagoengel: Doesn't apply anymore with the logic to rotate the canvas

Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
tiagoengel: Done

Backpack/src/main/java/net/skyscanner/backpack/flare/BpkFlare.kt
```diff
@@ -128,12 +146,26 @@ open class BpkFlare @JvmOverloads constructor(
...
 
-    if (insetPaddingMode == InsetPaddingMode.BOTTOM) {
-      val paddingBottom = child.paddingBottom + pointerMask.height
-      if (child.paddingStart > 0 || child.paddingEnd > 0) {
-        child.setPaddingRelative(child.paddingStart, child.paddingTop, child.paddingEnd, paddingBottom)
-      } else {
-        child.setPadding(child.paddingLeft, child.paddingTop, child.paddingRight, paddingBottom)
+    when (insetPaddingMode) {
+      InsetPaddingMode.NONE -> {}
+      InsetPaddingMode.BOTTOM, InsetPaddingMode.TOP -> {
```
bvitaliyg, line 151: Can we do this in the 2 different branches and use some sort of function/extension to handle this?

tiagoengel: Done

PR #438, opened 2020-04-14
[BPK-3835] Update calendar colours

tiagoengel:
Screenshot_1586872354
Screenshot_1586872347

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar/model/CalendarColoring.kt
```diff
@@ -54,8 +59,18 @@ sealed class CalendarCellStyle {
    * the calendar.
    */
   object Negative : CalendarCellStyle() {
-    override fun color(context: Context) = ContextCompat.getColor(context, R.color.bpkHillier)
-    override fun textStyle(context: Context) = TextStyle.Light
+    override fun color(context: Context) =
+      ContextCompat.getColor(context, R.color.__calendarCellStyleNegativeColor)
+
+    override fun textStyle(context: Context): TextStyle {
+      val nightModeFlags: Int = context.resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK
```
tiagoengel, line 66: Added this because the logic to automatic decide this doesn't work 100% here, the black font works fine when is bold, which is true in this case

PR #440, opened 2020-04-15
Bpk 3813 dev android create bar chart component

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/barchart/BpkBarChart.kt
```diff
@@ -0,0 +1,126 @@
...
+import net.skyscanner.backpack.util.Consumer
+import net.skyscanner.backpack.util.createContextThemeWrapper
+import net.skyscanner.backpack.util.use
+
+class BpkBarChart @JvmOverloads constructor(
+  context: Context,
+  attrs: AttributeSet? = null,
+  defStyleAttr: Int = 0
+) : FrameLayout(createContextThemeWrapper(context, attrs, R.attr.bpkBarChartStyle), attrs, defStyleAttr),
+  (BpkBarChart.Model) -> Unit {
```
tiagoengel, line 22: Shall we replace this with `Consumer`?

Backpack/src/main/java/net/skyscanner/backpack/util/ColorStateList.kt
```diff
@@ -0,0 +1,6 @@
+package net.skyscanner.backpack.util
```
tiagoengel, line 1: same as above

Backpack/src/main/java/net/skyscanner/backpack/util/Consumer.kt
```diff
@@ -0,0 +1,3 @@
+package net.skyscanner.backpack.util
```
tiagoengel, line 1: Can we not use the java `Consumer` interface?

Backpack/src/main/java/net/skyscanner/backpack/util/ItemHolder.kt
```diff
@@ -0,0 +1,39 @@
...
+  val context: Context
+    get() = itemView.context
+
+  val resources: Resources
+    get() = itemView.resources
+
+  var model: T? = null
+    private set
+
+  constructor(parent: ViewGroup, @LayoutRes layout: Int) :
```
tiagoengel, line 25: nit: constructor first

Backpack/src/main/java/net/skyscanner/backpack/util/ItemHolder.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.content.Context
+import android.content.res.Resources
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.IdRes
+import androidx.annotation.LayoutRes
+import androidx.recyclerview.widget.RecyclerView
+
+internal abstract class ItemHolder<T>(
```
tiagoengel, line 12: This seems a bit too specific to the BarChart component to be here, at lest for now. Can we move this inside the bar chart package?

Backpack/src/main/java/net/skyscanner/backpack/util/ViewGroup.kt
```diff
@@ -0,0 +1,16 @@
+package net.skyscanner.backpack.util
```
tiagoengel, line 1: same as above regarding the extension package

Backpack/src/main/res/values-night/internal.semantic.color.xml
```diff
@@ -21,4 +21,10 @@
   <color name="__textFieldHint">@color/bpkTextTertiaryDark</color>
   <color name="__textFieldIcon">@color/bpkTextSecondary</color>
   <color name="__bottomNavUnselected">@color/bpkTextSecondaryDark</color>
+
+  <color name="__barChartBarBackgroundColor">@color/bpkSkyGrayTint01</color>
```
tiagoengel, line 25: Is using sky gray the correct for dark mode?

Backpack/src/main/res/values/dimens.xml
```diff
@@ -30,4 +30,6 @@
   <dimen name="bpk_dialog_flare_max_width">440dp</dimen>
   <dimen name="bpk_snackbar_min_height">60dp</dimen>
   <dimen name="bpk_bottom_nav_height">56dp</dimen>
+  <dimen name="bpk_barchart_popup_triangle_width">12dp</dimen>
```
tiagoengel, line 33: Where does this come from? Isn't this pointer supposed to be the "flare" design?

app/src/main/java/net/skyscanner/backpack/demo/stories/BarChartStory.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.util.*
+import kotlin.collections.ArrayList
+
+class BarChartStory : Story() {
+
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
+
+    view.findViewById<BpkBarChart>(R.id.bar_chart).apply {
+      invoke(BpkBarChart.Model(
```
tiagoengel, line 16: I think at least for the external api we should avoid the `consumer` approach and call this `setModel`

app/src/main/java/net/skyscanner/backpack/demo/stories/BarChartStory.kt
```diff
@@ -0,0 +1,61 @@
...
+        )
+      )
+      )
+    }
+  }
+
+  private val random = Random()
+
+  private fun createMonth(month: Int) = BpkBarChart.Group(
+    title = arrayOf("January", "February", "March", "April", "May", "June", "Jule")[month % 6],
```
tiagoengel, line 37: nit: "July"

app/src/main/java/net/skyscanner/backpack/demo/stories/BarChartStory.kt
```diff
@@ -0,0 +1,61 @@
...
+      invoke(BpkBarChart.Model(
+        groups = listOf(
+          createMonth(0),
+          createMonth(1),
+          createMonth(2),
+          createMonth(3),
+          createMonth(4),
+          createMonth(5)
+        ),
+        legend = BpkBarChart.Legend(
```
tiagoengel, line 25: How does the legend relate to the data in the graph?

Backpack/src/main/java/net/skyscanner/backpack/barchart/internal/ChartGraphView.kt
```diff
@@ -0,0 +1,76 @@
...
+
+@SuppressLint("ViewConstructor")
+internal class ChartGraphView constructor(
+  context: Context,
+  colors: BpkBarChart.Colors,
+  onClick: Consumer<BpkBarChart.Column>
+) : FrameLayout(context), Consumer<List<BpkBarChart.Group>> {
+
+  private val onClickWrapper = { holder: ChartBarHolder ->
+    onClick(holder.model!!)
```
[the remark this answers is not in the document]
bvitaliyg: Not really. If the item view is not bound there will be no data. By using this operator we declare the contract that only the bound items should be displayed and be clickable.

Backpack/src/main/java/net/skyscanner/backpack/util/Canvas.kt
```diff
@@ -0,0 +1,9 @@
+package net.skyscanner.backpack.util
```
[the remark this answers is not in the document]
bvitaliyg: Extensions is sort of utils. I'd like to avoid confusion here as we might want to include some extensions to out public API in the future.

Backpack/src/main/java/net/skyscanner/backpack/util/Consumer.kt
bvitaliyg: Nope, it's only available from API 24. Also, I don't want to use interface here and just declare any function with this literal to be Consumer -- this will allow to use functional literals or even use primitive non-autoboxed types.

Backpack/src/main/java/net/skyscanner/backpack/util/ItemHolder.kt
bvitaliyg: According to the Kotlin guidelines, the properties should be declared first:

https://kotlinlang.org/docs/reference/coding-conventions.html#class-layout

Backpack/src/main/java/net/skyscanner/backpack/util/ItemHolder.kt
bvitaliyg: I want this to be used in every Backpack components using RecyclerView. This can be moved for now and then re-moved to util package when it's needed. Do you want me to do that?

Backpack/src/main/res/values-night/internal.semantic.color.xml
bvitaliyg: Yes, it supposed to have low contrast ratio with the background.

Backpack/src/main/res/values/dimens.xml
bvitaliyg: From design in Firma. Flare cannot be really used here as the popup is only a drawable and now a view.

app/src/main/java/net/skyscanner/backpack/demo/stories/BarChartStory.kt
bvitaliyg: It shares the same color scheme. The badges are also placed in the way to not affect to tooltips.

Backpack/src/main/java/net/skyscanner/backpack/barchart/internal/ChartGraphView.kt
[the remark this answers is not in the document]
tiagoengel: The problem is not the contract, but the exception that will be thrown in case an error happens here. Nullpointer exceptions are very uninformative so if this is the required contract can we throw a better exception?
E.g. `holder.model ?: throw IlegalStateException("View model not bound")`

Backpack/src/main/java/net/skyscanner/backpack/util/Canvas.kt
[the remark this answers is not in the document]
tiagoengel: Not gonna block the PR because of this but `utils` is too generic. Anything can be considered util if we follow this approach, which is not good, I think it's better to restrict  "global" extensions like this to one package, e.g. `extensions`

Backpack/src/main/java/net/skyscanner/backpack/util/ItemHolder.kt
tiagoengel: I think it's better to wait for abstractions to be needed before we generalise them like this, we will see it better when we have an actual need for this in more components, so I'd keep in the chart component for now. Up to you

Backpack/src/main/java/net/skyscanner/backpack/util/ItemHolder.kt
tiagoengel: That's new to me. I find it a bit confusing that they say to keep related stuff together and also recommend not to keep constructors together by declaring the props first 🤷. But not a blocker was just a nit pick

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -11,6 +11,7 @@ import org.threeten.bp.LocalDate
 object DocsRegistry {
   val screenshots = listOf(
     Screenshot("Badge", "all"),
+    Screenshot("Bar Chart", "default"),
```
tiagoengel, line 14: We should probably have one without legend (default) and one with legend, no?

docs/BarChart/README.md
```diff
@@ -0,0 +1,79 @@
+# Bar Chart
+
+## Installation
+
+Backpack Android is available through [Jitpack](https://jitpack.io/#Skyscanner/backpack-android). Check the main [Readme](https://github.com/skyscanner/backpack-android#installation) for a complete installation guide.
+
+## Usage
+
+The Bar Chart component can be used in both XML and Kotlin/Java
```
tiagoengel, line 9: Since it can't really be used only in XML I think we should change this to clarify it a bit.

E.g. The Bar Chart component can be used in both XML and Kotlin/Java, however, feeding the component with data can only be done programatically.

docs/BarChart/README.md
```diff
@@ -0,0 +1,79 @@
...
+          "No Price"
+        )
+      )
+```
+
+Another example can be found in `BarChartStory` in the sample project.
+
+## Theme Props
+
+All the colours support `ColorStateList`.
```
tiagoengel, line 68:
```suggestion
All colours support `ColorStateList`.
```

docs/BarChart/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+BpkBarChart(context).apply {
+    model = //.. initialisation
+}
+```
+
+To render the data you need to set a view model to Bar Chart in code.
+Detailed documentation on each property of the model is available in Javadoc.
+
+Here's some basic example in Kotlin:
```
tiagoengel, line 33:
```suggestion
To render data you need to set a view model to `BarChart` in code.
Detailed documentation on each property of the model is available [here](https://backpack.github.io/android/versions/latest/net.skyscanner.backpack.barchart).

Here's some basic example in Kotlin:
```

docs/BarChart/README.md
```diff
@@ -0,0 +1,79 @@
...
+barChart.model = BpkBarChart.Model(
+        groups = listOf(
+          createGroup(),
+          createGroup()
+        ),
+        legend = BpkBarChart.Legend(
+          "£",
+          "No Price"
+        )
+      )
```
tiagoengel, line 61:
```suggestion
barChart.model = BpkBarChart.Model(
  groups = listOf(
    createGroup(),
    createGroup()
  ),
  legend = BpkBarChart.Legend(
    "£",
     "No Price"
   )
 )
```

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
bvitaliyg: I think we can keep it for now.

docs/BarChart/README.md
bvitaliyg: Done!

PR #448, opened 2020-05-05
[BPK-3436] Add danger

tiagoengel:
This will prevent adding more files without license from now on.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/meta.json
+++ b/meta.json
@@ -17,11 +17,11 @@
       "github": "tiagoengel"
     },
     {
-      "github": "willhg"
-    },
-    {
       "npm": "ojcurt",
       "github": "olliecurtis"
+    },
+    {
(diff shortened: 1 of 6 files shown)


meta.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "npm": "shaundon",
+      "github": "shaundon"
+    },
+    {
+      "npm": "tiagohngl",
+      "github": "tiagoengel"
+    },
+    {
+      "github": "willhg"
```
shaundon, line 20: 😭

tiagoengel: Just realised @bvitaliyg is not in the file 😬. Will add it.

PR #508, opened 2020-11-16
Update README.md

amolgupta:
Documentation fix

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/docs/Calendar/README.md
+++ b/docs/Calendar/README.md
@@ -61,7 +61,7 @@ class ExampleBpkCalendarController(
   private val context: Context,
   override val selectionType: SelectionType
  ) : BpkCalendarController(selectionType) {
-  override fun onRangeSelected(range: CalendarSelection) {
+  override fun onRangeSelected(selection: CalendarSelection) {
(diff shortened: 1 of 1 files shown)


docs/Calendar/README.md
```diff
@@ -61,7 +61,7 @@ class ExampleBpkCalendarController(
   private val context: Context,
   override val selectionType: SelectionType
  ) : BpkCalendarController(selectionType) {
-  override fun onRangeSelected(range: CalendarSelection) {
+  override fun onRangeSelected(selection: CalendarSelection) {
```
georgegillams, line 64: I know nothing about Android, but I can only find usages of `onRangeSelected(range:...)`. Are you sure it's meant to say `selection`?

amolgupta: It's the variable name for the function. The body uses `selection` but it was named `range` leading to compilation error on copying from docs.

marianeum: Maybe we should rename the selection references to range instead, so it matches with the API

PR #509, opened 2020-11-16
[NO-JIRA] Fix avd name

georgegillams:
I'm not sure if this is actually an error, but I think it is?


shaundon:
Could you give some more context? What makes you think it's wrong etc?

georgegillams:
> Could you give some more context? What makes you think it's wrong etc?

Sure - all the other mentions in the file are different

shaundon (approved):

PR #521, opened 2020-12-17
2-line labels support for Calendar

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar/view/BpkCalendarLabelsViewModel.kt
```diff
@@ -38,10 +43,22 @@ internal class BpkCalendarLabelsViewModel(
 
   private var labels: Map<LocalDate, BpkCalendarLabelViewModel> = emptyMap()
 
+  private val placeholder = BpkCalendarLabelViewModel(
```
bvitaliyg, line 46: We're falling back to this placeholder if the labels is not available. The main reason is that drawing text on Canvas and drawing text using TextLayout requires different coordinates. We don't want to calculate the coordinates here or pass fallback coordinates to method each time, so we just use fallback object.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/BpkCalendarLabelsViewModel.kt
```diff
@@ -50,38 +67,46 @@ internal class BpkCalendarLabelsViewModel(
...
+    label.draw(canvas, x, y)
   }
 
   class BpkCalendarLabelViewModel(
     private val label: CalendarLabel,
     private val paint: TextPaint,
   ) {
 
-    private lateinit var ellipsizedText: String
+    private lateinit var text: StaticLayout
```
bvitaliyg, line 87: StaticLayout is a class from Android library responsible for proper multi-line slicing. It also supports text truncating

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/BpkCalendarLabelsViewModel.kt
```diff
@@ -50,38 +67,46 @@ internal class BpkCalendarLabelsViewModel(
...
 
     var width: Float = 0f
       set(value) {
         if (field != value) {
           field = value
-          ellipsizedText = TextUtils.ellipsize(label.text, paint, width, TextUtils.TruncateAt.END).toString()
+          text = TextLayoutCompat.staticLayout(
```
bvitaliyg, line 93: We're using custom helper class here, more on that later.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/BpkCalendarLabelsViewModel.kt
```diff
@@ -50,38 +67,46 @@ internal class BpkCalendarLabelsViewModel(
...
+            includePad = true,
+            ellipsize = TextUtils.TruncateAt.END,
+            maxLines = 2,
+          )
         }
       }
 
     fun draw(canvas: Canvas, x: Float, y: Float) {
-      canvas.drawText(ellipsizedText, x, y, paint)
+      canvas.withSave {
```
bvitaliyg, line 106: TextLayout doesn't support positioning using coordinates, so we translate the canvas itself.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt
```diff
@@ -99,7 +99,11 @@ internal class MonthView @JvmOverloads constructor(
   private val monthHeaderSize: Int = ResourcesUtil.dpToPx(52, context)
 
   private var rowHeight: Int = context.resources.getDimensionPixelSize(R.dimen.bpkSpacingXl) * 2
-  private var labelsYOffset: Int = context.resources.getDimensionPixelSize(R.dimen.bpkSpacingXl)
+  private var rowHeightLabeled: Int = context.resources.getDimensionPixelSize(R.dimen.bpkSpacingXxl) +
```
bvitaliyg, line 102: That's the value which provides 4dp gap to the bottom line, which is required by design.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt
```diff
@@ -194,6 +198,7 @@ internal class MonthView @JvmOverloads constructor(
         field = value
         labelsViewModel.update(value.calendarLabels)
         isRtl = value.isRtl
+        requestLayout()
```
bvitaliyg, line 201: Since the row can change the height now, entire view height might be changed. By calling this method we make sure the size will be correctly re-calculated.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt
```diff
@@ -224,6 +229,7 @@ internal class MonthView @JvmOverloads constructor(
   }
 
   override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+    val rowHeight = if (labelsViewModel.isEmpty()) rowHeight else rowHeightLabeled
```
bvitaliyg, line 232: Picking the correct row height

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+import android.util.Log
+import androidx.annotation.RequiresApi
+import java.lang.reflect.Constructor
+import net.skyscanner.backpack.BuildConfig
+
+internal object TextLayoutCompat {
+
+  private const val TAG = "TextLayoutCompat"
+
+  private val creator = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
bvitaliyg, line 37: Prior to Android API 23 (Marshmallow) Android doesn't provide any public API to make static layout multi-line.

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+    textDir = textDir,
+    spacingMult = spacingMult,
+    spacingAdd = spacingAdd,
+    includePad = includePad,
+    ellipsize = ellipsize,
+    ellipsizedWidth = ellipsizedWidth,
+    maxLines = maxLines,
+  )
+
+  interface StaticLayoutCreator {
```
bvitaliyg, line 73: We extracted all the creation logics to an interface and simply use it each time we needed. The similar logic is used in Android backward compatibility libraries.

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+      includePad: Boolean,
+      ellipsize: TextUtils.TruncateAt?,
+      ellipsizedWidth: Int,
+      maxLines: Int
+    ): StaticLayout
+  }
+
+  object StaticLayoutCreatorApi21 : StaticLayoutCreator {
+
+    private val ConstructorReference: Constructor<StaticLayout>? = try {
```
bvitaliyg, line 94: There's a way to create multi-line StaticLayout in Android <= 23, but the API is not public. It's used, however, by all default Android components.

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+      StaticLayout::class.java.getDeclaredConstructor(
+        CharSequence::class.java,
+        Int::class.javaPrimitiveType,
+        Int::class.javaPrimitiveType,
+        TextPaint::class.java,
+        Int::class.javaPrimitiveType,
+        Layout.Alignment::class.java,
+        TextDirectionHeuristic::class.java,
+        Float::class.javaPrimitiveType,
+        Float::class.javaPrimitiveType,
```
bvitaliyg, line 104: Here we're using reflection to find the constructor we need.

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+      includePad: Boolean,
+      ellipsize: TextUtils.TruncateAt?,
+      ellipsizedWidth: Int,
+      maxLines: Int,
+    ): StaticLayout = try {
+
+      ConstructorReference?.newInstance(
+        source, bufStart, bufEnd, paint, outerWidth, align, textDir, spacingMult,
+        spacingAdd, includePad, ellipsize, ellipsizedWidth, maxLines,
+      ) ?: throw NullPointerException("StaticLayoutConstructor is null!")
```
bvitaliyg, line 141: Allocating StaticLayout using reflection.

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+    } catch (t: Throwable) {
+
+      if (BuildConfig.DEBUG) {
```
bvitaliyg, line 144: This call should work fine as it's used by most of the Android default widgets. However, you can never be 100% sure if you use reflection. Therefore we log error if it happens..

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+        Log.e(TAG, "Failed to create StaticLayout instance using reflection.", t)
+      }
+
+      StaticLayout(
```
bvitaliyg, line 148: And return default StaticLayout created by public API. It won't be limited by 2 lines, but it will work in the same way. It's better than crashing the app, and this should never happen in real life.

Backpack/src/main/java/net/skyscanner/backpack/util/TextLayoutCompat.kt
```diff
@@ -0,0 +1,182 @@
...
+      outerWidth: Int,
+      align: Layout.Alignment,
+      textDir: TextDirectionHeuristic,
+      spacingMult: Float,
+      spacingAdd: Float,
+      includePad: Boolean,
+      ellipsize: TextUtils.TruncateAt?,
+      ellipsizedWidth: Int,
+      maxLines: Int,
+    ): StaticLayout = StaticLayout.Builder.obtain(source, bufStart, bufEnd, paint, outerWidth)
```
bvitaliyg, line 172: On API 23 and higher we can just used the public API.

app/src/androidTest/java/net/skyscanner/backpack/calendar/BpkCalendarTest.kt
```diff
@@ -151,6 +151,8 @@ class BpkCalendarTest : BpkSnapshotTest() {
           CalendarLabel(text = "£12", style = CalendarLabel.Style.PriceLow),
         LocalDate.of(initialStartDate.year, initialStartDate.month, initialStartDate.dayOfMonth + 4) to
           CalendarLabel(text = "£900000000000000", style = CalendarLabel.Style.PriceLow),
+        LocalDate.of(initialStartDate.year, initialStartDate.month, initialStartDate.dayOfMonth + 5) to
+          CalendarLabel(text = "£900000", style = CalendarLabel.Style.PriceLow),
```
bvitaliyg, line 155: Text will take 2 lines it won't be truncated. We use this to make sure each line will be centred.

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/BpkCalendarLabelsViewModel.kt
```diff
@@ -38,10 +43,22 @@ internal class BpkCalendarLabelsViewModel(
 
   private var labels: Map<LocalDate, BpkCalendarLabelViewModel> = emptyMap()
 
+  private val placeholder = BpkCalendarLabelViewModel(
+    label = CalendarLabel(
+      text = "-",
```
georgegillams, line 48: This should be an n-dash instead of a hyphen as it's (sort of) copy. It looks the same in code as the font is monospaced, but should look better in the app

```suggestion
      text = "–",
```

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt
georgegillams: As we require a very specific spacing it might be more readable to just hard code it?

georgegillams (dismissed):
LGTM. Just a couple of minor comments 👍

georgegillams (approved):

PR #543, opened 2021-02-11
Implemented Map markers.

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/map/BpkMap.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.google.android.gms.maps.GoogleMap
+import com.google.android.gms.maps.MapView
+import com.google.android.gms.maps.SupportMapFragment
+import com.google.android.gms.maps.model.Marker
+import net.skyscanner.backpack.map.view.createBpkMarkerView
+
+/**
+ * Receives the map in async mode and adds some Backpack customisations to it.
+ */
+fun MapView.getBpkMapAsync(onReady: (GoogleMap) -> Unit) {
```
bvitaliyg, line 32: Google provides 2 ways to implement maps - MapView (extends UIView) and SupportMapFragments (extends UIViewController). In the first case you have more flexibility, but have to write more code as you need to dispatch all the lifecycle methods.

Regardless the way you choose, you need to retrieve GoogleMaps object (which is a configurator for everything) using this API. We're wrapping the call to inject some Backpack customisation code there.

Backpack/src/main/java/net/skyscanner/backpack/map/BpkMap.kt
```diff
@@ -0,0 +1,59 @@
...
+    it.setInfoWindowAdapter(BpkWindowInfoAdapter(requireContext()))
+    onReady(it)
+  }
+}
+
+private class BpkWindowInfoAdapter(
+  private val context: Context,
+) : GoogleMap.InfoWindowAdapter {
+
+  override fun getInfoWindow(marker: Marker): View =
```
bvitaliyg, line 53: This code is responsible for showing the selected state of the marker.

Backpack/src/main/java/net/skyscanner/backpack/map/BpkMap.kt
```diff
@@ -0,0 +1,59 @@
...
+    createBpkMarkerView(context, marker.title, marker.icon, showPointer = !marker.pointerOnly)
+      .apply { isSelected = !marker.pointerOnly }
```
bvitaliyg, line 55: When isSelected set to true, the badge uses white background. We don't want to do this for pointers-only markers for now.

Backpack/src/main/java/net/skyscanner/backpack/map/BpkMapMarker.kt
```diff
@@ -0,0 +1,155 @@
...
+import com.google.android.gms.maps.model.Marker
+import com.google.android.gms.maps.model.MarkerOptions
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.map.view.createBpkMarkerView
+import net.skyscanner.backpack.util.rasterize
+
+/**
+ * An icon which will be displayed on the badge.
+ */
+var Marker.icon: Int
```
bvitaliyg, line 39: Google Maps doesn't have a way to put an icon in marker. You can, however, associate it with some tag (an object).

We use this object to store any custom-related properties. This is extension property - synthetic  way to expand the class functionality if you don't have access to its source code editing. Basically it's a syntax sugar for calling utility functions.
 
https://kotlinlang.org/docs/extensions.html

Backpack/src/main/java/net/skyscanner/backpack/map/BpkMapMarker.kt
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * An alternative solution to associate some object with given [Marker].
+ *
+ * [Marker.getTag] or [Marker.setTag] should never be used – Backpack utilises it
+ * to store its internal properties (e.g. [Marker.icon]).
+ *
+ * @see Marker.getTag
+ * @see Marker.setTag
+ */
+var Marker.bpkTag: Any?
```
bvitaliyg, line 54: Since we use tags for our own custom object, we provide another property to be used instead of tag.

Backpack/src/main/java/net/skyscanner/backpack/map/BpkMapMarker.kt
```diff
@@ -0,0 +1,155 @@
...
+import net.skyscanner.backpack.map.view.createBpkMarkerView
+import net.skyscanner.backpack.util.rasterize
+
+/**
+ * An icon which will be displayed on the badge.
+ */
+var Marker.icon: Int
+  @DrawableRes get() = extra.icon
+  set(@DrawableRes value) {
+    extra = extra.copy(icon = value)
```
bvitaliyg, line 42: extra is an immutable class. Therefore we're not changing the property but cloning the entire class with different property value.

Backpack/src/main/java/net/skyscanner/backpack/map/view/BpkMapMarkersTestBridge.java
```diff
@@ -0,0 +1,35 @@
...
+// This class is used for screenshots testing only.
+// It's not a part of public API and may be changed at any moment without being considered as a breaking change.
+import org.jetbrains.annotations.NotNull;
+
+/*package*/ enum BpkMapMarkersTestBridge {
+  Instance;
+
+  @NotNull
+  /*package*/ View create(@NotNull Context context, @NotNull String title, int icon, boolean showPointer) {
+    return BpkMapViewsKt.createBpkMarkerView(context, title, icon, showPointer);
```
bvitaliyg, line 33: As we declare createBpkMarkerView() method internal, it can only be used inside the same module. Our tests are located in a different module and therefore cannot use this method.

To bypass it, we use Java singleton with package visibility modifiers. This can be used from another module providing that the packages are the same.

There's a comment to prevent anyone to use the same trick for something else.

Backpack/src/main/java/net/skyscanner/backpack/map/internal/BpkMapMarkersTestBridge.java
```diff
@@ -0,0 +1,35 @@
...
+// This class is used for screenshots testing only.
+// It's not a part of public API and may be changed at any moment without being considered as a breaking change.
+import org.jetbrains.annotations.NotNull;
+
+/*package*/ enum BpkMapMarkersTestBridge {
+  Instance;
+
+  @NotNull
+  /*package*/ View create(@NotNull Context context, @NotNull String title, int icon, boolean showPointer) {
+    return BpkMapViewsKt.createBpkMarkerView(context, title, icon, showPointer);
```
bvitaliyg, line 33: As we declare createBpkMarkerView with internal visibility modifier, it cannot be used outside its module. However our tests are located in a different module(app).

To bypass it, we use Java singleton (as enum) with package visibility modifier. In this case any module can use this code providing the code has the same package.

To prevent any other usages we have the very special comment :)

Backpack/src/main/java/net/skyscanner/backpack/util/Drawables.kt
```diff
@@ -94,3 +96,14 @@ internal inline fun sizedDrawable(
   override fun getIntrinsicHeight(): Int =
     height
 }
+
+internal fun Drawable.rasterize(
```
bvitaliyg, line 100: That's another extensions function. You may see its call in BpkMapMarket.kt file.

Backpack/src/main/java/net/skyscanner/backpack/util/Views.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package net.skyscanner.backpack.util
+
+import android.graphics.Bitmap
+import android.graphics.Canvas
+import android.graphics.Color
+import android.view.View
+
+internal inline fun View.rasterize(
```
bvitaliyg, line 26: System uses 3 steps to draw a view. Here's we're mimicking it.

Backpack/src/main/java/net/skyscanner/backpack/util/Views.kt
```diff
@@ -0,0 +1,44 @@
...
+  widthMeasureSpec: Int = View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
+  heightMeasureSpec: Int = View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
+  bitmapConfig: Bitmap.Config = Bitmap.Config.ARGB_8888,
+  onLayoutDone: (View) -> Unit = {},
+): Bitmap {
+
+  measure(widthMeasureSpec, heightMeasureSpec)
```
bvitaliyg, line 33: The view is being measure with "wrap_content" inartistics.

Backpack/src/main/java/net/skyscanner/backpack/util/Views.kt
```diff
@@ -0,0 +1,44 @@
...
+  layout(0, 0, measuredWidth, measuredHeight)
```
bvitaliyg, line 34: Then it's being layout within given rect.

Backpack/src/main/java/net/skyscanner/backpack/util/Views.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  onLayoutDone(this)
+
+  val bitmap = Bitmap.createBitmap(width, height, bitmapConfig)
+  bitmap.eraseColor(Color.TRANSPARENT)
+  val canvas = Canvas(bitmap)
+  draw(canvas)
```
bvitaliyg, line 41: And then drawn on Canvas.

Backpack/src/main/res/drawable/bpk_map_marker_bg.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bvitaliyg, line 1: This draws the badge background in selected/unselected state.

Backpack/src/main/res/color/bpk_map_marker_content_color.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
```
bvitaliyg, line 2: These are colors for text and icons in selected/unselected state.

Backpack/src/main/res/drawable/bpk_map_marker_pointer.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bvitaliyg, line 1: Is draws circle below the badge

Backpack/src/main/res/layout/view_bpk_map_marker_label.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
bvitaliyg, line 2: Layout used to draw a marker without icon

Backpack/src/main/res/layout/view_bpk_map_marker_label_with_icon.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bvitaliyg, line 1: Layout used to draw a marker with icon

app/build.gradle
```diff
@@ -6,7 +6,10 @@ apply plugin: 'kotlin-android-extensions'
 
 apply plugin: 'com.facebook.testing.screenshot'
 
-def travis = System.getenv("TRAVIS")
+def properties = new Properties()
```
bvitaliyg, line 9: Parsing properties (e.g. singing configs) from local.properties

olliecurtis (approved):
LGTM! Tested locally and works as expected

PR #553, opened 2021-03-11
Removed isRtl property from BpkCalendarController.kt

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt
```diff
@@ -197,11 +197,17 @@ internal class MonthView @JvmOverloads constructor(
       if (value != null) {
         field = value
         labelsViewModel.update(value.calendarLabels)
-        isRtl = value.isRtl
+        isRtl = layoutDirection == LAYOUT_DIRECTION_RTL
```
georgegillams, line 200: Does this need to be set on line 109 and here?

bvitaliyg: This is just to keep the existing behaviour as much intact as possible. We can remove it in the future. I prefer to do big code changes gradually, step by step.

georgegillams (approved):

Backpack/src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt
georgegillams: Makes sense 👍

PR #555, opened 2021-03-17
Updated Rating design

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/rating/internal/RatingSelectors.kt
```diff
@@ -90,13 +91,22 @@ internal class RatingSelectors(
...
     )
-    color = {
-      colors[it.index]
+    backgroundColor = {
+      backgroundColors[it.index]
+    }
+
+    val contentColors = arrayOf(
+      ContextCompat.getColorStateList(context, R.color.bpkWhite)!!,
+      ContextCompat.getColorStateList(context, R.color.bpkSkyGray)!!,
```
georgegillams, line 105: I think this should technically be black in dark mode

bvitaliyg: This is the colours for the rating score, which is always displayed agains certain colours (equal for LM/DM). Therefore the text colour also remains the same regardless the mode.

georgegillams: The reason I think technically it should be black is that we don't use sky-grey in dark-mode.
https://backpack.github.io/guidelines/colour#palette

georgegillams (approved):

PR #565, opened 2021-04-14
Fixed demo app build.

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


.github/workflows/ci.yml
```diff
@@ -12,8 +12,8 @@ defaults:
     shell: bash -l {0}
 
 env:
-  flavour: Oss
-  config: Debug
+  flavour: oss
+  config: debug
```
bvitaliyg, line 16: It's used for Gradle tasks names, which are not case sensitive. However, we now use it for files as well, and the file names are in the lowercase.

README.md
```diff
@@ -27,9 +27,9 @@ repositories {
 #### Note that Backpack is expected to be used with AndroidX. Please refer to [AndroidX migration guide](https://developer.android.com/jetpack/androidx/migrate) to setup.
 
 ### Demo application
-The Backpack demo application is a good way of referring to the variants available for a component and their correct usage. The code is available under `/app` directory. [The app can be downloaded from HockeyApp](https://rink.hockeyapp.net/apps/dbf61ff1e7e245ddad0bd92ac24a85e1/app_versions/1) or by scanning the QR code below.
+The Backpack demo application is a good way of referring to the variants available for a component and their correct usage. The code is available under `/app` directory. [The app can be downloaded from App Center](https://install.appcenter.ms/orgs/Backpack/apps/Backpack-Android) or by scanning the QR code below.
 
-![QR code](https://chart.googleapis.com/chart?cht=qr&chl=https%3A%2F%2Frink.hockeyapp.net%2Fapps%2Fdbf61ff1e7e245ddad0bd92ac24a85e1%2Fapp_versions%2F1&chs=256x256)
+![QR code](qr.png)
```
bvitaliyg, line 32: Github doesn't support base64 encoding for images, so the qr image extracted to different file.

app/build.gradle
```diff
@@ -7,6 +7,7 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'com.facebook.testing.screenshot'
 
 def properties = new Properties()
+properties.putAll(System.getenv())
```
bvitaliyg, line 10: We put all the system env properties to the dictionary by default to have access to CI env tokens.

.github/workflows/ci.yml
```diff
@@ -165,19 +165,26 @@ jobs:
...
+        env:
+          KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
+          KEYSTORE_ALIAS: ${{ secrets.KEYSTORE_ALIAS }}
+          KEYSTORE_ALIAS_PASSWORD: ${{ secrets.KEYSTORE_ALIAS_PASSWORD }}
+          MAPS_API_KEY: ${{ secrets.MAPS_API_KEY }}
+          KEYSTORE_DECRYPT_KEY: ${{ secrets.KEYSTORE_DECRYPT_KEY }}
+          KEYSTORE_DECRYPT_IV: ${{ secrets.KEYSTORE_DECRYPT_IV }}
+
+      - name: Publish APK to AppCenter
+        if: contains('refs/tags/', github.ref)
```
olliecurtis, line 182: Should we also add an && condition to make sure this is when its on main branch too? as this looks to essentially say run if ref contains tags which could happen on any commit or PR that also includes any tags?

Or does this also push PR build versions too?

.github/workflows/ci.yml
```diff
@@ -165,19 +165,26 @@ jobs:
...
-#        uses: wzieba/AppCenter-Github-Action@v1
-#          with:
-#            appName: Skyscanner/Backpack-Android
-#            token: ${{secrets.APP_CENTER_TOKEN}}
-#            group: Testers
-#            file: app/build/outputs/apk/oss/release/app-oss-release.apk
-#            notifyTesters: false
-#            debug: false
+      - name: Building release APK
+        if: contains('refs/tags/', github.ref)
```
olliecurtis, line 169: Should we also add an && condition to make sure this is when its on main branch too? as this looks to essentially say run if ref contains tags which could happen on any commit or PR that also includes any tags?

Or does this also push PR build versions too?

olliecurtis (changes requested):
Looks good, just a couple of questions around when these builds are run 😄

.github/workflows/ci.yml
bvitaliyg: It was triggered by any main branch commits initially, but later on I decided that we need to sync our demo app builds with backpack releases -- to represent the latest version of the value the customers can get.

We can think about creating another app in App Center to represent the PR build for our internal review.

bvitaliyg: In any event, publishing link to PR apk build requires some research about app center API, therefore it's better to do in another ticket.

PR #566, opened 2021-04-23
[KOA-4478] Add nudger component

marianeum:
Added support for the nudger component to increase & decrease a value.

| **Light Mode** | **Dark Mode** |
|----------------|---------------|
| all | all_dm |

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+  @Before
+  fun setUp() {
+    context = InstrumentationRegistry.getInstrumentation().targetContext
+    subject = BpkNudger(context)
+  }
+
+  @Test
+  fun test_listenerIsNotInvoked_onSet() {
+    val listener = { _: Int -> }
```
bvitaliyg, line 48: Mockito needs the listener the be mocked, e.g. listener = mock<(Int) -> Unit>()

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  private val label by unsafeLazy<TextView> {
+    findViewById(R.id.bpk_nudger_label)
+  }
+
+  var value: Int = 0
+    set(value) {
+      if (field != value) {
+        if (value < minimumValue || value > maximumValue) {
```
bvitaliyg, line 55: What do you think about using Int.coerceIn(min, max) here instead?

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+
+  var minimumValue: Int = 0
+    set(minimumValue) {
+      if (minimumValue > maximumValue) {
+        throw IllegalArgumentException("Cannot set minimumValue $minimumValue when maximumValue is $maximumValue")
+      }
+      field = minimumValue
+      value = max(value, minimumValue)
```
bvitaliyg, line 69: coerceIn() can be used here as well

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,112 @@
...
+      update()
+    }
+
+  var maximumValue: Int = 100
```
bvitaliyg, line 73: We use 10 by default on web, might be a good choice.

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,112 @@
...
+      maximumValue = it.getInt(R.styleable.BpkNudger_nudgerMaximumValue, maximumValue)
+      value = it.getInt(R.styleable.BpkNudger_nudgerValue, value)
+    }
+
+    decrementButton.setOnClickListener {
+      value -= 1
+      onChangeListener?.invoke(value)
+    }
+    incrementButton.setOnClickListener {
+      value += 1
```
bvitaliyg, line 102: What do you think about using value++ here instead? :)

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,112 @@
...
+      if (field != value) {
+        if (value < minimumValue || value > maximumValue) {
+          return
+        }
+        field = value
+        update()
+      }
+    }
+
+  var minimumValue: Int = 0
```
bvitaliyg, line 63: Can we rename it to minValue and maxValue?

Backpack/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import androidx.test.platform.app.InstrumentationRegistry
+import net.skyscanner.backpack.R
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mockito.verifyNoInteractions
+
+@RunWith(AndroidJUnit4::class)
+class BpkNudgerTest {
```
bvitaliyg, line 35: You need to use test rules here in order for this to work, check out BpkDialogTest as a sample

app/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.runner.RunWith
+
+/**
+ * FIXME
+ *
+ * The buttons show up without icons. The snapshot test framework is not rendering
+ * the icons when using `setCompoundDrawablesRelativeWithIntrinsicBounds` instead of
+ * `setCompoundDrawablesWithIntrinsicBounds` on API-21 which is required to support RTL. We can add
+ * them when we update the emulators on the CI or move to a device farm for snapshot testing
+ */
```
bvitaliyg, line 35: Thank you for brining this in :)

bvitaliyg (commented):
Looks really good, I've added some comments there 👍

Backpack/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
marianeum: Thanks! Turns out I was running the wrong tests and thought they passed 🤦‍♀️

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
marianeum: Ah I got 100 from iOS

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
marianeum: Sure - again I used the same prop names as iOS - looks like web is different, though

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
marianeum: Looks like that returns an int, rather than a boolean

Backpack/src/main/java/net/skyscanner/backpack/nudger/BpkNudger.kt
marianeum: As we're only doing the max check here what would be the benefit of using `coerceIn`? I find the max check here a bit more readable/clearer what we're attempting

bvitaliyg (approved):
Great contribution, thank you!

PR #573, opened 2021-05-03
[KOA-4574] Update dependencies

marianeum:
Updated all dependencies that can be updated without any issues and created tickets for others.

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,7 +72,6 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "androidx.appcompat:appcompat:$rootProject.ext.androidxAppCompat"
   implementation "androidx.annotation:annotation:$rootProject.ext.androidxAnnotations"
-  implementation "androidx.legacy:legacy-support-v4:$rootProject.ext.androidXLegacy"
   implementation "androidx.constraintlayout:constraintlayout:$rootProject.ext.constraintLayout"
   implementation "com.google.android.material:material:$rootProject.ext.androidMaterial"
   implementation "com.jakewharton.threetenabp:threetenabp:$rootProject.ext.threeTenAbp"
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ buildscript {
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
     classpath "org.jetbrains.dokka:dokka-android-gradle-plugin:$dokka_version"
-    classpath 'com.facebook.testing.screenshot:plugin:0.13.0'
+    classpath 'com.facebook.testing.screenshot:plugin:0.14.0'
     classpath "org.jlleitschuh.gradle:ktlint-gradle:10.0.0"
   }
 }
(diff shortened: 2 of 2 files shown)


app/build.gradle
```diff
@@ -72,7 +72,6 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "androidx.appcompat:appcompat:$rootProject.ext.androidxAppCompat"
   implementation "androidx.annotation:annotation:$rootProject.ext.androidxAnnotations"
-  implementation "androidx.legacy:legacy-support-v4:$rootProject.ext.androidXLegacy"
```
marianeum, line 75: This is no longer needed

build.gradle
```diff
@@ -1,6 +1,6 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
-  ext.kotlin_version = '1.4.21'
+  ext.kotlin_version = '1.4.32'
```
marianeum, line 3: we can't quite upgrade to 1.5 yet as it has issues with lint - need to wait for the new android studio version for this

build.gradle
```diff
@@ -62,21 +62,20 @@ ext {
...
   androidxAppCompat = "1.2.0"
-  androidxAnnotations = "1.1.0"
-  androidXLegacy = "1.0.0"
+  androidxAnnotations = "1.2.0"
   androidMaterial = "1.2.1"
   cardView = "1.0.0"
-  googleMaps = "17.0.0"
-  constraintLayout = "1.1.3"
+  googleMaps = "17.0.1"
+  constraintLayout = "2.0.4"
```
marianeum, line 69: this surprisingly went without any issues (thanks screenshot testing!)

build.gradle
bvitaliyg: Does the app uses the same version?

build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
     classpath "org.jetbrains.dokka:dokka-android-gradle-plugin:$dokka_version"
-    classpath 'com.facebook.testing.screenshot:plugin:0.13.0'
+    classpath 'com.facebook.testing.screenshot:plugin:0.14.0'
```
bvitaliyg, line 16: Were there any issues with upgrading the ST library? There were some in the past.

build.gradle
marianeum: yep!

build.gradle
bvitaliyg: It may create issues in the app if they use our dependency transitively. Let's check it with app version.

build.gradle
marianeum: no issues at all, no. it'll actually be needed when we upgrade to gradle 7.0

bvitaliyg (commented):
Looks good, just few questions

build.gradle
marianeum: they're already on 2.0.4 as well

bvitaliyg (approved):

PR #575, opened 2021-05-04
KOA-4477 Add slider component

marianeum:
| **Light Mode** | **Dark Mode** |
|----------------|---------------|
| all | all_dm |

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/slider/BpkSlider.kt
```diff
@@ -0,0 +1,77 @@
...
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.util.BpkTheme
+import net.skyscanner.backpack.util.createContextThemeWrapper
+import net.skyscanner.backpack.util.use
+
+open class BpkSlider @JvmOverloads constructor(
+  context: Context,
+  attrs: AttributeSet? = null,
+  defStyleAttr: Int = 0
+) : RangeSlider(
```
marianeum, line 36: Using `RangeSlider` here, but I've added the value setter/getter/property which is the only difference between the range + default slider, so we can just have one component, rather than two

Backpack/src/main/java/net/skyscanner/backpack/slider/BpkSlider.kt
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private fun initialize(attrs: AttributeSet?, defStyleAttr: Int) {
+    context.theme.obtainStyledAttributes(
+      attrs,
+      R.styleable.BpkSlider,
+      defStyleAttr,
+      0
+    ).use {
+      val thumbColor = it.getInt(R.styleable.BpkSlider_sliderThumbColor, BpkTheme.getPrimaryColor(context))
```
bvitaliyg, line 64: You can use getColorStateList() here: in this cases customers will be able to use CSL as attributes instead.

bvitaliyg (commented):
Looks good, but I'm not sure we really need these rhombuses on the lines :D 

Is there an easy way to disable it? It does corresponds material.io website, but I've never seen it before anywhere and not sure the designers approve it.

marianeum:
yeah I asked about those here https://skyscanner.slack.com/archives/G01F5BWBVJR/p1620133794112800 - I've also never seen them before. We'll have to update to the latest version of material design, but that has a way of disabling it.

Backpack/src/main/java/net/skyscanner/backpack/slider/BpkSlider.kt
```diff
@@ -0,0 +1,78 @@
...
+      val trackColorActive = it.getInt(R.styleable.BpkSlider_sliderTrackColorActive, BpkTheme.getPrimaryColor(context))
+      val trackColorInactive =
+        it.getInt(R.styleable.BpkSlider_sliderTrackColorInactive, ContextCompat.getColor(context, R.color.bpkLine))
```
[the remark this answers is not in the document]
marianeum: i'll keep it for now, but i'll add a todo to remove, so I can write the UI tests etc without having to rely on the new emulator while you sort that out

app/src/main/res/layout/fragment_slider.xml
```diff
@@ -0,0 +1,50 @@
...
+    android:layout_marginBottom="@dimen/bpkSpacingMd"
+    android:text="Range slider"
+    app:textStyle="sm" />
+
+  <net.skyscanner.backpack.slider.BpkSlider
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:stepSize="1.0"
+    android:valueTo="100.0"
+    app:values="@array/initial_slider_values" />
```
bvitaliyg, line 48: I assume there's no way to set the values separately?

bvitaliyg (changes requested):
Looks good, but you also need to put a link to the new component docs in README.md :)

app/src/main/res/layout/fragment_slider.xml
marianeum: Unfortunately no as you could set more than just two

marianeum:
whoops - good catch!

bvitaliyg (approved):
👍

PR #586, opened 2021-05-13
Updated the testing framework.

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [x] Tests


olliecurtis (approved):
LGTM!

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -76,4 +86,37 @@ open class BpkSnapshotTest {
...
+
+    ViewHelpers.setupView(result)
+      .setExactHeightDp(height)
+      .setExactWidthDp(width)
+      .layout()
+
+    return result
+  }
+
+  // adding view to FrameLayout or creating custom ViewGroup breaks many tests for some reason.
```
marianeum, line 102: interesting - what were the errors? did you try setting the layoutparams when adding the view?

app/src/androidTest/java/net/skyscanner/backpack/card/BpkCardTest.kt
```diff
@@ -32,6 +32,7 @@ class BpkCardTest : BpkSnapshotTest() {
   @Before
   fun setup() {
     setDimensions(150, 150)
+    setBackground(android.R.color.transparent)
```
marianeum, line 35: what's the reason for setting some of these as transparent?

marianeum (commented):
This makes it so much easier to double check the dark mode screenshots!

I've noticed that the flare screenshots have a transparent background where the arrow is - do you know why that's happening?

app/src/androidTest/java/net/skyscanner/backpack/card/BpkCardTest.kt
bvitaliyg: It's hard to see a white card on a white background :)

marianeum: Maybe we could use a colour that isn't in our colour palette at all to avoid clashes

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
bvitaliyg: Yes, tried a lot. The errors might look like incorrect layout (aligning counter in Nudget component on the top but no centre), cutting off some content (TextField cuts the top and bottom parts) etc.

app/src/androidTest/java/net/skyscanner/backpack/card/BpkCardTest.kt
bvitaliyg: Yeah, I was also thinking about that :) But it has to be semantic to be contrast enough in the DM as well, so it just creates some extra colour management. With transparent BG it looks fine out of the box though both in the LM and the DM.

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
marianeum: super weird!

marianeum (approved):

PR #589, opened 2021-05-14
Fixed BpkSnackbar tests.

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/snackbar/BpkSnackbar.kt
```diff
@@ -70,6 +70,7 @@ class BpkSnackbar private constructor(
 
   private val snackbar = Snackbar.make(view, "", duration).apply {
     setBackgroundColorCompat(backgroundColor)
+    animationMode = BaseTransientBottomBar.ANIMATION_MODE_SLIDE
```
bvitaliyg, line 73: After analysing the screenshots I've found out that the only difference is the text brightness.  I've double-check our code to make sure we do set consistent colour. It looked like there's some alpha animation running on the text.

I've dive into the code and found out there's alpha animation indeed. Previously it was either not a part of library or it was restricted not to run on API 21. Disabling animations for the tests don't work as some critical callbacks are being invoked when animation is in progress.

Therefore I've just used the public API to restrict animation to slide only :)

marianeum, line 73: Does this behave any different for the user? Have we disabled animations when running tests?

marianeum (approved):

PR #591, opened 2021-05-20
Add accessibility label

marianeum:
Small thing I've noticed while testing accessibility - the settings option didn't have a label.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="1.8" />
+    <bytecodeTargetLevel target="11" />
   </component>
 </project>
--- a/Backpack/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
+++ b/Backpack/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
@@ -88,8 +88,10 @@ class BpkNudgerTest {
       invocationsCount++
       lastCurrent = current
     }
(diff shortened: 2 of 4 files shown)


.idea/compiler.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="1.8" />
+    <bytecodeTargetLevel target="11" />
```
marianeum, line 4: these always update when i do a gradle sync since the new android studio

Backpack/src/androidTest/java/net/skyscanner/backpack/nudger/BpkNudgerTest.kt
```diff
@@ -88,8 +88,10 @@ class BpkNudgerTest {
       invocationsCount++
       lastCurrent = current
     }
-    subject.value = 2
-    subject.onChangeListener = listener
+    activityRule.scenario.onActivity {
```
marianeum, line 91: i'm not sure why this suddenly started failing, but it was complaining about this done on the wrong thread

app/src/main/res/menu/settings.xml
```diff
@@ -2,7 +2,7 @@
   xmlns:app="http://schemas.android.com/apk/res-auto">
   <item
     android:id="@+id/settings_button"
-    android:title=""
+    android:title="Settings"
```
marianeum, line 5: before this would just say "unknown label"

shaundon (approved):
I think this makes sense, with the usual caveat that I don't really understand Android yet 😅

PR #595, opened 2021-05-27
BpkCalendar remake

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarDaySelectionBackground.kt
```diff
@@ -0,0 +1,171 @@
...
+
+      selectedDaySameDayCircleFillColor = it.getColorStateList(
+        R.styleable.BpkCalendar_calendarDateSelectedSameDayBackgroundColor,
+      ) ?: context.getColorStateList(R.color.__calendarSameDayBackground)
+
+      rangeBackgroundColor = it.getColorStateList(
+        R.styleable.BpkCalendar_calendarDateSelectedRangeBackgroundColor,
+      ) ?: context.getColorStateList(R.color.__calendarRangeBackground)
+    }
+    paint.strokeWidth = ResourcesUtil.dpToPx(1f, context)
```
marianeum, line 80: should this be a dimension?

Backpack/src/main/java/net/skyscanner/backpack/util/ResourcesUtil.kt
```diff
@@ -41,10 +45,14 @@ internal object ResourcesUtil {
   }
 }
 
-internal inline fun <R> TypedArray?.use(block: (TypedArray) -> R): R? {
+@OptIn(ExperimentalContracts::class)
```
marianeum, line 48: nice - i've been tempted to do that as well

Backpack/src/main/res/layout/view_bpk_calendar_day.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content"
+  android:gravity="center_horizontal|top"
+  android:orientation="vertical"
+  android:paddingBottom="24dp">
```
marianeum, line 9: this is used in a few places - should this be a dimension?

Backpack/src/main/res/layout/view_bpk_calendar_day.xml
```diff
@@ -0,0 +1,35 @@
...
+
+  <net.skyscanner.backpack.text.BpkText
+    android:id="@+id/bpk_calendar_cell_date"
+    android:layout_width="match_parent"
+    android:layout_height="36dp"
```
marianeum, line 14: same here

app/src/main/java/net/skyscanner/backpack/demo/stories/Calendar2Story.kt
```diff
@@ -0,0 +1,61 @@
...
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
+    val calendar = view.findViewById<BpkCalendar>(R.id.calendar2)!!
+
+    calendar.setParams(
+      CalendarParams(
+        range = LocalDate.now() - Period.ofYears(1)..(LocalDate.now() + Period.ofYears(1)),
+        selectionMode = CalendarParams.SelectionMode.Range,
+        cells = mapOf(
+          LocalDate.of(2020, 6, 24) to CalendarParams.Info(status = CalendarParams.Status.Highlighted),
```
marianeum, line 40: instead of fixed dates can we use current date + x days, so it'll keep working in the future?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/BpkCalendar.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlinx.coroutines.plus
+import net.skyscanner.backpack.calendar2.adapter.CalendarAdapter
+import net.skyscanner.backpack.calendar2.adapter.CalendarSpanSizeLookup
+import net.skyscanner.backpack.calendar2.data.CalendarStateMachine
+import net.skyscanner.backpack.calendar2.view.CalendarHeaderView
+import net.skyscanner.backpack.util.ResourcesUtil
+import org.threeten.bp.LocalDate
+import org.threeten.bp.Period
+
+class BpkCalendar private constructor(
```
marianeum, line 40: is there a callback for the consumer for when the selection updates? also looks like there as a scroll listener as well in the old calendar

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.threeten.bp.YearMonth
+import org.threeten.bp.format.TextStyle
+import org.threeten.bp.temporal.WeekFields
+
+data class CalendarParams(
+  val range: ClosedRange<LocalDate>,
+  val selectionMode: SelectionMode,
+  val cells: Map<LocalDate, Info> = emptyMap(),
+  val footers: List<YearMonth> = emptyList(),
+  val locale: Locale = Locale.getDefault(),
```
marianeum, line 32: what's this for? do we have cases where the calendar locale differs to the rest of the app?

marianeum (commented):
nice! took a first stab at looking at it

Backpack/src/main/java/net/skyscanner/backpack/calendar2/adapter/CalendarAdapter.kt
```diff
@@ -0,0 +1,98 @@
...
+  private val output: Consumer<CalendarDay>,
+) : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+
+  private var data: List<CalendarMonth> = input.value.months
+
+  init {
+    setHasStableIds(true)
+    input.onEach {
+      this.data = it.months
+      notifyDataSetChanged()
```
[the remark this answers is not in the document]
haroldolivieri: I might be wrong, but not sure if that would make much difference as we're using `setHasStableIds` on this adapter already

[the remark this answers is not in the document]
marianeum: yes I think you're right - it's been a long time since I used stable ids - afaik stable ids were mainly introduced to make it easier to migrate from `ListView` - I remember Yigit (who did a lot of the original `RecyclerView` work) mentioned that he regrets adding support for it - using diffutil is the recommended approach (and would also remove the need for `YearMonth`

Backpack/src/main/java/net/skyscanner/backpack/util/StateMachine.kt
```diff
@@ -0,0 +1,86 @@
...
+
+  val state = MutableStateFlow(initial)
+  val effects = MutableSharedFlow<Effect>()
+  val mutex = Mutex()
+
+  return object : MutableStateMachine<State, Effect> {
+
+    override val effects: SharedFlow<Effect> = effects
+
+    override val state: StateFlow<State> = state
```
haroldolivieri, line 61: what is the main reason for having a hot flow of data for the state in a Calendar? Do we have cases where it has to change in reaction to external interactions? My two cents is that a listener method or interface would cover the job here in a much simpler way.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/adapter/CalendarAdapter.kt
```diff
@@ -19,47 +19,66 @@
...
   private var data: List<CalendarMonth> = emptyList()
 
+  var footerAdapter: CalendarFooterAdapter<*> = DefaultCalendarFooterAdapter
+    set(value) {
+      field = value
+      notifyDataSetChanged()
+    }
+
   init {
     setHasStableIds(true)
```
marianeum: i think you might not need stable ids + overwriting the `getItemId` method anymore if you're using diffutil

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
```diff
@@ -34,7 +34,7 @@ internal sealed class CalendarCell {
   abstract val id: Long
 }
 
-internal data class CalendarSpace(
+internal data class CalendarCellSpace(
```
marianeum, line 37: I like to put data classes for sealed classes within the sealed class and then you get the `CalendarCell` name from the sealed class and can name the data class just `Space` - e.g. you'd refer to it as `CalendarCell.Space`

Backpack/src/main/java/net/skyscanner/backpack/calendar2/BpkCalendar.kt
```diff
@@ -73,83 +81,38 @@ class BpkCalendar private constructor(
...
-      addOnScrollListener(object : RecyclerView.OnScrollListener() {
-        override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
-
-          val firstItemPosition = calendarLayoutManager.findFirstVisibleItemPosition()
-          val item = state.value.cells[firstItemPosition]
-
-          scrollListeners.forEach {
-            it.invoke(item.yearMonth)
-          }
+        scrollListeners.forEach {
```
marianeum, line 95: i actually had a look and the scroll listener isn't used at all right now, so maybe we don't need to add it until somebody needs it and we know what they actually need

Backpack/src/main/res/layout/view_bpk_calendar_2.xml
```diff
@@ -0,0 +1,46 @@
...
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="match_parent"
+  tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout">
+
+  <net.skyscanner.backpack.calendar2.view.CalendarHeaderView
+    android:id="@+id/bpk_calendar_header"
+    android:layout_width="0dp"
+    android:layout_height="50dp"
```
marianeum, line 12: should this be a dimension?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarMonth.kt
```diff
@@ -41,7 +41,7 @@ internal inline fun CalendarMonth(
   locale: Locale,
   monthsTextStyle: TextStyle,
   weekFields: WeekFields,
-  footers: List<YearMonth>,
+  footers: Set<YearMonth>,
```
marianeum, line 44: wouldn't we want the footers be a list to have an order?

Backpack/src/main/java/net/skyscanner/backpack/util/ExperimentalBackpackApi.kt
```diff
@@ -0,0 +1,21 @@
...
+  AnnotationTarget.LOCAL_VARIABLE,
+  AnnotationTarget.VALUE_PARAMETER,
+  AnnotationTarget.CONSTRUCTOR,
+  AnnotationTarget.FUNCTION,
+  AnnotationTarget.PROPERTY_GETTER,
+  AnnotationTarget.PROPERTY_SETTER,
+  AnnotationTarget.TYPEALIAS,
+)
+@MustBeDocumented
+annotation class ExperimentalBackpackApi
```
marianeum, line 21: nice!

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
```diff
@@ -21,28 +21,44 @@ package net.skyscanner.backpack.calendar2.data
...
+  abstract val id: Long
 }
 
 internal data class CalendarSpace(
   val selected: Boolean,
   override val yearMonth: YearMonth,
-) : CalendarCell()
+) : CalendarCell() {
+
+  override val id: Long =
```
marianeum, line 42: if we don't need a "global" id for the adapter we could also first check the class before comparing the id with diffutil and then we won't have to the math here

marianeum:
when it comes to accessibility - is there a way to quickly go through the months or do you have to navigate through the days (or at least rows)? i remember that was a big painpoint in accessibility testing

app/src/main/java/net/skyscanner/backpack/demo/stories/Calendar2Story.kt
```diff
@@ -83,40 +85,40 @@ class Calendar2Story : Story() {
...
         val price = random.nextInt(100)
-        CalendarParams.Info(
+        CellInfo(
           status = when (price) {
-            in 0..25 -> CalendarParams.Status.Empty
-            in 25..50 -> CalendarParams.Status.Positive
-            in 50..75 -> CalendarParams.Status.Neutral
-            in 75..100 -> CalendarParams.Status.Negative
-            else -> CalendarParams.Status.Empty
+            in 0..25 -> CellStatus.Empty
```
marianeum: maybe these could be defined as constants (especially as we have to define them twice as well)

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -29,7 +29,9 @@ import org.threeten.bp.temporal.WeekFields
 data class CalendarParams(
   val range: ClosedRange<LocalDate>,
   val selectionMode: SelectionMode,
-  val cellInfo: Map<LocalDate, CellInfo> = emptyMap(),
+  val cellsInfo: Map<LocalDate, CellInfo> = emptyMap(),
+  val defaultCellInfo: CellInfo = CellInfo(),
+  val disabledDates: Set<LocalDate> = emptySet(),
```
marianeum, line 34: how come this is extracted?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/adapter/CalendarAdapter.kt
bvitaliyg: TBH I see no harm in that and assume that this still may improve the performance in some cases

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
bvitaliyg: In was done in this manner, but I moved in out in order not to have too much code in the parent classes. What do you think, shall we put it back?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
bvitaliyg: We're also using the ID to determine whether the items the same in DiffUtil

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
bvitaliyg: We use CellInfo as default as well now (for instance, to provide a dash for the cells without price). If disabling information will be available inside the info it'll be possible to disable all the cells in the calendar, which I believe not something we should "promote".

Therefore it was extracted.

Backpack/src/main/java/net/skyscanner/backpack/util/StateMachine.kt
bvitaliyg: You may need to have a sync access to the state, as well as do some filtering operations as well. You may also want to know about some parameters changes or have some access to it.

This may be hidden beyond the listener, but I believe working with state is much better options. Thoughts?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
marianeum: I feel like we're mixing configuration/state with other information here - I think this comes back to all the logic being in data classes, rather than having a viewmodel or similar where it would be easier to inject things like a locale or the current time (I believe `now` is not used here either - can this be removed?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
marianeum: I generally prefer it as the context is clearer and if you rename `CalendarCell` you only have to rename it once, rather than in every implementation

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarMonth.kt
marianeum: we are creating cells for each footer in `CalendarMonth`, though, so it is being used as a list, rather than a set

Backpack/src/main/java/net/skyscanner/backpack/util/ExperimentalBackpackApi.kt
marianeum: can you mark all files in the package with this annotation? otherwise everything is red right now as it's using the experimental apis which makes it hard to parse the code

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
marianeum: that's what i meant - in `DiffUtil` we can first check if it's the same type, then compare the id. This would be an advantage of removing the stable id implementation as this way of generating the ids seems error prone if we extend it in the future

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
marianeum: Maybe instead of a separate array having the `disabled` state be defined as part of `CellInfo` (but outside of the `status`?

Backpack/build.gradle
```diff
@@ -37,18 +37,23 @@ android {
...
   }
+  kotlinOptions {
+    jvmTarget = JavaVersion.VERSION_1_8.toString()
+  }
 }
 
 dependencies {
   implementation "androidx.annotation:annotation:$rootProject.ext.androidxAnnotations"
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+  implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.0"
```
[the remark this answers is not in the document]
bvitaliyg: Coroutines versioning is different from Kotlin :)

Backpack/src/main/java/net/skyscanner/backpack/calendar2/BpkCalendar.kt
marianeum: I'm not sure if we should build features that aren't actually needed and part of the design system. i'd rather keep it simple and then add what is needed later on. looking at the component in figma it looks like we don't have a footer support right now (but there is a legend at the top that the new version doesn't have)

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -0,0 +1,73 @@
...
+  enum class SelectionMode {
+    Disabled,
+    Single,
+    Range,
+  }
+}
+
+@ExperimentalBackpackApi
+data class CellInfo(
+  val status: CellStatus? = null,
```
marianeum, line 56: we probably need a way to announce the positive/neutral/negative status for accessibility as i think currently there's no way to know about that with voiceover

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
marianeum: I'm actually confused how this works with the spacers as there's multiple spacers who have the same id as they have the same `YearMonth`, right?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarAdapter.kt
```diff
@@ -61,32 +53,22 @@ internal class CalendarAdapter(
...
   override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int) =
-    if (holder is ItemHolder<*>) {
-      holder as ItemHolder<CalendarCell>
-      holder.invoke(data[position])
-    } else {
-      val footers = footerAdapter as CalendarFooterAdapter<RecyclerView.ViewHolder>
-      val item = data[position] as CalendarCell.Footer
-      footers.onBindViewHolder(holder, item.yearMonth)
-    }
+    (holder as ItemHolder<CalendarCell>).invoke(data[position])
```
marianeum, line 67: You could update the viewholder generic in the constructor to avoid the casting here now that the footer adapter is gone

Backpack/src/main/java/net/skyscanner/backpack/util/StateMachine.kt
marianeum: A callback would be helpful to actually be notified about _user_ updates, rather than updates the clients made - for example if the client updates some data on the calendar that is different to a user selecting a different range

Backpack/src/main/res/layout/view_bpk_calendar_day.xml
marianeum: i think this is still hardcoded in the other views

Backpack/src/main/res/layout/view_bpk_calendar_day.xml
marianeum: i think this is still hardcoded in the other views

Backpack/src/main/res/layout/view_bpk_calendar_day.xml
bvitaliyg: I've double-checked it, looks okay

Backpack/src/main/res/layout/view_bpk_calendar_day.xml
bvitaliyg: Same

Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarDayLabelContentColor.kt
```diff
@@ -0,0 +1,45 @@
...
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.calendar2.CellStatus
+
+internal typealias CalendarDayLabelContentColor = (CellStatus?) -> ColorStateList
+
+internal fun CalendarDayLabelContentColor(
+  context: Context,
+): CalendarDayLabelContentColor {
+
+  val default = context.getColorStateList(R.color.bpkTextSecondary)
```
marianeum, line 32: do these need theming support?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarDayStatusBackground.kt
```diff
@@ -0,0 +1,98 @@
...
+
+private class CalendarDayStatusDrawable(context: Context) : Drawable() {
+
+  var status: CellStatus? = null
+    set(value) {
+      field = value
+      invalidateSelf()
+    }
+
+  private val colorPositive = context.getColorStateList(R.color.__calendarCellStylePositiveColor)
```
marianeum, line 54: same here

Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarDayStatusContentColor.kt
```diff
@@ -0,0 +1,47 @@
...
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.calendar2.CellStatus
+
+internal typealias CalendarDayStatusContentColor = (CellStatus?) -> ColorStateList
+
+internal fun CalendarDayStatusContentColor(
+  context: Context,
+): CalendarDayStatusContentColor {
+
+  val default = context.getColorStateList(R.color.bpkTextPrimary)
```
marianeum, line 32: and here

Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarHeaderView.kt
```diff
@@ -0,0 +1,83 @@
...
+    orientation = HORIZONTAL
+  }
+
+  private var lastState: CalendarParams? = null
+
+  override fun invoke(state: CalendarParams) {
+    if (lastState == state) return
+
+    lastState = state
+    removeAllViews()
```
marianeum, line 55: this seems quite inefficient to remove and re-add all views whenever the calendar params change - the numbers of days a week are always the same - just the first day might be different, so the views could be added either in the layout or once and then just updated based on the first day and text

Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarDayLabelContentColor.kt
bvitaliyg: We keep the same approach as in original calendar and theme only those properties which were themed there. I'd removed that theming as well s it doesn't seem to be used, but I'd prefer to do these changes gradually.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/view/CalendarHeaderView.kt
bvitaliyg: Done

marianeum (approved):

PR #598, opened 2021-06-08
KOA-4626 Add lint check for component usage

marianeum:
Screenshot 2021-06-08 at 14 50 48

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/build.gradle
```diff
@@ -64,6 +64,8 @@ dependencies {
   androidTestImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:$rootProject.ext.mockitoKotlin"
   androidTestImplementation "org.mockito:mockito-android:$rootProject.ext.mockitoAndroid"
   testImplementation "junit:junit:$rootProject.ext.junit"
+
+  lintPublish project(':backpack-lint')
```
marianeum, line 68: this publishes the lint checks with backpack, so when you integrate backpack the lint checks are added

backpack-lint/src/main/java/net/skyscanner/backpack/lint/IssueRegistry.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package net.skyscanner.backpack.lint
+
+import com.android.tools.lint.client.api.IssueRegistry
+import com.android.tools.lint.detector.api.CURRENT_API
+import com.android.tools.lint.detector.api.Issue
+import net.skyscanner.backpack.lint.check.BpkComponentUsageDetector
+
+@Suppress("unused", "UnstableApiUsage")
+class IssueRegistry : IssueRegistry() {
```
marianeum, line 27: we need to keep a list of our lint checks in a registry, so they get picked up by the system

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.intellij.psi.PsiMethod
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.UClass
+import org.w3c.dom.Element
+
+@Suppress("UnstableApiUsage")
+class BpkComponentUsageDetector : Detector(), SourceCodeScanner, XmlScanner {
+
+  companion object {
+    val ISSUE = Issue.create(
```
marianeum, line 43: This is the overall issue definition. When reporting there's a specific message with some further information about the actual component

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+      severity = Severity.WARNING,
+      implementation = Implementation(
+        BpkComponentUsageDetector::class.java,
+        Scope.JAVA_AND_RESOURCE_FILES
+      )
+    )
+
+    private const val BPK_BUTTON = "net.skyscanner.backpack.BpkButton"
+
+    private val COMPONENTS = mapOf(
```
marianeum, line 57: this is a map of native components to backpack component. The native component is used as the key so we can use it as a lookup table to find the component to replace it with. There can be multiple native components - e.g. for the button there's the default button, an appcompat version and for xml just the button without any package name (the native components generally don't specify the package name in xml)

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+      "Button" to BPK_BUTTON,
+      "androidx.appcompat.widget.AppCompatButton" to BPK_BUTTON,
+    )
+
+    private val APPLICABLE_TYPES = COMPONENTS.keys.toList()
+  }
+
+  private val classCache = mutableListOf<String>()
+
+  override fun getApplicableConstructorTypes(): List<String> = APPLICABLE_TYPES
```
marianeum, line 69: these methods tell lint which constructors, classes and elements we're interested in for checks

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+
+  override fun applicableSuperClasses(): List<String> = APPLICABLE_TYPES
+
+  override fun getApplicableElements(): Collection<String> = APPLICABLE_TYPES
+
+  override fun appliesTo(folderType: ResourceFolderType): Boolean = folderType == ResourceFolderType.LAYOUT
+
+  override fun visitClass(context: JavaContext, declaration: UClass) {
```
marianeum, line 77: this covers extending a native component vs using the backpack component.

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+    val superClass = declaration.javaPsi.superClass?.qualifiedName ?: return
+    val component = COMPONENTS[superClass]
+    if (component != null) {
+      if (declaration.name == null || classCache.contains(declaration.name)) {
+        // in some cases we might have two relevant super classes (e.g. when there's an AppCompat and default component) - only report issue once
```
marianeum, line 82: As `AppCompatButton` extends `Button` the `applicableSuperClasses` check gets triggered twice. This is the best way I could find to fix this - we _do_ want to check that if you're extending either `Button` or `ApCompatButton` we trigger the check, but there doesn't seem to be any hint about which one triggered, so we can compare it with the class that was actually extended. So keeping a memory of checks that already is the only thing I could think of to not report the error twice

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+        return
+      }
+      classCache.add(declaration.name!!)
+      context.report(context.getNameLocation(declaration), superClass, component)
+    }
+  }
+
+  override fun visitConstructor(context: JavaContext, node: UCallExpression, constructor: PsiMethod) {
```
marianeum, line 90: this checks that no native components are created in code

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -0,0 +1,108 @@
...
+    val className = node.returnType?.canonicalText ?: return
+    val component = COMPONENTS[className]
+    if (component != null) {
+      context.report(context.getLocation(node), className, component)
+    }
+  }
+
+  override fun visitElement(context: XmlContext, element: Element) {
```
marianeum, line 98: and this checks that they're not used in xml

Backpack/build.gradle
olliecurtis: Does this mean anyone using Backpack Android does not need to worry about enabling the rules and if you are using the library 'it just works'?

olliecurtis (changes requested):
Looks good, wondering @marianeum do you know if its possible to provide a link to the docs for the consumer to link to so they can read more about the component and make changes to?

In your example for the button maybe a link to the [Android docs](https://backpack.github.io/android/versions/25.3.0/net.skyscanner.backpack.button/index.html) or even just a link to the [docs site component page](https://backpack.github.io/components/button/?platform=android)?

Backpack/build.gradle
marianeum: correct - as long as they use lint the checks will run

olliecurtis (approved):
LGTM! Great work

PR #601, opened 2021-06-14
KOA-4634 Fix publishing for OSS library

### LINKED ISSUE #600: Android resource linking failed
I'm trying to import Backpack into my project and I'm getting resource linking errors. I'm using version 29.0.0 because at the time of writing, the newer ones aren't building successfully on Jitpack.

```
/Users/amunim/.gradle/caches/transforms-3/13cf8cefe233e09aab453a1214846e51/transformed/jetified-backpack-android-29.0.0/res/values/values.xml: AAPT: error: resource font/skyscanner_relative_android_book (aka com.example:font/skyscanner_relative_android_book) not found.
    
/Users/amunim/.gradle/caches/transforms-3/13cf8cefe233e09aab453a1214846e51/transformed/jetified-backpack-android-29.0.0/res/values/values.xml: AAPT: error: resource font/skyscanner_relative_android_bold (aka com.example:font/skyscanner_relative_android_bold) not found.
    
/Users/amunim/.gradle/caches/transforms-3/13cf8cefe233e09aab453a1214846e51/transformed/jetified-backpack-android-29.0.0/res/values/values.xml: AAPT: error: resource font/skyscanner_relative_android_black (aka com.example:font/skyscanner_relative_android_black) not found.
```

If I add some arbitrary fonts (with those names) in my `res/font` directory, I get past that error and onto another issue:

```
AAPT: error: resource style/BpkTheme (aka com.example:style/BpkTheme) not found.
error: failed linking references.
```

Instead of using `BpkTheme`, I can use `BpkDefaultTheme` with no issue (is the README outdated?), but it's also necessary not to use `AppCompatActivity`, nor any Material widgets like `MaterialButton`. Then it builds and runs.

---

I'm unsure if that's the _correct_ approach though, and whether I'm missing things because of these workarounds.

marianeum:
Publishing of the OSS variant broke due to various dependency updates. Removed the need to publish two separate versions for internal & OSS use. Fixes #600  

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


.github/workflows/ci.yml
```diff
@@ -58,9 +58,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        module:
-          - app
-          - Backpack
+        module: [app, Backpack]
+        include:
```
marianeum, line 62: app still needs the variant, backpack doesn't

Backpack/gradle-maven-push.gradle
```diff
@@ -35,10 +35,10 @@ afterEvaluate { project ->
             maven(MavenPublication) {
                 groupId 'com.github.skyscanner'
                 artifactId 'backpack-android'
-                version "$version-internal"
+                version "$version"
```
marianeum, line 38: no longer need a separate `internal` version

Backpack/build.gradle
```diff
@@ -20,18 +20,7 @@ android {
...
-      dimension "version"
-      versionNameSuffix "-oss"
-    }
-    internal {
-      dimension "version"
-      versionNameSuffix "-internal"
-    }
-  }
-  defaultPublishConfig "ossRelease"
+  defaultPublishConfig "release"
```
marianeum, line 23: this defined the two separate flavours which are no longer needed

Backpack/src/main/res/values/themes.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-  <style name="BpkThemeBase" parent="Theme.MaterialComponents.DayNight.NoActionBar.Bridge">
+  <style name="BpkTheme" parent="Theme.MaterialComponents.DayNight.NoActionBar.Bridge">
```
marianeum, line 4: we don't have two separate variants anymore, so no longer need the base

Backpack/src/main/res/values/themes.xml
```diff
@@ -10,6 +10,10 @@
     <item name="windowActionBar">false</item>
     <item name="colorButtonNormal">@android:color/white</item>
 
+    <item name="bpkFontFamilyBase">sans-serif</item>
```
marianeum, line 13: we do need to define some config for fonts, so you can use the library - these can be overridden by apps

app/build.gradle
```diff
@@ -41,6 +41,17 @@ android {
       signingConfig signingConfigs.release
     }
   }
+  flavorDimensions "version"
```
marianeum, line 44: the demo app now has the two variants

app/src/internal/res/values/themes.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <style name="BpkTheme" parent="BpkThemeBase">
+  <style name="AppThemeBase" parent="BpkTheme">
```
marianeum, line 3: and a base theme, so we can override the fonts per variant - this is the internal one that moved to the app

build.gradle
```diff
@@ -48,6 +48,11 @@ subprojects {
   }
 }
 
+task install {
```
marianeum, line 51: this fixes publishing to jitpack. this is probably a temporary solution - this essentially replicates the plugin that was used in the past that is deprecated to avoid having to mess with the build task (which attempts to build the whole project). I suspect eventually they will remove support for the install task

olliecurtis (approved):

PR #602, opened 2021-06-15
KOA-4634 Update font definition

marianeum:
When using relative font we still need to provide some different font attributes to serif. Provide a relative theme to simplify the usage.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/res/values/backpack.text.xml
```diff
@@ -45,6 +45,10 @@
     <item name="android:letterSpacing">-0.0125</item>
   </style>
 
+  <style name="bpkTextBaseRelative" parent="bpkTextBase">
```
marianeum, line 48: I am 90% sure that this actually doesn't do anything as the line spacing multiplier isn't part of the text appearance, but to avoid breaking some usages outside of our components I kept these

Backpack/src/main/res/values/themes.xml
```diff
@@ -35,4 +35,26 @@
     <item name="bpkTextXxxlHeavyAppearance">@style/bpkTextXxxlHeavySerif</item>
   </style>
 
+  <style name="BpkTheme.Relative">
```
marianeum, line 38: At first this was defined in the app theme, but this would require quite a lot of duplicated definitions when the only thing that needs to be outside of the library is the actual font definitions.

Backpack/src/main/res/values/backpack.font.xml
```diff
@@ -1,79 +1,80 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <!-- Text should have no lineSpacingMultiplier to work with relative -->
-  <style name="bpkTextBaseInternal" parent="bpkTextBase">
+
+  <style name="bpkTextBaseRelative" parent="bpkTextBase">
```
marianeum, line 4: I am 90% sure that these don't actually don't do anything, but keeping it in case somebody is doing something outside of our components that requires this

olliecurtis (approved):

PR #603, opened 2021-06-16
KOA-4635 Add snapshot testing for icons

marianeum:
Added a new snapshot test for icons to make it easier to spot errors when updating icons.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


app/src/androidTest/java/net/skyscanner/backpack/icon/BpkIconTest.kt
```diff
@@ -0,0 +1,87 @@
...
+class BpkIconTest : BpkSnapshotTest() {
+
+  @Before
+  fun setup() {
+    setDimensions(600, 300)
+  }
+
+  @Test
+  fun testAllSmallIcons() {
+    assumeTrue(BpkTestVariant.current == BpkTestVariant.Default)
```
marianeum, line 44: this runs it only on the default variant as theming doesn't affect the raw icons

app/src/androidTest/java/net/skyscanner/backpack/icon/BpkIconTest.kt
```diff
@@ -0,0 +1,87 @@
...
+    icons.forEach { icon ->
+      val imageView = ImageView(testContext).apply { setImageDrawable(icon) }
+      layout.addView(imageView)
+    }
+    return layout
+  }
+
+  private fun getIcons(small: Boolean): List<Drawable> {
+    return R.drawable::class.java.fields.mapNotNull { field ->
+      if (field.name.startsWith("bpk_") && field.name.endsWith("_sm") == small) {
```
marianeum, line 76: this finds all relevant icons. unfortunately our naming isn't very strict - we've got a few other drawables in the project which start with `bpk_`, but the check for `VectorDrawable` below catches those. i was considering also checking the size, but as the size might change that won't work going forward - might have to revisit this in the future or avoid prefixing other drawables with `bpk_` or at least giving them some other prefix we can exclude

olliecurtis (approved):

PR #604, opened 2021-06-18
KOA-3596 Improve accessibility for star rating component

marianeum:
The star rating component was lacking accessibility - the interactive component was not possible to be controlled while using talkback. This adds accessibility support to both the static and interactive rating control by adding a new `accessibilityStatus` attribute to be set by consumers. For the interactive component there is a big difference between Android 11/sdk version 30 and below as a new API was added to improve announcements for controls.

(sorry the video + audio is massively out of sync - thank Android Studio for that)

https://user-images.githubusercontent.com/1529791/122533410-72941800-d019-11eb-9c34-23c2fc5e6f97.mp4

https://user-images.githubusercontent.com/1529791/122533402-7031be00-d019-11eb-8606-3d9934aa61ec.mp4

https://user-images.githubusercontent.com/1529791/122533413-732cae80-d019-11eb-8e2b-120b1d612cf8.mp4

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
@@ -19,9 +19,17 @@
 package net.skyscanner.backpack.starrating
 
 import android.content.Context
+import android.os.Build
+import android.os.Bundle
 import android.util.AttributeSet
 import android.view.MotionEvent
 import android.view.View
+import android.view.accessibility.AccessibilityEvent
+import android.view.accessibility.AccessibilityNodeInfo
+import android.view.accessibility.AccessibilityNodeInfo.AccessibilityAction
+import android.view.accessibility.AccessibilityNodeInfo.RangeInfo
+import android.widget.RatingBar
+import android.widget.SeekBar
 import kotlin.math.max
 import kotlin.math.roundToInt
 import net.skyscanner.backpack.R
@@ -31,7 +39,7 @@ import net.skyscanner.backpack.util.createContextThemeWrapper
 open class BpkInteractiveStarRating @JvmOverloads constructor(
   context: Context,
   attrs: AttributeSet? = null,
-  defStyleAttr: Int = 0
+  defStyleAttr: Int = 0,
 ) : BpkStarRatingBase(
   context = createContextThemeWrapper(context, attrs, R.attr.bpkInteractiveStarRatingStyle),
   attrs = attrs,
@@ -41,6 +49,12 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
   full = R.drawable.bpk_star,
   starSize = context.resources.getDimensionPixelSize(R.dimen.bpkSpacingXxl)
 ) {
+
+  init {
+    if (importantForAccessibility == IMPORTANT_FOR_ACCESSIBILITY_AUTO) {
+      importantForAccessibility = IMPORTANT_FOR_ACCESSIBILITY_YES
+    }
+  }
 
   var onRatingChangedListener: ((Float, Float) -> Unit)? = null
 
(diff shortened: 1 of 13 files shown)


Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
```diff
@@ -76,6 +90,65 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
     return true
   }
 
+  override fun performAccessibilityAction(action: Int, arguments: Bundle?): Boolean {
```
marianeum, line 93: this handles the volume control interaction to set the rating

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
```diff
@@ -76,6 +90,65 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
...
+
+    return false
+  }
+
+  override fun onInitializeAccessibilityEvent(event: AccessibilityEvent) {
+    super.onInitializeAccessibilityEvent(event)
+    event.itemCount = maxRating
+    event.currentItemIndex = rating.toInt()
+    if (event.eventType == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED) {
+      updateStateDescription()
```
marianeum, line 119: ensure we also set the state description when focusing the view - this only works in api level 30+ even when using `accounceForAccessibility` below 30 (which works in other cases)

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
```diff
@@ -76,6 +90,65 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
...
+    }
+  }
+
+  override fun getAccessibilityClassName(): CharSequence {
```
marianeum, line 123: on pre-30 the accessibility class name somehow decides what controls are available

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkInteractiveStarRating.kt
```diff
@@ -76,6 +90,65 @@ open class BpkInteractiveStarRating @JvmOverloads constructor(
...
+    return RatingBar::class.java.name
+  }
+
+  override fun onInitializeAccessibilityNodeInfo(info: AccessibilityNodeInfo) {
+    super.onInitializeAccessibilityNodeInfo(info)
+    val rangeInfo = RangeInfo.obtain(
```
marianeum, line 129: set the range possible and allow forward/backwards controls to control the component

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkStarRating.kt
```diff
@@ -38,6 +39,18 @@ open class BpkStarRating @JvmOverloads constructor(
   starSize = context.resources.getDimensionPixelSize(R.dimen.bpkSpacingBase)
 ) {
 
+  override fun onInitializeAccessibilityNodeInfo(info: AccessibilityNodeInfo) {
+    super.onInitializeAccessibilityNodeInfo(info)
+    val text = getAccessibilityText()
+    if (text != null) {
+      if (info.contentDescription != null) {
+        info.contentDescription = "$text ${info.contentDescription}"
```
marianeum, line 47: This is an annoying hack - i couldn't get both an announcement and the standard content description being announced (at least not pre-30) - pre-30 without a content description nothing was announced, so use the accessibility text and append the content description if available, otherwise only announce the accessibility text

Backpack/src/main/java/net/skyscanner/backpack/starrating/internal/BpkStarRatingBase.kt
```diff
@@ -92,6 +101,9 @@ open class BpkStarRatingBase internal constructor(
     update()
   }
 
+  protected fun getAccessibilityText(): String? =
```
marianeum, line 104: this allows a `x out of y` announcement rather than the default x% out of 100% announcement - this only works for 30+, however when using the control and increasing/decreasing stars it does announce this, rather than the percentage

Backpack/src/main/java/net/skyscanner/backpack/starrating/BpkStarRating.kt
olliecurtis: Might be worth adding the above or similar as a comment alongside this code so that we have a record for the future as why its like this?

olliecurtis (changes requested):
Looks good to me, just one small comment above but otherwise looks fine and the videos show it working!

https://github.com/Skyscanner/backpack-android/pull/604#discussion_r654564591

olliecurtis (approved):
LGTM!

PR #611, opened 2021-07-07
KOA-4605 Update Button component to match Figma

marianeum:
This updates the button component to match the Figma design system based on the Figma audit. Mainly colour changes, but also some spacing changes as well as updating the text to not be all caps anymore - e.g. here's the new destructive button:

Screenshot_1625659871

Also took the chance to clean up some of the icon positioning logic and use the `MaterialButton` instead as that's all possible now.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonLayout.kt
```diff
@@ -1,138 +0,0 @@
...
-      return
-    }
-
-    val icon = compoundDrawables[0] ?: compoundDrawables[2]
-    val iconWidth = icon?.intrinsicWidth ?: 0
-    val iconSpacing = if (icon != null && text.isNotEmpty()) maxIconSpacing else 0
-
-    super.setCompoundDrawablePadding(iconSpacing)
-
-    if (icon != null) {
```
marianeum, line 127: we used to manually update the button position to align the icon with the text, rather than the end/start of the view. this is now supported by `MaterialButton` so in the next commit i'll change to the materialbutton which also removes a lot of the logic in general

app/src/main/res/layout/fragment_button_destructive.xml
```diff
@@ -49,7 +49,7 @@
     android:layout_width="wrap_content"
     android:layout_height="wrap_content"
     android:text="With icon"
-    app:buttonIcon="@drawable/bpk_long_arrow_right"
+    app:buttonIcon="@drawable/bpk_long_arrow_right_sm"
```
marianeum, line 52: just for completeness use the right icon

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -90,11 +89,10 @@ open class BpkButton(
   private val paddingVertical = resources.getDimensionPixelSize(R.dimen.bpkSpacingMd) +
     (resources.getDimensionPixelSize(R.dimen.bpkBorderSizeLg) / 2)
 
-  final override var icon: Drawable? = null
-    set(value) {
-      field = value
-      updateIconState()
-    }
+  override fun setIcon(icon: Drawable?) {
```
marianeum, line 92: that's from `MaterialButton`

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -68,11 +67,11 @@ open class BpkButton(
       var paddingHorizontal = paddingHorizontal
       val paddingVertical = paddingVertical
 
-      if (iconPosition == ICON_ONLY) {
+      if (value == ICON_ONLY) {
         paddingHorizontal = paddingHorizontalIconOnly
-        compoundDrawablePadding = 0
+        iconPadding = 0
```
marianeum, line 72: that's from `MaterialButton`

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -157,16 +155,6 @@ open class BpkButton(
     applyStyle(style)
   }
 
-  override fun setTextColor(color: Int) {
```
marianeum, line 153: moved to the base class as it was duplicated across link + button

Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
```diff
@@ -19,16 +19,12 @@
...
 
 import android.content.Context
-import android.graphics.drawable.Drawable
 import android.text.TextUtils
 import android.util.AttributeSet
 import android.view.Gravity
+import net.skyscanner.backpack.R
 import net.skyscanner.backpack.text.BpkText
 
-internal const val ICON_POSITION_START = 0
```
marianeum, line 26: moved to the icon base class as well as that's where it's used

Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
```diff
@@ -40,13 +36,13 @@ abstract class BpkButtonBase internal constructor(
 
   abstract var iconPosition: Int
 
-  abstract var icon: Drawable?
```
marianeum, line 41: that's in the material button now

Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonBase.kt
```diff
@@ -40,13 +36,13 @@ abstract class BpkButtonBase internal constructor(
...
-
   init {
     maxLines = 1
     gravity = Gravity.CENTER
     ellipsize = TextUtils.TruncateAt.END
     isClickable = isEnabled
+    iconSize = resources.getDimensionPixelSize(R.dimen.bpkSpacingBase)
```
marianeum, line 44: this automatically resizes the icon to the right side, so we no longer have to do this manually

Backpack/src/main/java/net/skyscanner/backpack/button/internal/BpkButtonWithIcon.kt
```diff
@@ -19,40 +19,31 @@
...
   attrs: AttributeSet?,
-  defStyleAttr: Int
-) : AppCompatButton(context, attrs, defStyleAttr) {
+  defStyleAttr: Int,
+) : MaterialButton(context, attrs, defStyleAttr) {
 
   internal var iconDrawablePosition: Int = ICON_POSITION_END
     set(value) {
       field = value
+      iconGravity = when (value) {
```
marianeum, line 40: this aligns the icon with the text for us

app/src/main/res/layout/fragment_button_icon_alignment.xml
```diff
@@ -40,7 +40,7 @@
     android:layout_margin="@dimen/bpkSpacingSm"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:text="Icon end with a really long text abcdefghi"
+    android:text="Icon end with a really long text abcdefghijklmnopqrstuvwxyz"
```
marianeum, line 43: the text wasn't long enough without all caps :D

Backpack/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonLinkTest.kt
```diff
@@ -49,8 +49,8 @@ class BpkButtonLinkTest {
   fun test_icon_end() {
     val trainIcon = AppCompatResources.getDrawable(activity, R.drawable.bpk_train)
     val button = BpkButtonLink(activity).apply {
-      icon = trainIcon
       iconPosition = BpkButton.END
+      icon = trainIcon
```
marianeum, line 53: the icon needs to be set after the icon position in tests - this is not the case when running the app, but is due to how the code is updated when no layout was performed like in this test

olliecurtis (approved):
LGTM!

UNRELEASED.md
```diff
@@ -3,6 +3,13 @@
 > Place your changes below this line.
 >
 
+**Fixed:**
+
+- Updated Button component to match Design System:
+  - Updated various colours
+  - Updated icon spacing + horizontal padding
+  - Removed all caps property
```
bvitaliyg, line 11: We need to mention that it's false by default now, as this is kinda breaking change.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
bvitaliyg: This may potentially be breaking change as we replace property with setter. It may not be automatically converted to property, let's just double-check the compatibility for Kotlin here.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -68,29 +67,32 @@ open class BpkButton(
...
+      if (value == ICON_ONLY) {
+        paddingHorizontal = paddingHorizontalIconOnly
+        iconPadding = 0
       } else {
-        compoundDrawablePadding = resources.getDimensionPixelSize(R.dimen.bpkSpacingSm)
+        iconPadding = resources.getDimensionPixelSize(R.dimen.bpkSpacingMd)
       }
 
-      setPadding(paddingHorizontal, paddingVertical, paddingHorizontal, paddingVertical)
+      setPaddingRelative(paddingHorizontal, paddingVertical, paddingHorizontal, paddingVertical)
```
bvitaliyg, line 77: Why did we switch here to padding relative?

app/src/main/res/layout/fragment_buttons_changeable.xml
```diff
@@ -11,15 +11,15 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:text="Short Text"
-    app:buttonIcon="@drawable/bpk_close"
+    app:buttonIcon="@drawable/bpk_close_sm"
     android:layout_marginBottom="@dimen/bpkSpacingBase" />
 
   <net.skyscanner.backpack.button.BpkButton
     android:id="@+id/button_decrease"
     android:layout_width="wrap_content"
```
bvitaliyg, line 19: This should be match_parent for testing purposes

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -39,7 +39,6 @@ import net.skyscanner.backpack.demo.stories.IconType
 import net.skyscanner.backpack.demo.stories.IconsStory
 import net.skyscanner.backpack.demo.stories.InteractiveStarRatingStory
 import net.skyscanner.backpack.demo.stories.LabeledCalendarStory
-import net.skyscanner.backpack.demo.stories.LoadingButtonStory
```
bvitaliyg, line 42: We need to have a story when loading property changes in runtime

bvitaliyg (changes requested):

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
marianeum: yep this works - we actually use the property within the button component as well (e.g. in the constructor)

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
marianeum: that's for good measure for RTL support - in this case it doesn't change anything as we've got the same padding on both sides, but it's good practise to use `relative`

app/src/main/res/layout/fragment_buttons_changeable.xml
marianeum: this hasn't changed from before - isn't this testing that the button size decreases as the text size decreases with `wrap_content`?

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
marianeum: TIL!

marianeum: good catch - this actually uncovered a bug - fixed that and also added a test for it

bvitaliyg (approved):
👍

PR #612, opened 2021-07-13
Koa 4601 dev android calendar remake tests

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCells.kt
```diff
@@ -87,3 +90,25 @@ internal fun CalendarCells(
 
   return CalendarCells(months = months)
 }
+
+private fun daysBeforeRangeStarts(params: CalendarParams): List<LocalDate> {
```
marianeum, line 94: this looks like new behaviour compared to the existing calendar - what's the aim for this?

app/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTest.kt
```diff
@@ -158,7 +158,7 @@ class BpkCalendarTest : BpkSnapshotTest() {
 
     Espresso
       .onView(withId(R.id.bpk_calendar_recycler_view))
-      .perform(RecyclerViewActions.actionOnItemAtPosition<RecyclerView.ViewHolder>(0, ViewActions.click()))
+      .perform(RecyclerViewActions.actionOnItemAtPosition<RecyclerView.ViewHolder>(18, ViewActions.click()))
```
marianeum, line 161: can we have some constants for these to make it easier to understand what's happening here?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarDispatchers.kt
```diff
@@ -0,0 +1,28 @@
+package net.skyscanner.backpack.calendar2.data
+
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import net.skyscanner.backpack.util.InternalBackpackApi
+
+@InternalBackpackApi
+@VisibleForTesting
+object CalendarDispatchers {
```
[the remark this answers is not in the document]
bvitaliyg: That's the approach Kotlin Coroutines use – e.g. usually you refer to Dispatchers.Main, but you can do Dispatchers.setMain() for testing purposes. I consider this as a reasonable solution if we have such holders related to each component (e.g. avoiding god-holder patters).

Injecting something into view may be more difficult, besides it's much easier to make a mistake when you can View constructor to call internal or testing constructors – IDE will show all of the constructors. You need to type CalendarDispatchers explicitly to find this api :)

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarAdapter.kt
```diff
@@ -37,11 +39,20 @@ internal class CalendarAdapter(
 
   private var data: CalendarCells = CalendarCells(emptyList())
 
+  val spanSizeLookup: GridLayoutManager.SpanSizeLookup = object : GridLayoutManager.SpanSizeLookup() {
```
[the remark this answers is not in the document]
bvitaliyg: Great idea

app/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTest.kt
bvitaliyg: These constants are specific for each tests case so I wouldn't extract it as constants. Besides it's clear to understand its meaning from the invocation place itself - e.g. cell position here.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCells.kt
bvitaliyg: The existing calendar fills the gap for the week (e.g. when your range starts with Wednesday) to avoid weird looking beginnings. This, however, is the behaviour used in iOS version of Calendar, so this code unifies it and cutting off this edgecase. It's also easier to implement and test then filling a week gap.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarDispatchers.kt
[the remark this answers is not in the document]
marianeum: What's the reason for not using the built in `Dispatchers.setMain` approach here, rather than creating essentially another implementation of it here? it looks like `Dispatchers.setMain` is also only available in the testing package, which would avoid somebody using it to update the application code to the main thread and we wouldn't have to have the `InternalBackpackApi` annotation for it

app/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTest.kt
marianeum: Yeah I understand it's positions, but it was unclear why a random 18 position - I assume this had to do with the range changes? Maybe we could have a `RANGE_START` constant and calculate other changes from that?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarDispatchers.kt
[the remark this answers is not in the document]
bvitaliyg: I would do that, but it only allows to inject the main dispatcher :(

app/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTest.kt
bvitaliyg: It may be easier to understand its meaning from the screenshot itself to see which dates are chosen exactly rather then look into the calculation code (which may also contain errors, in theory).

What do you think about extracting these constants into local variables with meaningful name, e.g. cellIndexOf19Jan?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarDispatchers.kt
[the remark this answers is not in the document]
marianeum: In that case I'm back to my original point that it feels like we're doing something we shouldn't do - this is just for the screenshot tests, right? I prefer to keep UI tests as close to the "real" implementation as possible, so forcing everything onto the main thread might be the wrong approach

app/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTest.kt
marianeum: yeah i like that!

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarDispatchers.kt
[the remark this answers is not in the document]
bvitaliyg: Injecting dispatchers is a common thing in coroutines. That's why there's a way to override main dispatcher provided (along with TestCoroutineDispatcher we use here).

https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/
https://medium.com/swlh/unit-testing-with-kotlin-coroutines-the-android-way-19289838d257

marianeum (dismissed):

marianeum (approved):

PR #623, opened 2021-08-02
KOA-4598 Fix fullscreen dialog test

marianeum:
The fullscreen dialog test was failing on CI as the scrollbar disappeared before the screenshot was taken. This is a workaround for this issue to increase the timeout when running on CI to ensure the scrollbar is still visible when taking the test.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


.github/workflows/ci.yml
```diff
@@ -139,6 +139,14 @@ jobs:
             ./gradlew :app:record${{ env.flavour }}${{ env.config }}AndroidTestScreenshotTest -Pandroid.testInstrumentationRunnerArguments.variant=${{ matrix.variant }}
             ./scripts/check-no-changes.sh
 
+      - name: Upload Screenshot Tests Artifact
```
marianeum, line 142: this will help us look at the screenshot test outputs in general

app/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
```diff
@@ -168,7 +169,41 @@ class BpkDialogTest : BpkSnapshotTest() {
     record(dialog, asyncScreenshot)
   }
 
+  @Test
+  fun screenshotTestFlareDialog() {
```
marianeum, line 173: this test just moved as it was below the test helper functions

app/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
```diff
@@ -168,7 +169,41 @@ class BpkDialogTest : BpkSnapshotTest() {
...
+        }
+      )
+    }
+
+    record(dialog, asyncScreenshot)
+  }
+
   private fun record(dialog: BpkDialog, asyncScreenshot: AsyncSnapshot) {
+    // not ideal, but the scrollbar disappears too early when running on CI causing test failures if visible
+    dialog.window?.decorView?.findScrollView()?.scrollBarDefaultDelayBeforeFade = 5000
```
marianeum, line 205: I have no idea why it disappears on CI as the screenshot is taken immediately when running locally. I don't like the workaround, but don't have a better idea

app/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
```diff
@@ -196,34 +231,19 @@ class BpkDialogTest : BpkSnapshotTest() {
...
-        R.drawable.bpk_tick,
-        context.getColor(R.color.bpkMonteverde),
-      )
-
-      image!!.setImageBitmap(bitmap)
-
-      addActionButton(
-        BpkButton(context).apply {
-          text = "Continue"
+  private fun View.findScrollView(): ScrollView? {
```
marianeum, line 234: didn't want to expose an id for the scrollview as clients shouldn't really access it, so added a function to find the scrollview instead

.github/workflows/ci.yml
georgegillams: This is a great idea 👍

app/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
georgegillams: Seems like a good approach to me. We do something similar in iOS where we access private selectors by name. Not appropriate for production code, but for tests it's 💯

app/src/androidTest/java/net/skyscanner/backpack/dialog/BpkDialogTest.kt
georgegillams: Just to make sure I understand - the test is flaky because the scroll bar is sometimes visible and sometimes not. This guarantees that the scrollbar will always be visible?

marianeum: On CI it seems to always be invisible with the default timeout, whereas locally it is visible. I can't see what's happening, but my suspicion is that on CI the test takes longer to run and the scrollbar disappears before the screenshot is taken. Increasing the timeout in tests ensures that it is always visible

georgegillams: got it 👍

georgegillams (approved):

PR #625, opened 2021-08-04
KOA-4677 Setup lint config

marianeum:
This sets up a more strict lint config to spot errors sooner and keep our code quality high. I've setup warnings to be treated as errors and ignored a few issues that we're not interested in. 
All current issues are added to a baseline and will be fixed separately.

I originally took the configuration from the app which was already treating many things as an error that are usually just warnings, but then decided to make it simpler and treat warnings as errors in general, while ignoring the few issues we actually don't care about. It is not possible to share the lint config with a separate project or have multiple lint config files, so even if we were to create a script to copy the file into the project the app couldn't then define some of their exclusions for specific files that are currently added, so this wouldn't be a workable solution either.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/build.gradle
+++ b/Backpack/build.gradle
@@ -34,7 +34,7 @@ android {
   lintOptions {
     baseline file("lint-baseline.xml")
     lintConfig file("../lint.xml")
-    checkReleaseBuilds false
+    checkReleaseBuilds false // we're already running lint separately
     warningsAsErrors true
   }
 }
--- a/Backpack/build.gradle
+++ b/Backpack/build.gradle
@@ -31,6 +31,12 @@ android {
       "-Xuse-experimental=net.skyscanner.backpack.util.ExperimentalBackpackApi",
     ]
   }
+  lintOptions {
+    baseline file("lint-baseline.xml")
+    lintConfig file("../lint.xml")
+    checkReleaseBuilds false
+    warningsAsErrors true
+  }
 }
 
 dependencies {
--- a/Backpack/lint-baseline.xml
+++ b/Backpack/lint-baseline.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="5" by="lint 4.2.2" client="gradle" variant="debug" version="4.2.2">
+
+    <issue
+        id="CustomViewStyleable"
+        message="By convention, the custom view (`MonthView`) and the declare-styleable (`BpkCalendar`) should have the same name (various editor features rely on this convention)"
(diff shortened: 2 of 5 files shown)


app/build.gradle
```diff
@@ -53,14 +53,10 @@ android {
     }
   }
   lintOptions {
-    // Backups not required for demo app
-    disable 'AllowBackup',
```
marianeum, line 57: these are now defined in the `lint.xml`

app/build.gradle
```diff
@@ -53,14 +53,10 @@ android {
...
-      // Gradle dependencies to be in sync with app
-      'GradleDependency',
-      // App indexing not required for demo app
-      'GoogleAppIndexingWarning',
-      // A11y support TBD
-      'ContentDescription'
```
marianeum, line 63: I removed this ignore to make accessibility a requirement

Backpack/build.gradle
```diff
@@ -31,6 +31,12 @@ android {
       "-Xuse-experimental=net.skyscanner.backpack.util.ExperimentalBackpackApi",
     ]
   }
+  lintOptions {
+    baseline file("lint-baseline.xml")
+    lintConfig file("../lint.xml")
+    checkReleaseBuilds false
```
marianeum, line 37: this would run lint as part of a release build, but as we're already running it separately this isn't needed (it'd just unnecessarily increase build times)

Backpack/lint-baseline.xml
```diff
@@ -0,0 +1,92 @@
...
+        errorLine1="    val a = this.context.obtainStyledAttributes(attrs, R.styleable.BpkCalendar, R.attr.bpkCalendarStyle, 0)"
+        errorLine2="                                                       ~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/main/java/net/skyscanner/backpack/calendar/view/MonthView.kt"
+            line="117"
+            column="56"/>
+    </issue>
+
+    <issue
+        id="CustomViewStyleable"
```
olliecurtis, line 16: Is this the same issue as above?

Backpack/lint-baseline.xml
```diff
@@ -0,0 +1,92 @@
...
```
olliecurtis, line 27: Is this the same issue as above?

olliecurtis (changes requested):

Backpack/lint-baseline.xml
marianeum: I can't see any difference between the three either, but if I remove the duplicates the lint errors come back 🤷‍♀️

Backpack/build.gradle
georgegillams: Is it possible to put a comment to that effect in the file?

PR #641, opened 2021-08-06
KOA-4690 Update to AS arctic fox

marianeum:
Update to the latest stable Android Studio version. This requires to update the gradle wrapper and update the lint version which introduced a few new checks that needed to be dealt with.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


.github/workflows/ci.yml
```diff
@@ -68,6 +68,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Set up JDK 11
```
marianeum, line 71: The new AS needs java 11

.idea/runConfigurations.xml
```diff
@@ -1,13 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
marianeum, line 1: this got automatically deleted when updating. I don't think we had anything custom in here anyways and it was just the default run configurations that get generated

Backpack/build.gradle
```diff
@@ -17,7 +17,6 @@ android {
   defaultConfig {
     minSdkVersion rootProject.ext.minSdkVersion
     targetSdkVersion rootProject.ext.targetSdkVersion
-    versionName "1.0"
```
marianeum, line 20: this isn't needed anymore too

Backpack/build.gradle
```diff
@@ -32,9 +31,11 @@ android {
     ]
   }
   lintOptions {
+    baseline file("lint-baseline.xml")
     lintConfig file("../lint.xml")
     checkReleaseBuilds false // we're already running lint separately
     warningsAsErrors true
+    disable 'UnusedResources' // we're exposing resources for consumers
```
marianeum, line 38: not sure why this didn't trigger before, but we don't care about unused resources here as we're exposing things for the app that we don't use in the library

Backpack/lint-baseline.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 7.0.0" type="baseline" client="gradle" name="AGP (7.0.0)" variant="all" version="7.0.0">
+
+    <issue
```
marianeum, line 4: i created a ticket to improve this as it's not a minor tweak

app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,7 @@
       android:name="com.google.android.geo.API_KEY"
       android:value="${MAPS_API_KEY}" />
 
-      <activity android:name=".MainActivity">
+      <activity android:name=".MainActivity" android:exported="true">
```
marianeum, line 18: we now need to explicitly specify this activities with an intent filter as exported. the main launcher activity should always be exported, so other apps (and android studio!) can launch it

olliecurtis (approved):

.github/workflows/ci.yml
georgegillams: Can this be put under `defaults` to make it run for each stage instead of duplicating it?

marianeum: TIL! perfect :)

marianeum: actually it doesn't look like you can add a step in `defaults` :(

georgegillams: Ah yeah I was juts looking at [the docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions?query=defaults#defaults) 😢 
I guess we'd have to create a separate action, which is probably overkill

PR #675, opened 2021-08-30
Fix for calendar2 not formatting months with less days than the current day

zgeor:
Calendar2 was failing to correctly create the label for any month which had less days than the current day. 
For example, if it tried formatting February on 30th of any month then the label would come off as March. This was due to using a `Date()` without resetting the day(date) value, thus the formatter incorrectly interpreting this as the next month. 

The fix is to reset the `Date()` object to a well know day. In order to avoid TZ issues I've set it up as 2.

Adding new automated test cases will require a discussion around the design of the component.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarMonth.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarMonth.kt
@@ -81,8 +81,10 @@ internal inline fun CalendarMonth(
 // see https://github.com/ThreeTen/threetenbp/issues/55
 @Suppress("DEPRECATION")
 internal fun MonthTitle(yearMonth: YearMonth, formatter: SimpleDateFormat, locale: Locale): String {
-  val date = Date()
-  date.year = yearMonth.year - 1900
-  date.month = yearMonth.monthValue - 1
+  val date = Date(
+    yearMonth.year - 1900,
+    yearMonth.monthValue - 1,
+    2 // Lock the formatter date to 2nd, so it accounts for -12h tz and the month is predictable
+  )
   return formatter.format(date).capitalize(locale)
 }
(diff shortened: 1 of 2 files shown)


Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarMonth.kt
```diff
@@ -81,8 +81,10 @@ internal inline fun CalendarMonth(
 // see https://github.com/ThreeTen/threetenbp/issues/55
 @Suppress("DEPRECATION")
 internal fun MonthTitle(yearMonth: YearMonth, formatter: SimpleDateFormat, locale: Locale): String {
-  val date = Date()
-  date.year = yearMonth.year - 1900
-  date.month = yearMonth.monthValue - 1
+  val date = Date(
+    yearMonth.year - 1900,
+    yearMonth.monthValue - 1,
+    2 // Lock the formatter date to 2nd, so it accounts for -12h tz and the month is predictable
```
marianeum, line 87: I'm not sure if that's the intended behaviour as this will modify the day in production code to a fixed one

zgeor: MonthTitle()'s purpose is to transform a year and a month to a title for the month, the day is irrelevant. This particular Date() value is only used so the month formatter could work. However, it wasn't working correctly because when creating a new Date(), it would take the current date as default, e.g. 2021-08-30, then we were changing the year and month without updating the day. This led to having a date of 2021-02-30 which is technically invalid and the formatter interpreted this as March rather than February.

So given the scoped use of this Date() object it is ok to use a static day.

marianeum: aaaah got it

marianeum (approved):

PR #689, opened 2021-09-09
Improved dependabot compatibility.

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/build.gradle
+++ b/Backpack/build.gradle
@@ -60,12 +60,13 @@ android {
 
 dependencies {
   api rootProject.ext.androidxAnnotations
+  api "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   api rootProject.ext.coroutines
   api rootProject.ext.androidxAppCompat
   api rootProject.ext.androidMaterial
   api rootProject.ext.constraintLayout
   api rootProject.ext.cardView
-  api rootProject.ext.threeTenAbp
+  api rootProject.ext.threeTenAbpAndroid
   api rootProject.ext.googleMaps
   implementation rootProject.ext.swiperefreshLayout
   implementation rootProject.ext.ktx
--- a/Backpack/build.gradle
+++ b/Backpack/build.gradle
@@ -1,3 +1,21 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -63,24 +78,27 @@ ext {
...
+  stetho = "com.facebook.stetho:stetho:1.6.0"
+  junitAndroid = "androidx.test.ext:junit:1.1.2"
+  espressoCore = "androidx.test.espresso:espresso-core:3.3.0"
+  espressoContrib = "com.android.support.test.espresso:espresso-contrib:3.1.0"
+  testRules = "androidx.test:rules:1.3.0"
+  mockitoKotlin = "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
+  mockitoAndroid = "org.mockito:mockito-android:3.9.0"
+  coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.0"
+  coroutinesTest = "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.5.0"
+  lint = "com.android.tools.lint:lint:30.0.2"
```
marianeum, line 101: I think dependabot recognises having a separate version + dependency definition as long as it's still in the same file (e.g. the kotlin updates worked) - maybe we can still provide version + dependency separately where we've got multiple dependencies all using the same version - just defined in the same file. otherwise i'm not sure how upgrades work - it might try and update all three dependencies on their own and therefore it'll probably fail (same for coroutines)

marianeum: yeah e.g. this worked https://github.com/Skyscanner/backpack-android/pull/668

Backpack/build.gradle
```diff
@@ -60,12 +60,13 @@ android {
 
 dependencies {
   api rootProject.ext.androidxAnnotations
+  api "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
marianeum, line 63: to match everything else can we define this in the `ext` part as well?

bvitaliyg: Nope – since it's used in classpath dep[endencies of top-level buildscript it fails.

marianeum (approved):
dependabot tomorrow:

PR #704, opened 2021-09-15
Added the compose module.

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


build.gradle
```diff
@@ -80,7 +81,7 @@ ext {
   targetSdkVersion = 30
   minSdkVersion = 24
 
-  androidxAppCompat = "androidx.appcompat:appcompat:1.2.0"
+  androidxAppCompat = "androidx.appcompat:appcompat:1.3.1"
```
bvitaliyg, line 84: This is needed for Compose to work properly.

Backpack/build.gradle
```diff
@@ -80,6 +80,7 @@ dependencies {
   testImplementation rootProject.ext.threeTenAbp
 
   lintPublish project(':backpack-lint')
+  implementation project(':Backpack-Common')
```
marianeum, line 83: a nit on the module names - module names are generally lowercase - `Backpack` is the odd one out (maybe we can fix that too)

Backpack-Compose/build.gradle
```diff
@@ -0,0 +1,91 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+
+version = rootProject.ext.version
+group = 'net.skyscanner.backpack-compose'
```
marianeum, line 23: how about we keep the group as `backpack`, but use a different artifact id? - e.g. the old one is `net.skyscanner.backpack:backpack-android` and we publish another artifact `net.skyscanner.backpack:backpack-compose`

Backpack-Compose/build.gradle
```diff
@@ -0,0 +1,91 @@
...
+
+repositories {
+  google()
+  mavenCentral()
+}
+android {
```
marianeum, line 29: there seems to be quite a lot of duplication - can the generic bits be moved into a separate gradle file that gets applied?

app/build.gradle
```diff
@@ -91,11 +91,19 @@ android {
     targetCompatibility JavaVersion.VERSION_1_8
   }
   kotlinOptions {
+    useIR = true
```
marianeum, line 94: Is this needed? the [docs](https://kotlinlang.org/docs/whatsnew14.html#new-jvm-ir-backend) mention that it gets automatically enabled for compose

Backpack-Common/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'kotlin'
+
+dependencies {
```
marianeum, line 21: the extra test tasks are missing here - can we also have a kotlin only version of a generic setup gradle file for this + lint?

build.gradle
```diff
@@ -17,7 +17,8 @@
  */
 
 buildscript {
-  ext.kotlin_version = '1.5.30'
+  ext.kotlin_version = '1.5.21'
```
bvitaliyg, line 20: CI won't pass without this. We probably need to delay it until Compose stable for 1.5.30 is released, or switch to 1.1.0-alpha03.

build.gradle
marianeum: It might be worth doing the upgrade in the dependabot PR as it requires a bunch of lint fixes

Backpack/build.gradle
```diff
@@ -102,10 +78,5 @@ task generateDokkaDocs(type: DokkaTask) {
...
 }
 
-task testOssDebugUnitTest {
-  dependsOn("testDebugUnitTest")
-}
-
-task lintOssDebug {
-  dependsOn("lintDebug")
-}
+apply from: "$rootProject.projectDir/android-test-configuration.gradle"
```
marianeum, line 81: how come these are separate files, rather than being part of `android-configuration`?

bvitaliyg: Testing task with the same name is defined in app module by flavours – we get gradle error if we redefine it manually.

PR #710, opened 2021-09-21
KOA-3595 Improve snackbar accessibility

marianeum:
Improved accessibility for Snackbar by:
- Increasing the duration Snackbar is visible if screenreader is enabled to allow for navigation to the component
- Requiring a `contentDescription` for icon actions

Also fixed a warning by removing the `setAction` and `setOnDismissed` extension functions in favour of member functions

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/snackbar/BpkSnackbar.kt
```diff
@@ -161,6 +161,15 @@ class BpkSnackbar private constructor(
     callback?.let(callbacks::add)
   }
 
+  fun setOnDismissed(ignoreDismissAfterAction: Boolean = true, callback: () -> Unit) =
```
marianeum, line 164: moved from `Extensions`. Using the extension for `setActions` caused a warning:
> Candidate resolution will be changed soon, please use fully qualified name to invoke the following closer candidate explicitly

Backpack/src/main/java/net/skyscanner/backpack/util/Context.kt
```diff
@@ -31,3 +33,13 @@ private fun Context.unwrapped(): Context {
   }
   return context
 }
+
+fun Context.isScreenReaderOn(): Boolean {
```
bvitaliyg, line 37: Can we make this internal?

Backpack/src/main/java/net/skyscanner/backpack/snackbar/BpkSnackbar.kt
```diff
@@ -253,10 +266,16 @@ class BpkSnackbar private constructor(
         backgroundColor = it.getColor(R.styleable.BpkSnackbar_snackbarBackgroundColor, backgroundColor)
       }
 
+      val adjustedDuration = if (duration != BpkSnackbar.LENGTH_INDEFINITE && context.isScreenReaderOn()) {
```
bvitaliyg, line 269: BpkSnackbar is not necessary here, LENGTH_INDEFINITE works just fine as well.

Backpack/src/main/java/net/skyscanner/backpack/snackbar/BpkSnackbar.kt
bvitaliyg: A warning in the app or in Backpack?

marianeum: in the app and in the backpack sample app - whenever it is used. i think it's because both functions now provide the same signature - i think in the past you had to explicitly specify that it was a click listener (hence the extension function for simplification probably), but this is no longer needed

bvitaliyg (approved):
👍

PR #714, opened 2021-09-28
KOA-4749 Add compose tab

marianeum:
Added a new `TabStory` which allows having tabs for entries. Also added the option to add a compose story without much boilerplate.
device-2021-09-28-115521

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
@@ -18,8 +18,13 @@
 
 package net.skyscanner.backpack.demo.data
 
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
 import net.skyscanner.backpack.demo.R
 import net.skyscanner.backpack.demo.stories.BarChartStory
 import net.skyscanner.backpack.demo.stories.BottomNavStory
@@ -127,6 +132,9 @@ private infix fun String.story(story: NodeData): Pair<String, NodeItem> {
  * Helper class to register the fragments for components
  */
 object ComponentRegistry {
+
+  private val TAB_TITLE_COMPOSE = "Compose"
+  private val TAB_TITLE_VIEW = "View"
 
   private val COMPONENTS_TREE = mapOf(
     "Badge" story NodeData { Story of R.layout.fragment_badge },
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -86,14 +90,23 @@ class NodeItem(
     var fullName = this.name
 
     while (parent != null) {
-      fullName = "${parent.name} - $name"
+      fullName = "${parent.name} - $fullName"
```
marianeum, line 93: this was a bug if there were more than 2 node levels

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -86,14 +90,23 @@ class NodeItem(
...
       parent = parent.getParent()
     }
 
     return fullName
   }
 }
 
+class ComposeNode(
```
marianeum, line 101: To make it easier to add a compose screen - with compose we'll have to have separate composables for each story anyways - rather than having to provide a separate story class everywhere we can just define the composable in the registry (similar to how we set the layout id)

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -284,6 +301,13 @@ object ComponentRegistry {
     "Text Field" story NodeData { Story of R.layout.fragment_text_fields },
     "Text Spans" story NodeData { TextSpansStory of R.layout.fragment_text_spans },
     "Toast" story NodeData { ToastStory of R.layout.fragment_toasts },
+    "Sneak peek" story NodeData(
+      { children -> TabStory of children },
+      mapOf(
+        "Compose" composeStory { Text(text = ("Hello!")) },
```
marianeum, line 307: this is how the `composeStory` is being used

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -324,7 +348,7 @@ object ComponentRegistry {
     story ?: throw IllegalArgumentException("Invalid story name - $fullyQualifiedName")
 
     return rest.fold(story) { result, item ->
-      return result.subItems[item]
+      result.subItems[item] as? NodeItem
```
marianeum, line 351: another bug for more than 2 levels

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
bvitaliyg: This is exactly how Compose is designed to be used :)

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
bvitaliyg: What was the bug?

marianeum: It would skip some levels when assembling the full name - e.g. if the level is Button - Compose - Primary it would end up with Button - Primary as it would just use the name, rather than the partial full name (as in Primary instead of Compose - Primary)

bvitaliyg: Do we need to update the website screenshots taking code to reflect that change?

marianeum: No - the `DocsRegistry` takes a list of ids (which is the full names), so we define the full name and then starts the details fragment. The correct item is then resolved based on that id using the same code as here. 
We currently don't have any items that have more than 2 levels, so we never ran into this issue. However with tabs this is much more likely - e.g. say we add a button component for compose we'll have button -> view -> all the different options (and probably also for the compose version)

bvitaliyg (approved):

PR #719, opened 2021-10-04
KOA-4775 Add compose indicator

marianeum:
Add an indicator on the component list for components available in compose. Also fixed a bug with dividers disappearing that bothered me for ages.
device-2021-10-04-171254

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
@@ -60,10 +60,6 @@ class StoriesRecyclerViewAdapter internal constructor(
 
     if (values[position] is StoryItem) {
       holder.itemView.setOnClickListener(onClickListener)
-
-      if (position + 1 == itemCount || values[position + 1] is HeaderItem) {
-        holder.view.findViewById<View>(R.id.component_divider).visibility = View.GONE
-      }
     }
   }
 
@@ -72,7 +68,7 @@ class StoriesRecyclerViewAdapter internal constructor(
   }
 
   override fun getItemViewType(position: Int): Int {
-    return if (values[position] is HeaderItem) 0 else 1
+    return if (values[position] is HeaderItem) HEADER_VIEW_TYPE else STORY_VIEW_TYPE
   }
 
   inner class ViewHolder(val view: View) : RecyclerView.ViewHolder(view) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
```diff
@@ -60,10 +60,6 @@ class StoriesRecyclerViewAdapter internal constructor(
 
     if (values[position] is StoryItem) {
       holder.itemView.setOnClickListener(onClickListener)
-
-      if (position + 1 == itemCount || values[position + 1] is HeaderItem) {
```
marianeum, line 64: this bothered me for so long - the divider always has the wrong visibility because a) we're just hiding it here and never showing it again and b) items might not always be rebound (although that's less of an issue for a static list like this)

app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
```diff
@@ -75,16 +78,9 @@ class StoriesRecyclerViewAdapter internal constructor(
     val name: TextView = view.findViewById(R.id.component_name)
   }
 
-  interface ListItem {
```
marianeum, line 82: just cleaned this up into a sealed class while i'm at it

app/src/main/res/layout/component_list_content.xml
```diff
@@ -3,17 +3,27 @@
   xmlns:app="http://schemas.android.com/apk/res-auto"
   android:layout_width="match_parent"
   android:layout_height="wrap_content"
-  android:orientation="vertical">
+  xmlns:tools="http://schemas.android.com/tools"
+  android:paddingStart="@dimen/bpkSpacingLg"
+  android:paddingEnd="@dimen/bpkSpacingLg"
+  android:background="?attr/selectableItemBackground"
+  android:orientation="horizontal">
```
marianeum, line 10: this is why i ended up making the divider change here too - this is just a linearlayout right now

app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
```diff
@@ -72,21 +71,19 @@ class StoriesRecyclerViewAdapter internal constructor(
   }
 
   override fun getItemViewType(position: Int): Int {
-    return if (values[position] is HeaderItem) 0 else 1
+    return if (values[position] is HeaderItem) HEADER_VIEW_TYPE else STORY_VIEW_TYPE
```
bvitaliyg, line 74: 👍

bvitaliyg (approved):

PR #725, opened 2021-10-07
KOA-4757 Add publishing for common

marianeum:
This adds support for publishing in common. Right now whenever releasing a new version common will also increase. The next step is to only increase common when changes were made in which case we can also rewrite this in Kotlin/python/something else to avoid adding more dependencies.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/build.gradle
+++ b/Backpack/build.gradle
@@ -55,7 +55,7 @@ dependencies {
   testImplementation rootProject.ext.coroutinesTest
   testImplementation rootProject.ext.threeTenAbp
 
-  implementation project(':backpack-common')
+  api project(':backpack-common')
   lintPublish project(':backpack-lint')
 }
 


Backpack/dokka.gradle
```diff
@@ -27,11 +27,11 @@ tasks.register("generateDokkaStatics") {
 def create404Page() {
   createNewFile(rootProject.rootDir, "$rootProject.ext.dokkaOutputDir/404.html").text = """<!DOCTYPE html>
 <html>
-  <meta http-equiv="refresh" content="1;url=/android/versions/$rootProject.ext.version" />
+  <meta http-equiv="refresh" content="1;url=/android/versions/$rootProject.ext.backpackAndroidVersion" />
```
marianeum, line 30: eventually this will need to be made generic as well

release.js
```diff
@@ -23,14 +23,15 @@ const semver = require('semver');
 const replace = require('replace-in-file');
 const releaseit = require('release-it');
 
-const pkg = require('./package.json');
+const versions = require('./versions.json');
```
marianeum, line 26: decided to create a new file to keep the versions, rather than relying on the package.json. this will be helpful when we move away from npm and allows us to keep multiple versions

build.gradle
```diff
@@ -77,10 +77,12 @@ task clean(type: Delete) {
   delete rootProject.buildDir
 }
 
-def BpkVersion = '33.0.1'
+def BpkAndroidVersion = '33.0.1'
```
zgeor, line 80: Any reason to keep these here, rather than delegate versioning to the modules themselves?

zgeor (approved):

PR #727, opened 2021-10-08
Koa 4754 dev android setup compose foundations for token generation

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkBorderRadius.kt
```diff
@@ -0,0 +1,36 @@
...
+
+package net.skyscanner.backpack.compose.tokens
+
+import androidx.compose.runtime.Stable
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+public object BpkBorderRadius {
+  @Stable
+  public val Lg: Dp = 24.dp
```
bvitaliyg, line 23: No way not to use val type with KotlinPoet.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkBorderRadius.kt
```diff
@@ -0,0 +1,36 @@
+// Backpack for Android - Skyscanner's Design System
```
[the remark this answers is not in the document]
zgeor: Would this be a problem when we generate the dokka docs?

backpack-compose/tokens.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import net.skyscanner.backpack.tokens.*
+
+tasks {
+
+  val tokensPackage = "net.skyscanner.backpack.compose.tokens"
+  val group = "tokens"
+  val src = project.projectDir.resolve("src/main/kotlin").path
```
marianeum, line 24: Is it a conscious decision to move away from /java here like we have in the rest of the project?

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
+repositories {
```
marianeum, line 1: have you looked into composite builds instead of buildsrc?

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/Case.kt
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.skyscanner.backpack.tokens
+
+import com.google.common.base.CaseFormat
+
+fun String.changeCase(from: CaseFormat, to: CaseFormat): String =
```
zgeor, line 22: This seems to be a simple wrapper. Is it necessary? Do you think it will grow into something else ?

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkDimensions.kt
```diff
@@ -0,0 +1,113 @@
...
+): BpkDimensions {
+
+  val props = source.getValue("props") as Map<String, Map<String, String>>
+  val data = props.filter { (_, value) -> value["type"] == "size" && value["category"] == category }
+
+  val map = data
+    .mapValues {
+      val value = it.value.getValue("value")
+      if (evaluate) {
+        JsEvaluator.eval(value).toString()
```
marianeum, line 78: what is this doing?

backpack-compose/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
+/**
```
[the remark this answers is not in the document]
bvitaliyg: I think eventually we will, but for now it's fine to have a part of the project being in kts.

backpack-compose/tokens.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+    doLast {
+      source
+        .parseAs(BpkDimension.Category.Radii)
+        .transformTo(BpkDimension.Format.Compose(namespace = "BpkBorderRadius"))
+        .saveTo(BpkOutput.KotlinFile(src, tokensPackage))
+        .execute()
+    }
+  }
+
+  val sizeTokens by creating {
```
marianeum, line 62: should we generate a `BpkDimension` file as well to contain all the different types of dimension tokens for easier discoverabilit?

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/Pipeline.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.skyscanner.backpack.tokens
+
+interface Pipeline<T> {
```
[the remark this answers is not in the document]
bvitaliyg: They effectively do the same -- mapping one type to another using a function. The reason why it's have different aliases to have more clear, type-safe, DSL-like semantics.

I think we can put it in one file. Do you think we need to? :)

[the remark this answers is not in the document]
marianeum: +1 I find it very difficult to understand what's happening. It looks like there's a lot of abstraction where I'm not sure what it actually does or what the purpose is

backpack-compose/build.gradle.kts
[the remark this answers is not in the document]
marianeum: what's the reason for converting this now?

buildSrc/build.gradle.kts
bvitaliyg: Yes, and I think it'll be easier for now to go with buildSrc.

bvitaliyg:
I was planning to do this once all of our token generation tasks will be completed – we need to check not only the sizes but other stuff as well.

backpack-compose/tokens.gradle.kts
bvitaliyg: Just having it in one package should be discoverable enough. If we do nested object, notations like BpkDimenson.BpkElevation.Sm may be harder to read and write. I think people will get used to the dimensions quite quickly, considering we also have autosuggest in IDE.

marianeum: I'm not sure the package + auto complete is discoverable enough - there'll be many things starting with `Bpk` but you won't easlily know what's a dimension, rather than being able to use `BpkDimension.x` - you can also add imports for `BpkElevation` directly for the common ones or the ones you know

bvitaliyg: It's pretty much similar to tokens naming in XMLs right now (e.g. bpkElevationXs). Should be very intuitive to find, providing that we'll also have examples on the website. If not, we can always change it later :)

bvitaliyg: When I type @dimen/something, my focus is on "something". I already know type I'm looking for.

What we can do, however, is manually create a bridge (e.g. keeping original classes intact). Something like:

```
object BpkDimension {

   val Elevation = BpkElevation

   val Spacing = BpkSpacing

}
```

But to me honest, I've never seen such notation is used in Compose. People just put constants in objects, like here:
https://github.com/androidx/androidx/blob/androidx-main/compose/material/material/src/commonMain/kotlin/androidx/compose/material/Button.kt#L290

marianeum: Yep that's what I originally proposed :) 

these dimensions are more our global dimensions than tied to a specific component like a button where it makes sense to keep them scoped to the button (and keep in the same file as Button as well)

backpack-compose/tokens.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+    doLast {
+      source
+        .parseAs(BpkDimension.Category.Elevation)
+        .transformTo(BpkDimension.Format.Compose(namespace = "BpkElevation"))
+        .saveTo(BpkOutput.KotlinFile(src, tokensPackage))
+        .execute()
+    }
+  }
+
+  val spacingTokens by creating {
```
marianeum, line 47: where is the `creating` coming from? i can't find it anywhere. i find it a bit difficult to understand what's happening here

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkDimensions.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.MemberName
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeSpec
+
+interface BpkDimensions : Map<String, Int>
+
+object BpkDimension {
+
+  sealed class Category : BpkParser<BpkDimensions> {
```
marianeum, line 31: It looks like there's multiple size token categories that we're not parsing (e.g. typesettings and carousel-indicator-dots) - should we parse these and add some check to fail the task if there are any categories we're not parsing? Even if we don't have components for carousel-indicator-dots the app might need them and we don't miss out on new tokens. typesettings we'll need anyways.

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkDimensions.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    object Spacing : Category() {
+      override fun invoke(source: Map<String, Any>): BpkDimensions =
+        parseDimensions(source, "spacings", "SPACING_") {
+          it.key.toLowerCase() in setOf("base", "lg", "md", "none", "sm", "xl", "xxl")
```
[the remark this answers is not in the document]
marianeum: I wonder why that is and if that's a good thing (similar to the comment above) - surely the tokens should be the source of truth, rather than us having to filter what we know about?

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkDimensions.kt
marianeum: There's also lineheight, but that should probably also be treated as sp instead of dp. Maybe we need to update the types at some point to be more explicit.

I'm not necessarily talking about these specific ones right now - just if we actually want to necessarily restrict ourselves to the known types or rather take what comes from foundations as the source of truth. but probably for that we first need to clean up what comes from foundations and revisit that later

PR #754, opened 2021-11-03
KOA-4787 Add text input layout

marianeum:
Added a new `BpkTextInputLayout` component similar to the material `TextInputLayout` in api (but not in behaviour) to add a label and/or error message or helper text wrapper to an input field

| Default | Dark Mode |
| --- | --- |
| labels | labels_dm |

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextField.kt
```diff
@@ -102,9 +108,6 @@ open class BpkTextField @JvmOverloads constructor(
     setTextColor(
       colorStateList(
         color = textColor,
-        pressedColor = textColor,
```
marianeum, line 105: I've added a default param in the `colorStateList` for these so these aren't needed anymore

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextField.kt
```diff
@@ -68,6 +68,12 @@ open class BpkTextField @JvmOverloads constructor(
       setCompoundDrawablesRelativeWithIntrinsicBounds(iconStart, null, iconEnd, null)
     }
 
+  var hasError: Boolean = false
```
marianeum, line 71: this is to add an error border in case of an error mode

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+
+  var errorIcon: Drawable? = null
+    set(value) {
+      field = value?.mutate()
+      field?.setBounds(0, 0, iconSize, iconSize)
+      updateErrorIconTint()
+      updateIndicator()
+    }
+
+  var label: String? = null
```
marianeum, line 65: all of these follow the naming of `TextInputLayout` to make the migration easier

app/src/main/res/layout/fragment_text_fields.xml
```diff
@@ -11,68 +11,35 @@
...
     app:textFieldIconEnd="@drawable/bpk_search" />
 
   <net.skyscanner.backpack.text.BpkTextField
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:layout_marginBottom="@dimen/bpkSpacingBase"
-    android:hint="Input disabled"
-    android:enabled="false"/>
-
-  <net.skyscanner.backpack.text.BpkTextField
```
marianeum, line 45: I split things up a little as it was getting a bit big

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComponentUsageDetector.kt
```diff
@@ -223,6 +223,16 @@ class BpkComponentUsageDetector : Detector(), SourceCodeScanner, XmlScanner {
       webName = "text",
       componentsToReplace = setOf("android.widget.TextView", "TextView", "androidx.appcompat.widget.AppCompatTextView")
     ),
+    TEXT_FIELD(
+      fullName = "net.skyscanner.backpack.text.BpkTextField",
```
bvitaliyg, line 227: I think we need to add this to a new component checklist, it's easy to forget to update the rules.

marianeum: There is a danger check for this :)

docs/TextField/README.md
```diff
@@ -28,28 +28,83 @@ BpkTextField(context).apply {
...
+
+  addView(textField)
+}
+```
+
+**Note:** If both error message and helper text are supplied the error message takes precedence.
 
 ## Theme Props
 
+### `BpkTextField`
```
bvitaliyg, line 76: I would add a link to new component to root README.md with link to this anchor

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextField.kt
bvitaliyg: This is potentially dangerous situation as we now has 2 source of truth for error (border display flag):
* drawable state
* boolean flag

I wonder if we can use the drawable state as backing field here, e.g. `get() = android.R.attr.state_error in drawableState`

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+import android.widget.FrameLayout
+import android.widget.LinearLayout
+import androidx.appcompat.content.res.AppCompatResources
+import androidx.core.view.isVisible
+import net.skyscanner.backpack.R
+import net.skyscanner.backpack.util.colorStateList
+import net.skyscanner.backpack.util.unsafeLazy
+import net.skyscanner.backpack.util.use
+
+class BpkTextInputLayout @JvmOverloads constructor(
```
bvitaliyg, line 36: We usually make to components open. Do we have any reason not to do it here?

docs/TextField/README.md
marianeum: I was thinking about that but that'll be difficult for the site to refer to both readmes - the web component is already supporting labels etc, so we can't create a separate component on the site for it

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+  var error: String? = null
+    set(value) {
+      field = value
+      editText?.hasError = value != null
+      updateIndicator()
+    }
+
+  var helperText: String? = null
```
bvitaliyg, line 79: `hint`, perhabs? `helperText` looks confusing for me.

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+  init {
+    orientation = VERTICAL
+    inflate(context, R.layout.view_bpk_text_input_layout, this)
+    context.obtainStyledAttributes(
+      attrs,
+      R.styleable.BpkTextInputLayout,
+      defStyleAttr, 0
+    ).use {
+      errorTextColor = it.getColorStateList(R.styleable.BpkTextInputLayout_textInputErrorTextColor)
+        ?: ColorStateList.valueOf(context.getColor(R.color.bpkPanjin))
```
bvitaliyg, line 94: We can get ColorStateList from resources itself, no need to use `valueOf`. Just use color id instead of color state list id.

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+          disabledColor = context.getColor(R.color.__textFieldTextDisabled)
+        )
+      )
+      updateErrorIconTint()
+    }
+  }
+
+  override fun childDrawableStateChanged(child: View?) {
+    super.childDrawableStateChanged(child)
+    if (child is FrameLayout) {
```
bvitaliyg, line 114: User can add any amount of FrameLayouts to the component. We can replace it with id check.

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  override fun setEnabled(enabled: Boolean) {
+    super.setEnabled(enabled)
+    editText?.isEnabled = enabled
+  }
+
+  override fun addView(child: View?, index: Int, params: ViewGroup.LayoutParams?) {
+    if (child is BpkTextField) {
```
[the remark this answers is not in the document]
bvitaliyg: Also, the overloads configurations of these methods may have been changed in the future. I believe `onViewAdded()` would be more reliable choice here.

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
marianeum: `hint` could be confused with the edit text hint - i replicated the naming from `TextInputLayout`

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextInputLayout.kt
marianeum: Nope good shout!

Backpack/src/main/java/net/skyscanner/backpack/text/BpkTextField.kt
marianeum: Unfortunately this doesn't work as to add it to the drawable state we need to refresh the drawable state which then needs to know if we need to add it to the drawable state, so we end up in a stackoverflow error - we can't add it directly

bvitaliyg: What if instead of using field there we'll check it in array itself?

marianeum: It won't be in the array without it being added in the `onCreateDrawableState` method

bvitaliyg (approved):

PR #760, opened 2021-11-09
KOA-4801 Add compose typography styles

marianeum:
Added text styles for compose:
device-2021-11-05-132928

Some of the new tokens don't exist yet and will be replaced when they are added. Some text sizes need to be updated, however I used the naming for the updated styles, so the tokens can be updated under the hood. Haven't added line height yet, as these tokens might change naming before we adopt them.

I also created a basic theme for compose as it's needed for the different internal/oss variants.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Typography
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.text.font.FontWeight
+import net.skyscanner.backpack.compose.tokens.BpkTypography
+
+val LocalBpkTypography = staticCompositionLocalOf {
```
marianeum, line 29: this is following the material typography & theme setup

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkTypography.kt
```diff
@@ -0,0 +1,125 @@
...
+  val footnote: TextStyle,
+  val caption: TextStyle,
+  val label16: TextStyle,
+  val label14: TextStyle
+) {
+  constructor(
+    defaultFontFamily: FontFamily = FontFamily.Default,
+    hero64: TextStyle = TextStyle(
+      fontWeight = FontWeight.Bold,
+      fontSize = 64.sp, // TODO - replace with token when available
```
marianeum, line 49: these will be added in a future change when the tokens are available

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkTypography.kt
```diff
@@ -0,0 +1,125 @@
...
+
+import androidx.compose.runtime.Immutable
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.em
+import androidx.compose.ui.unit.sp
+
+@Immutable
+data class BpkTypography internal constructor(
```
marianeum, line 29: again this all follows the material typography setup

app/src/internal/java/net/skyscanner/backpack/demo/AppBpkTheme.kt
```diff
@@ -0,0 +1,48 @@
...
+      Font(R.font.skyscanner_relative_android_book),
+      Font(R.font.skyscanner_relative_android_bold, weight = FontWeight.Bold)
+    )
+  )
+  CompositionLocalProvider(
+    LocalBpkTypography provides typography,
+  ) {
+    BpkTheme(
+      typography = typography,
+      content = content
```
bvitaliyg, line 45: Can you please add trailing commas?

app/src/oss/java/net/skyscanner/backpack/demo/AppBpkTheme.kt
```diff
@@ -0,0 +1,40 @@
...
+package net.skyscanner.backpack.demo
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.theme.LocalBpkTypography
+import net.skyscanner.backpack.compose.tokens.BpkTypography
+
+@Composable
+fun AppBpkTheme(
```
bvitaliyg, line 28: BackpackDemoTheme?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
bvitaliyg: I'm not sure we actually need to create a composition local just for fonts. I'd rather prefer to have the theme attributes as a data class injected by single composition local. Compose may have its own reason to state typography explicitly, but we're not.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
```diff
@@ -0,0 +1,66 @@
...
+  BpkTypography()
+}
+
+@Composable
+fun BpkTheme(
+  typography: BpkTypography = BpkTheme.typography,
```
bvitaliyg, line 35: I don't think typography should be customisable for the theme. Instead we could try to search for the fonts using string constants.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
```diff
@@ -0,0 +1,66 @@
...
+  content: @Composable () -> Unit
+) {
+  CompositionLocalProvider(
+    LocalBpkTypography provides typography,
+  ) {
+    val materialTypography = Typography(
```
bvitaliyg, line 41: Was the mapping approved by the designers?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkTypography.kt
```diff
@@ -0,0 +1,125 @@
...
+    bodyLongform = bodyLongform.withDefaultFontFamily(defaultFontFamily),
+    bodyDefault = bodyDefault.withDefaultFontFamily(defaultFontFamily),
+    footnote = footnote.withDefaultFontFamily(defaultFontFamily),
+    caption = caption12.withDefaultFontFamily(defaultFontFamily),
+    label16 = label16.withDefaultFontFamily(defaultFontFamily),
+    label14 = label14.withDefaultFontFamily(defaultFontFamily),
+  )
+}
+
+private fun TextStyle.withDefaultFontFamily(default: FontFamily): TextStyle {
```
bvitaliyg, line 123: It'd be more easier for me if we'd just inlined it to single copy call with default fontFamily.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkTypography.kt
bvitaliyg: I don't think we need to use Immutable here anymore, since Compose compiler knows how to calculate it itself for cases like this.
There some context you may find userful:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/compiler/compiler-hosted/integration-tests/src/test/java/androidx/compose/compiler/plugins/kotlin/ClassStabilityTransformTests.kt;l=30?q=ClassStabilityTransformTests&sq=&ss=androidx

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
marianeum: We'll have to make it customizable to some extent - at least to override the font. i'm not sure what you mean with searching using string constants

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
marianeum: yep this was provided by design

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
marianeum: What would be the advantage of a data class? That feels like an additional extra wrapping layer that feels unnecessary - especially considering things like typography need to be somewhat customisable for the custom font

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
bvitaliyg: Can we take defaultFontFalimy here instead of entire typography?

bvitaliyg (approved):

PR #780, opened 2021-11-17
KOA-4818 Add basic support for publishing compose package

marianeum:
This adds a very basic support for being able to publish a compose package. This is just a temporary solution until we migrate away from JS for publishing (aka terrible code ™️)

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


.github/workflows/ci.yml
```diff
@@ -127,7 +127,7 @@ jobs:
           api-level: ${{ env.api-level }}
           force-avd-creation: false
           emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          script: ./gradlew :Backpack:connected${{ env.config }}AndroidTest
+          script: ./gradlew :Backpack:connected${{ env.config }}AndroidTest :backpack-compose:connected${{ env.config }}AndroidTest
```
marianeum, line 130: make sure we also build the compose tests on CI (there are none right now, but there will be in the future)

backpack-common/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <color name="bpkTextPrimary">#000000</color>
```
marianeum, line 3: this is... annoying. i found that when running the compose tests that there was a compilation error. we refer to the primary colour in the icons. in compose tinting will happen elsewhere. in the old view system this will get overridden in the other module

backpack-compose/build.gradle.kts
```diff
@@ -20,6 +20,12 @@ plugins {
   kotlin("android")
 }
 
+ext {
```
marianeum, line 23: apply publishing in compose

backpack-compose/build.gradle.kts
```diff
@@ -29,6 +35,13 @@ android {
   composeOptions {
     kotlinCompilerExtensionVersion = rootProject.ext.get("compose_version") as String
   }
+
+  packagingOptions {
```
marianeum, line 39: this was also needed for running the tests

package.json
```diff
@@ -70,7 +70,7 @@
     "husky": "^7.0.0",
     "inquirer": "^8.1.2",
     "js-yaml": "^4.1.0",
-    "lint-staged": "^12.0.2",
+    "lint-staged": "^11.2.6",
```
marianeum, line 73: had to downgrade this as v12 is broken

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "preinstall": "npx ensure-node-env",
     "emulator:build": "./gradlew build",
-    "emulator:instrumentation": "./gradlew Backpack:connectedDebugAndroidTest -PlocalTests=true",
+    "emulator:instrumentation": "./gradlew Backpack:connectedDebugAndroidTest backpack-compose:connectedDebugAndroidTest -PlocalTests=true",
```
marianeum, line 12: same as on CI - run compose tests as well. this was the easiest way to do it - we just always run the full test suite no matter what gets published

release.js
```diff
@@ -25,39 +25,24 @@ const releaseit = require('release-it');
...
-        name: `${key} (${i[key]})`,
-        value: `${i[key]}`,
-      };
-    }),
-  },
-];
+const compose = "compose"
+const android = "android"
+
+const package_questions = [
```
marianeum, line 35: add a new question for what should be released

release.js
```diff
@@ -104,17 +89,63 @@ async function amendVersionsFile(key, version) {
   }
 }
 
+async function versionPrompt(currentVersion) {
```
marianeum, line 92: just moved into a function

release.js
```diff
@@ -104,17 +89,63 @@ async function amendVersionsFile(key, version) {
...
     childProcess.execSync('./gradlew :Backpack:checkMavenCredentials');
-    const { version } = await inquirer.prompt(questions);
+    const packageResult = await inquirer.prompt(package_questions);
+    const packageType = packageResult.type
+    const packageVersion = (packageType == android) ? versions.android : versions.compose
+    const { version } = await versionPrompt(packageVersion);
     const commonVersion = semver.inc(versions.common, 'minor')
 
-    await amendGradleFiles('BpkAndroidVersion', version);
+    if (packageType == android) {
```
marianeum, line 134: just a bunch of if checks depending on which package to publish. not ideal, but it'll do for now

release.js
```diff
@@ -134,7 +165,8 @@ async function release() {
     await releaseit(releaseOptions);
     const publishTask = 'publishMavenPublicationToSonatypeRepository';
     const releaseTask = 'closeAndReleaseSonatypeStagingRepository';
-    childProcess.execSync(`./gradlew :backpack-common:${publishTask} :Backpack:${publishTask} ${releaseTask}`);
+    const publishModule = (packageType == android) ? 'Backpack' : 'backpack-compose'
```
marianeum, line 168: publish the right package depending on the type

release.js
olliecurtis: You could always extract it to a function given it looks to be doing the same thing for each version and then you can just pass in the Name, type and version?

olliecurtis (approved):

backpack-compose/build.gradle.kts
```diff
@@ -29,6 +35,13 @@ android {
...
+    resources.excludes.add("**/attach_hotspot_windows.dll")
```
zgeor, line 40: What is this specific DLL for?

marianeum: It was as part of the coroutines test dependency i think

package.json
frugoman: I've read that from 12, we don't need to `git add` anymore on the staged task. have you tried this?

backpack-compose/build.gradle.kts
zgeor: right I see. I went through the coroutines issue https://github.com/Kotlin/kotlinx.coroutines/issues/2023

package.json
zgeor: This is downgrading quite a few dependencies. Likely will introduce vulnerabilities.

zgeor (approved):

package.json
marianeum: This was upgraded & released very recently (few days ago), but the issue only appeared once we made changes (as lint-staged only throws errors on modified code). basically the error message is not relevant to the code, so you don't know what the lint issue is with v12 - I had a chat with @olliecurtis about this - it's to do with how we use husky to call lint-staged - in v12 it is no longer possible to do this. we'll have the same issue on all other projects

PR #812, opened 2021-12-09
KOA-3791 Create compose BpkText component

marianeum:
Created a simple wrapper around `Text`.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -93,7 +93,9 @@ object DocsRegistry {
     ViewScreenshot("Text - View - Default", "default"),
     ViewScreenshot("Text - View - Emphasized", "emphasized"),
     ViewScreenshot("Text - View - Heavy", "heavy"),
-    ComposeScreenshot("Text - Compose", "compose"),
+    ComposeScreenshot("Text - Compose - Hero", "hero"),
```
marianeum, line 96: split these up, so they fit on the screenshots

app/src/androidTest/java/net/skyscanner/backpack/compose/text/BpkTextTest.kt
```diff
@@ -29,6 +32,21 @@ class BpkTextTest : BpkSnapshotTest() {
 
   @Test
   fun default() = composed {
-    Text(text = "Sample")
+    DefaultText()
```
marianeum, line 35: something i thought of when setting up tests here: maybe we should have previews for what we test as that might be helpful when making changes to quickly verify different properties

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
```diff
@@ -21,39 +21,118 @@ package net.skyscanner.backpack.demo.compose
...
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.withStyle
+import androidx.compose.ui.tooling.preview.Preview
+import net.skyscanner.backpack.compose.text.BpkText
 import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.tokens.BpkDimension
+import net.skyscanner.backpack.demo.BackpackDemoTheme
 
+@Preview
```
marianeum, line 36: same for the demo app composables: create a preview for everything for quick iteration. only overhead is having to supply the theme for everything as we've set it up to crash when it isn't wrapped in a theme when accessing typography, colours etc

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/text/BpkText.kt
```diff
@@ -0,0 +1,86 @@
...
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.TextLayoutResult
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.text.style.TextDecoration
+import androidx.compose.ui.text.style.TextOverflow
+import net.skyscanner.backpack.compose.theme.BpkTheme
+
+@Composable
+fun BpkText(
```
marianeum, line 35: this is basically `Text` with all the text style properties removed apart from color + style

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/text/BpkText.kt
```diff
@@ -0,0 +1,86 @@
...
+    onTextLayout = onTextLayout,
+    style = style,
+  )
+}
+
+@Composable
+fun BpkText(
+  text: AnnotatedString,
+  modifier: Modifier = Modifier,
+  color: Color = BpkTheme.colors.textPrimary,
```
bvitaliyg, line 65: Why not Color.Unspecified?

docs/compose/Text/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Example of a text composable with `label1` style and secondary text color
+
+```Kotlin
+import androidx.compose.runtime.Composable
+import net.skyscanner.backpack.compose.text.BpkText
+import net.skyscanner.backpack.compose.theme.BpkTheme
+
+@Composable
+fun FlightText() {
```
bvitaliyg, line 17: I think we can just have a sample of the component without the wrapper function :)

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
```diff
@@ -21,39 +21,118 @@ package net.skyscanner.backpack.demo.compose
...
         text = style.second
       )
     }
   }
 }
+
+@Preview
+@Composable
+fun DefaultText() {
+  BackpackDemoTheme {
```
bvitaliyg, line 98: Can we add BackpackDemo theme to the base Story/Story composable?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/text/BpkText.kt
marianeum: ah that's a mistake!

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
```diff
@@ -21,39 +21,118 @@ package net.skyscanner.backpack.demo.compose
...
+  }
+}
+
+@Preview
+@Composable
+fun ColoredText() {
+  BackpackDemoTheme {
+    BpkText(
+      text = "Sample",
+      color = BpkTheme.colors.textSecondary,
```
bvitaliyg, line 109: Let's use primary (skyBlue) for better visibility.

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
```diff
@@ -21,39 +21,118 @@ package net.skyscanner.backpack.demo.compose
...
+}
+
+@Preview
+@Composable
+fun AnnotatedText() {
+  BackpackDemoTheme {
+    BpkText(
+      text = buildAnnotatedString {
+        append("Sample ")
+        withStyle(style = SpanStyle(color = BpkTheme.colors.textSecondary)) {
```
bvitaliyg, line 132: And primary here as well

docs/compose/Text/README.md
marianeum: yeah i was wondering about that - i thought it might be a little clearer, but it's probably repetitive

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
marianeum: It is there - the theme is needed for the preview

bvitaliyg: I'd remove the preview – we have an app for this :) In reality it's faster and easier to assemble the app then wait for preview :)

docs/compose/Text/README.md
bvitaliyg: Yeah, it should be clear that you need to use @ Composable functions for the components - so no need to have a wrapper for each sample :)

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
marianeum: I know they're working on making previews a lot faster - I find it quite helpful to see all different options in one place in the preview. The reason why it doesn't work in the preview is because we throw an error when somebody accesses typography/colour from the theme without something being wrapped in the `BpkTheme`

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -106,16 +102,12 @@ open class BpkSnapshotTest {
...
-                    .background(background.takeOrElse { BpkTheme.colors.background }),
-                  content = { content() },
-                )
-              }
-            }
+            BackpackPreview(
+              modifier = Modifier.size(size.width.dp, size.height.dp),
+              background = background,
+              providers = providers,
+              content = { content() }
```
bvitaliyg, line 109: We can just pass the content here without lambda :) Also, could you please add a trailing comma?

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
```diff
@@ -21,39 +21,117 @@ package net.skyscanner.backpack.demo.compose
...
         style = style.first,
         text = style.second
       )
     }
   }
 }
+
+@Preview
+@Composable
+fun DefaultText() {
```
bvitaliyg, line 96: Add we add Preview or Story suffix for this cases?

bvitaliyg (commented):
Just few small things

app/src/main/java/net/skyscanner/backpack/demo/compose/Text.kt
marianeum: I used `Example` as that felt more "correct" - i.e. we're previewing an example of a default text component

bvitaliyg:
First Compose component! 🎉

PR #813, opened 2021-12-09
Added BpkCard + demo

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


app/src/main/java/net/skyscanner/backpack/demo/compose/CardStory.kt
```diff
@@ -0,0 +1,73 @@
...
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+
+@Composable
+fun CardStory() {
+  Column(
+    modifier = Modifier.padding(horizontal = BpkSpacing.Base, vertical = BpkSpacing.Xxl),
+    verticalArrangement = Arrangement.spacedBy(BpkSpacing.Base),
+  ) {
+
+    BpkCard(
```
marianeum, line 28: can the individual cards be separate functions with preview to make it easier to read?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
```diff
@@ -0,0 +1,74 @@
...
+  onClick: (() -> Unit)? = null,
+  interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+  onClickLabel: String? = null,
+  role: Role? = null,
+  content: @Composable () -> Unit,
+) {
+
+  val focused by interactionSource.collectIsFocusedAsState()
+  val pressed by interactionSource.collectIsPressedAsState()
+  val elevated = focused || pressed
```
marianeum, line 39: If a card is pressed AND focused it won't look any different to a focused card - is that intended design?
also can multiple cards have a focused state or just one? i believe right now the focused property is set on multiple cards, rather than just a single one

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
```diff
@@ -0,0 +1,74 @@
...
+      else -> BpkTheme.colors.backgroundElevation01
+    }
+  )
+
+  val size = when (corner) {
+    BpkCardCorner.Small -> BpkBorderRadius.Md
+    BpkCardCorner.Large -> BpkBorderRadius.Lg
+  }
+
+  Card(
```
marianeum, line 60: Looking at the material card there's two separate card implementations depending on if there's click support vs not. looking at surface the `clickAndSemanticsModifier` is quite different depending on if it is or not. i suspect there might be subtle differences in the behaviour and the difference is there for a reason - especially around accessibility. have you tested with talkback etc?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
```diff
@@ -0,0 +1,74 @@
...
+    modifier = modifier,
+    shape = RoundedCornerShape(size = size),
+    backgroundColor = backgroundColor,
+    contentColor = BpkTheme.colors.textPrimary,
+    elevation = elevation,
+    onClick = onClick ?: {},
+    onClickLabel = onClickLabel,
+    interactionSource = interactionSource,
+    enabled = onClick != null,
```
marianeum, line 69: similar to above the default card has a separate enabled & onclick behaviour. semantically there is a difference between a card that has no click behaviour and a card that has one, but is disabled

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
```diff
@@ -0,0 +1,74 @@
...
+import net.skyscanner.backpack.compose.tokens.BpkElevation
+
+enum class BpkCardCorner {
+  Small,
+  Large,
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun BpkCard(
```
marianeum, line 27: the previous card had a padded vs not behaviour to keep the padding consistent to 16dp - is it intentional to remove that?

docs/compose/Card/README.md
```diff
@@ -0,0 +1,44 @@
...
+```Kotlin
+import net.skyscanner.backpack.compose.card.BpkCard
+
+val interactionSource = remember {
+  MutableInteractionSource().apply {
+    tryEmit(FocusInteraction.Focus())
+  }
+}
+
+BpkCard(interfactionSource = interactionSource) {
```
marianeum, line 41: can we also include the click behaviour here if it's a clickable card?

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -50,11 +50,12 @@ object DocsRegistry {
     ViewScreenshot("Calendar 2 - Pre-selected range", "range", ::setupCalendar2),
     ViewScreenshot("Calendar 2 - Day colours", "colored", ::setupCalendar2),
     ViewScreenshot("Calendar 2 - Day labels", "labeled", ::setupCalendar2),
-    ViewScreenshot("Card - Default", "default"),
-    ViewScreenshot("Card - Without padding", "without-padding"),
-    ViewScreenshot("Card - Selected", "selected"),
-    ViewScreenshot("Card - Corner style large", "corner-style-large"),
-    ViewScreenshot("Card - With divider", "with-divider"),
+    ViewScreenshot("Card - View- Default", "default"),
```
marianeum, line 53: i think there's spaces missing

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -50,11 +50,12 @@ object DocsRegistry {
...
+    ViewScreenshot("Card - View- Without padding", "without-padding"),
+    ViewScreenshot("Card - View- Selected", "selected"),
+    ViewScreenshot("Card - View- Corner style large", "corner-style-large"),
+    ViewScreenshot("Card - View- With divider", "with-divider"),
+    ComposeScreenshot("Card - Compose", "default"),
```
marianeum, line 58: looks like the screenshot wasn't generated. does all the content fit without scrolling?

app/src/main/java/net/skyscanner/backpack/demo/compose/CardStory.kt
bvitaliyg: It will not always be possible. For instance, in this particial case we have card wich switches between focused and non-focused state – if we extract it as a preview, it will display only one state be default.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
bvitaliyg: We have only one variant of focused card in our design system and I checked with designers what we want it either for focused or pressed state. Focused by definition means something one – do you have any samples when several cards are focused?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
bvitaliyg: If you check out the `Card` implementation code, it reffers to `Surface`, which have `clickAndSemanticsModifier` you've mentioned. This modifier is set to be standart `Modifier.clickable`, so the behaviour is practically the same. Is difference is in the modifiyers order (and that's why its extracted as an argument): the click modifiiyer is being applied only after shadow/background/clip modifiers.

The reason for that is basically the same reason why the overload with onClick lambda exists: if you just apply `Modifier.clickable` to default Card modifier, the ripple from the selection won't be clip according to the card shape – as it's being applied befire the clip.

So, it's should be exactly the same semantically – and even more precise since shadow should also not be in a screen reader selector indicator :D

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
bvitaliyg: The padding is not a part of the Firma design – and I think there could be some cases when padding will not be needed for the card content, so it should not be a property.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
bvitaliyg: There's a difference in the way the disabled state is handled for accessibility. Still though, the disabled flag will be set if there's no onclick listener.

docs/compose/Card/README.md
bvitaliyg: You mean to add a sample of a clickable card?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
marianeum: that's fine - still better to have a few separate functions to describe what they are

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
marianeum: currently all platforms have a padded & non-padded version, so this would be consistent with the rest of the implementations

docs/compose/Card/README.md
marianeum: the description above says it's a clickable example, but the code just has the focused state

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
marianeum: If you search for the focused property set in the app there's a bunch of cases where it is set to true as a general style, rather than a specifically focused state. This might be against the design system, but probably worth double checking with design if focused is meant as just a temporary state or can be set in general for multiple cards (especially as it is right now)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
marianeum: Yeah so there's two `Surface` implementations - one is `Modifier.clickable` - the other is `Modifier.semantics` the big difference is that one has `mergeDescendants` set to true (the clickable) vs false, which might mean that if there's clickable items within the card it might not behave correctly as it's treated as one card, rather than just a container. I believe this is similar to atomic vs non-atomic where it either gets treated as one item or multiple items.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
```diff
@@ -0,0 +1,92 @@
...
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.tokens.BpkBorderRadius
+import net.skyscanner.backpack.compose.tokens.BpkElevation
+
+enum class BpkCardCorner {
+  Small,
+  Large,
+}
+
+@OptIn(ExperimentalMaterialApi::class)
```
marianeum, line 43: is there any indicator around why `Card` is experimental?

bvitaliyg: I assume it's is mostly because it's a new API.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
bvitaliyg: I see. I would rather not have 2 diffrerent functions but rather used different card callds depending on whether onClick is null – what do you think?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
bvitaliyg: I think it's might be a mistake, but I will check with the design just in case.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
marianeum: Given that the non-clickable version doesn't have a lot of the properties in the clickable version it would be very confusing if you could set them, but they wouldn't do anything, so two different functions are probably better

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
marianeum: Just found it's been there since the 1.0 release - Is it possible to "swallow" the opt-in annotation, so anybody using `BpkCard` doesn't have to supply it? As we're wrapping it with our own API it isn't very relevant

app/src/main/java/net/skyscanner/backpack/demo/compose/CardStory.kt
marianeum: that's fine - still better to have a few separate functions to describe what they are

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/BpkCard.kt
```diff
@@ -0,0 +1,146 @@
...
+@Composable
+fun BpkCard(
+  onClick: () -> Unit,
+  modifier: Modifier = Modifier,
+  corner: BpkCardCorner = BpkCardCorner.Small,
+  padding: BpkCardPadding = BpkCardPadding.Small,
+  contentAlignment: Alignment = Alignment.TopStart,
```
bvitaliyg, line 61: There're some cases when you need to use Box inside card – thus contentAlignment with Box scope was also exposed, since we use a box inside anyway.

marianeum (approved):

PR #823, opened 2021-12-16
Add check on minValue and maxValue in BpkNudger

reneargento:
When using more than one `BpkNudger` in a screen, if we need to re-create them and set the value of `minValue` and `maxValue`, the update() method is called even if the field value has not changed.
This causes screen readers to read the values again, even if there was no change.
By adding an extra check we enable apps with this scenario to re-create `BpkNudgers` and only have one value (the one that was updated) to be read by the screen reader. 

+ [x] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


zgeor:
Hi @reneargento,

Can you please validate the functionality of a nudger with minus disabled and value of 0? Looks like your PR is failing on a changed screenshot for that functionality. Digging a bit more, shows that the lightmode RTL component looks like this 
nudger BpkNudgerTest_screenshotTestNudger_minusDisabled

while is should look like this 
nudger RTL light

PR #828, opened 2021-12-21
Enable uploading all failure artifacts from screenshots

zgeor:
Currently when screenshots fail we try to upload an archive containing the results. However, due to running 4 flavours, and defining the same name for the zipped artifact the results get overwriten. 

This PR adds the matrix variant name to the screenshot testing artifact. Which would allow us to upload all falures.

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
         if: ${{ failure() }}
         uses: actions/upload-artifact@v2
         with:
-          name: ${{ env.flavour }}${{ env.config }}-screenshot-test-output
+          name: ${{ env.flavour }}${{ env.config }}-${{ matrix.variant }}-screenshot-test-output
           path: app/build/screenshots${{ env.flavour }}${{ env.config }}AndroidTest/
           retention-days: 5
 


marianeum:
I found even when the file name was different only one of them ended up being uploaded (which is why I ended up naming it to a generic version to avoid the confusion of the file name being wrong) :(

zgeor:
I was playing with the pipeline yesterday. I really wanted to get the changed file names as a comment on the PR too. It turns out it is a bit more involved than I wanted it to be. But anyway, for the screenshot upload, you can have a look at this action run https://github.com/Skyscanner/backpack-android/actions/runs/1603210562 which demonstrates a failure in multiple screenshot test variants. The artifacts were uploaded succesfully. I had a few failure runs, and the behaviour was consistent

marianeum:
Amazing! Yes the comment would be the ideal state (or even showing the screenshot). Maybe we could add a comment directing to the artifacts page, like on web as a start? https://github.com/Skyscanner/backpack/blob/main/.github/workflows/ci.yml#L72

marianeum (approved):

PR #829, opened 2021-12-21
Updated the README.md

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,28 +31,23 @@ repositories {
 ```
 
 
-To set up **Backpack for Android View system**, add the following dependency
-to your `build.gradle` (in your app module) in the `dependencies` block:
+Add the following dependencies to your `build.gradle` (in your app module) in the `dependencies` block:
 
 ```gradle
-implementation 'net.skyscanner.backpack:backpack-android:33.2.0'
+implementation 'net.skyscanner.backpack:backpack-android:33.2.0' // for Android View system
+implementation 'net.skyscanner.backpack:backpack-compose:x.x.x' // for Compose (not released yet)
 ```
 
-You also need to inherit your app theme from Backpack:
+All Backpack components must to be used within `BpkTheme` scope:
+
+Android View system:
 ```xml
   <style name="AppTheme" parent="BpkTheme">
     <!-- Customize your theme here. -->
   </style>
 ```
 
-To set up **Backpack for Jetpack Compose**, add the following dependency
-to your `build.gradle` (in your app module) in the `dependencies` block:
-
-```gradle
-implementation 'net.skyscanner.backpack:backpack-compose:x.x.x' // not released yet
-```
-
-All Compose components must to be used within `BpkTheme` scope:
+Compose:
 
 ```kotlin
 BpkTheme {
@@ -68,11 +63,12 @@ You can find the list of the available components, as well as the code samples a
 
 All design system tokens exist as public Android resources. Here's the list of the token types and the samples of its usage:
 
-* border radii – `@dimen/bpkBorderRadiusSm` (XML), `R.dimen.bpkBorderRadiusSm` (Java/Kotlin)
-* colours – `@color/bpkSkyBlue` (XML), `R.color.bpkSkyBlue` (Java/Kotlin)
-* elevations – `@dimen/bpkElevationSm` (XML), `R.dimen.bpkElevationSm` (Java/Kotlin)
-* spacings – `@dimen/bpkSpacingLg` (XML), `R.dimen.bpkSpacingLg` (Java/Kotlin)
-* text styles – `?bpkTextBaseSize` (XML)
+* [border radii](https://backpack.github.io/tokens/radius) – `@dimen/bpkBorderRadiusSm` (XML), `R.dimen.bpkBorderRadiusSm` (Java/Kotlin)
+* [colours](https://backpack.github.io/guidelines/colour) – `@color/bpkSkyBlue` (XML), `R.color.bpkSkyBlue` (Java/Kotlin)
+* [semantic colours](https://backpack.github.io/guidelines/colour) – `@color/bpkTextPrimary` (XML), `R.color.bpkTextPrimary` (Java/Kotlin)
+* [elevations](https://backpack.github.io/tokens/shadow) – `@dimen/bpkElevationSm` (XML), `R.dimen.bpkElevationSm` (Java/Kotlin)
+* [spacings](https://backpack.github.io/tokens/spacing) – `@dimen/bpkSpacingLg` (XML), `R.dimen.bpkSpacingLg` (Java/Kotlin)
+* [text styles](https://backpack.github.io/tokens/typesetting) – `?bpkTextBaseSize` (XML)
 
 ### Jetpack Compose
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -8,21 +8,38 @@
...
+Backpack for Android supports two targets: **Android View** system and **Jetpack Compose**.
 
-Backpack is available through [Maven Central](https://search.maven.org/artifact/net.skyscanner.backpack/backpack-android). To install
-all of it, add the following line to your `build.gradle` (in your app module) in the `dependencies` block:
+Backpack is available through Maven Central. Thus, before adding Backpack to your project,
+make sure Maven Central is in your repositories list. Add this to your root `build.gradle`:
 
 ```gradle
-implementation 'net.skyscanner.backpack:backpack-android:33.2.0'
+repositories {
```
bvitaliyg, line 17: This is common part of Android and Compose, so it's extracted

README.md
```diff
@@ -8,21 +8,38 @@
...
+    mavenCentral()
+}
 ```
 
-If your app resolves dependencies through Maven Central you're all set, if not add in your root `build.gradle`
+### Demo application
```
bvitaliyg, line 22: Demo app is an important part of the app, so I pushed it up.

README.md
```diff
@@ -8,21 +8,38 @@
...
+The Backpack demo application is a good way of referring to the variants available for a component and their correct usage.
+The code is available under `/app` directory.
+[The app can be downloaded from App Center](https://install.appcenter.ms/orgs/backpack/apps/backpack-android/distribution_groups/everyone)
+or by scanning the QR code below.
+
+![QR code](qr.png)
+
+## Android View system
+
+You can find the list of the available components, as well as the code samples and the screenshots [here](https://backpack.github.io/components/text?platform=android).
```
bvitaliyg, line 32: Instead of having a duplicated list of the components (refering to Github without screenshots in place) it's better to redirect users to our website. The only issue is that we don't have (yet!) a link to all Android components, so I'm using most common one as a reference for now – with plans to support the URLs later

README.md
```diff
@@ -32,106 +49,31 @@ You also need to inherit your app theme from Backpack:
...
-
-## Usage
-
-### Radii
-
-The Backpack radii tokens are available as [dimension resource](Backpack/src/main/res/values/backpack.radii.xml).
-
-The supported tokens are
-
-+ `bpkBorderRadiusSm`
```
bvitaliyg, line 82: I have replaced it with the mentioning of how to use the tokens on each target

README.md
```diff
@@ -32,106 +49,31 @@ You also need to inherit your app theme from Backpack:
...
-R.color.bpkSkyBlue
+The design system tokens are located in a `net.skyscanner.backpack.compose.tokens` package.
+For instance, you can refer to color using following syntax: `BpkColor.SkyBlue`.
+
+### Installation
+
+Add the following dependency to your `build.gradle` (in your app module) in the `dependencies` block:
+
+```gradle
+implementation 'net.skyscanner.backpack:backpack-compose:1.0.0'
```
bvitaliyg, line 68: Not perfect – we don't have any release yet, and there's no a version autoupdate when making the release. However the version autoupdate can be done as a part of a Compose release ticket

README.md
marianeum: Maybe we can have a common installation part which has both `backpack-android` and `backpack-compose` defined (as in two lines for both - maybe a comment specifying what's what). Having the installation parts split across two sections feels a bit disconnected - I generally want to know: how can i add this thing to my app. and then: how do i use this.

README.md
```diff
@@ -8,21 +8,38 @@
...
+or by scanning the QR code below.
+
+![QR code](qr.png)
+
+## Android View system
+
+You can find the list of the available components, as well as the code samples and the screenshots [here](https://backpack.github.io/components/text?platform=android).
+
+All design system tokens exist as public Android resources.
+For instance, you can use colours by using `@color/bpkSkyBlue` in XML or using `R.color.bpkSkyBlue` in Java/Kotlin.
```
marianeum, line 35: Maybe we can still link to the different types - listing them all doesn't provide much benefit, but knowing where they are defined can be helpful (more so in compose than the existing system, but still)

README.md
```diff
@@ -32,106 +49,31 @@ You also need to inherit your app theme from Backpack:
...
-```xml
-<TextView
-  android:text="This is Backpack Blue 500!"
-  android:textColor="@color/bpkSkyBlue" />
-```
+You can find the list of the available components, as well as the code samples and the screenshots [here](https://backpack.github.io/components/text?platform=compose).
 
-```kotlin
-R.color.bpkSkyBlue
+The design system tokens are located in a `net.skyscanner.backpack.compose.tokens` package.
```
marianeum, line 60: same here - can we add a link as well? Also as a bunch of them need to be used via the theme, so that needs clarifying as well

README.md
marianeum, line 68: maybe x.x.x and a comment to make clear that this doesn't actually exist yet

README.md
```diff
@@ -32,106 +49,31 @@ You also need to inherit your app theme from Backpack:
...
-### Demo application
-The Backpack demo application is a good way of referring to the variants available for a component and their correct usage. The code is available under `/app` directory. [The app can be downloaded from App Center](https://install.appcenter.ms/orgs/backpack/apps/backpack-android/distribution_groups/everyone) or by scanning the QR code below.
+> Backpack for Jetpack Compose status is experimental.
 
-![QR code](qr.png)
+Backpack for Compose targets **stable Compose 1.0.0** release.
 
-## Components
-
-* [Badge](docs/view/Badge/README.md)
```
marianeum, line 44: if we're no longer listing the components we should also remove the danger warning & PR template part for this

README.md
```diff
@@ -8,131 +8,88 @@
...
-### Color
-```xml
-<TextView
-  android:text="This is Backpack Blue 500!"
-  android:textColor="@color/bpkSkyBlue" />
+```gradle
+implementation 'net.skyscanner.backpack:backpack-compose:x.x.x' // not released yet
 ```
 
+All Compose components must to be used within `BpkTheme` scope:
```
marianeum, line 55: that feels more like it should be in the section below - it's not really about installation. the view system might also need a blob about using the theme

README.md
```diff
@@ -8,131 +8,88 @@
...
-BpkGradients.getPrimary(context);
-BpkGradients.getPrimary(context, GradientDrawable.Orientation.LEFT_RIGHT);
-```
+### Android View system
+
+You can find the list of the available components, as well as the code samples and the screenshots [here](https://backpack.github.io/components/text?platform=android).
+
+All design system tokens exist as public Android resources. Here's the list of the token types and the samples of its usage:
+
+* border radii – `@dimen/bpkBorderRadiusSm` (XML), `R.dimen.bpkBorderRadiusSm` (Java/Kotlin)
```
marianeum, line 71: can you add the link back to these as well and maybe add an e.g. before the example
also drawables are missing in both sections

README.md
```diff
@@ -8,131 +8,88 @@
...
+> Backpack for Jetpack Compose status is experimental.
+
+Backpack for Compose targets **stable Compose 1.0.0** release.
+
+You can find the list of the available components, as well as the code samples and the screenshots [here](https://backpack.github.io/components/text?platform=compose).
+
+The design system tokens are located in a `net.skyscanner.backpack.compose.tokens` package.
+Here's the list of the token types and the samples of its usage:
+
+* border radii – `BpkBorderRadius.Lg`
```
marianeum, line 88: same here - an e.g. and a link to the file for all these + icons

README.md
```diff
@@ -8,131 +8,88 @@
...
+* colours – `BpkColor.SkyBlue`
```
marianeum, line 89: can you also link/add the semantic colours? these should be used when possible, rather than direct access to the colours

README.md
bvitaliyg: Drawables are not tokens, and they have their own page as components:
https://backpack.github.io/components/icon?platform=android

marianeum: we should add a compose version for this!

bvitaliyg: Good point

marianeum (approved):

PR #878, opened 2022-01-26
Update Compose design elements

bvitaliyg:

+ [x] `UNRELEASED.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/text/BpkText.kt
```diff
@@ -61,7 +62,7 @@ fun BpkText(
 fun BpkText(
   text: AnnotatedString,
   modifier: Modifier = Modifier,
-  color: Color = Color.Unspecified,
+  color: Color = BpkTheme.colors.textPrimary,
```
marianeum, line 65: Was this meant to be updated? I think just updating the alpha might be the better option, so we can still rely on the local colour for use in non-backpack components (e.g. the examples in the ThemeStory right now) - if it is intended then we should update it in both versions of `BpkText` - otherwise it'll be a different behaviour when using `AnnotatedString`

bvitaliyg: Good catch, it was committed by mistake :D

marianeum (approved):

PR #879, opened 2022-01-26
KOA-4805 Update typography styles

marianeum:
Updated typography styles based on the latest foundation version. Introduced new hero text styles for the existing view system and ensured all tokens are generated correctly for compose.

+ [x] `UNRELEASED.md`
+ [x] Tests


buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkTextUnits.kt
```diff
@@ -102,10 +105,15 @@ private fun toCompose(
           .builder(CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.UPPER_CAMEL, name), TextUnitClass)
           .addModifiers(if (internal) KModifier.INTERNAL else KModifier.PUBLIC)
           .initializer(buildCodeBlock {
+            val extension = if (namespace == "BpkLetterSpacing") {
```
marianeum, line 108: I'm not happy with this, but this is the only difference between letter spacing & the other text tokens. I originally thought of adding an enum for the type, but as this ultimately gets called from the `tokens.gradle.kts` file it's not a very nice solution either

Backpack/src/androidTest/java/net/skyscanner/backpack/text/BpkTextTest.kt
```diff
@@ -150,20 +150,25 @@ class BpkTextTest {
...
+        40,
         null,
         fontsAccessor(BpkText.XXXL, Weight.EMPHASIZED)
       ),
       arrayOf("bpkTextXlHeavy", "sans-serif-black", 24, null, fontsAccessor(BpkText.XL, Weight.HEAVY)),
-      arrayOf("bpkTextXxlHeavy", "sans-serif-black", 30, null, fontsAccessor(BpkText.XXL, Weight.HEAVY)),
-      arrayOf("bpkTextXxxlHeavy", "sans-serif-black", 36, null, fontsAccessor(BpkText.XXXL, Weight.HEAVY))
+      arrayOf("bpkTextXxlHeavy", "sans-serif-black", 32, null, fontsAccessor(BpkText.XXL, Weight.HEAVY)),
+      arrayOf("bpkTextXxxlHeavy", "sans-serif-black", 40, null, fontsAccessor(BpkText.XXXL, Weight.HEAVY)),
+      arrayOf("bpkTextHero1", "sans-serif-medium", 120, null, fontsAccessor(BpkText.HERO1, Weight.EMPHASIZED)),
```
bvitaliyg, line 167: I feel like at some point we need to remove thouse tests and use screenshot tests + generator tool tests instead

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -66,7 +71,7 @@ open class BpkText @JvmOverloads constructor(
     HEAVY
   }
 
-  @IntDef(XS, SM, BASE, LG, XL, XXL, XXXL, CAPS)
+  @IntDef(XS, SM, BASE, LG, XL, XXL, XXXL, CAPS, HERO1, HERO2, HERO3, HERO4, HERO5)
```
bvitaliyg, line 74: How do the old enum values combine with heading3/footnote etc?

app/src/main/res/layout/fragment_text_hero.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bvitaliyg, line 1: Can we include this to the same fragmet we use for texts?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkLineHeight.kt
```diff
@@ -18,15 +18,27 @@ import androidx.compose.ui.unit.TextUnit
...
+  internal val `5xl`: TextUnit = 64.0.sp
 
-  internal val Lg: TextUnit = 23.0.sp
+  internal val `6xl`: TextUnit = 76.0.sp
 
-  internal val Sm: TextUnit = 16.0.sp
+  internal val `7xl`: TextUnit = 96.0.sp
 
-  internal val Xl: TextUnit = 28.0.sp
+  internal val `8xl`: TextUnit = 120.0.sp
```
bvitaliyg, line 27: We may need to ask the designers to change thouse names, it may be causing dev experience issues on Android and some other issues on other platform.

Backpack/src/androidTest/java/net/skyscanner/backpack/text/BpkTextTest.kt
marianeum: Yeah we already have screenshot tests for all of these - I think I'll just remove these as they'll be a pain to maintain for all the other typography changes going forward

Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
marianeum: That will happen in a later change - the plan is to deprecate all the existing styles with the new ones being added. It's the same process we follow on iOS & Web as well

app/src/main/res/layout/fragment_text_hero.xml
marianeum: We currently already have various fragments for text. This follows the split for compose where it'll be hero, heading & body. I should add it as a screenshot for the docs, though!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkLineHeight.kt
marianeum: We specifically wanted to avoid naming them xxxxxxxxl to avoid mistyping the number of x's. These tokens will also only be internal to backpack and not external to the app - only the styles will which don't contain numbers.

bvitaliyg (approved):

PR #901, opened 2022-02-08
Koa 4840 android implement compose bpk button component

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] Tests

### DIFF
--- a/app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
+++ b/app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
@@ -1,3 +1,21 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package net.skyscanner.backpack.compose.button
 
 import androidx.compose.foundation.layout.Box
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButton.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButton.kt
@@ -1,3 +1,21 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package net.skyscanner.backpack.compose.button
 
 import androidx.compose.foundation.interaction.MutableInteractionSource
--- a/app/src/main/java/net/skyscanner/backpack/demo/compose/ButtonsStory.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/compose/ButtonsStory.kt
@@ -1,3 +1,21 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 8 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButton.kt
```diff
@@ -0,0 +1,93 @@
...
+import net.skyscanner.backpack.compose.button.internal.BpkButtonImpl
+import net.skyscanner.backpack.compose.button.internal.ButtonIcon
+import net.skyscanner.backpack.compose.button.internal.ButtonText
+
+@Composable
+fun BpkButton(
+  text: String,
+  modifier: Modifier = Modifier,
+  size: BpkButtonSize = BpkButtonSize.Default,
+  colors: BpkButtonColors = BpkButtonColors.Primary,
```
marianeum, line 18: naming nitpick: as a backpack user i would expect this to still be type - it is what we internally refer to the different options and while it mainly refers to colors there might be other differences. for example specifically `link` should not have horizontal padding to align it with other text (same as the old version)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButtonSize.kt
```diff
@@ -0,0 +1,48 @@
...
+    BpkSpacing.Base
+
+  internal open val horizontalSpacing: Dp =
+    BpkSpacing.Md
+
+  object Default : BpkButtonSize() {
+
+    override val iconSize = BpkSpacing.Base
+
+    override val minHeight = BpkSpacing.Xl + BpkSpacing.Sm
```
marianeum, line 28: while this is true right now i'm not sure doing math with spacing tokens is a good idea long term - we might see unexpected changes when the token values change. the other part is i believe design wanted the button to scale like sp as well

marianeum (commented):
so much simpler :D just a couple of comments

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButtonColors.kt
```diff
@@ -0,0 +1,156 @@
...
+    @Composable
+    override fun contentColor(interactionSource: InteractionSource): Color =
+      dynamicColorOf(BpkColor.White, BpkColor.Black)
+
+  }
+
+  object Secondary : BpkButtonColors() {
+
+    @Composable
+    override fun backgroundColor(interactionSource: InteractionSource): Color =
```
bvitaliyg, line 49: I don't like that we have a lot of internal code in public space – it makes reading such code as API reference more difficult. What do you think about moving these to extensions?

marianeum: Yes that's a good point - it'll also expose those colours externally and may be abused. Either extensions or use an enum again - some of the colours are also the same across different styles, so a when statement might be the simpler option for these

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -86,18 +88,22 @@ open class BpkSnapshotTest {
...
   }
 
+  fun assumeVariant(vararg variants: BpkTestVariant) {
+    assumeThat(BpkTestVariant.current, Matchers.isOneOf(*variants))
+  }
+
   protected fun composed(
     size: IntSize = IntSize(width, height),
     background: Color = Color.Unspecified,
+    className: String = TestNameDetector.getTestClass().removePrefix("net.skyscanner.backpack."),
```
marianeum, line 98: rather than exposing these could we add another property, like `testParameters` and update the `getScreenshotName` method accordingly, so it also supports parameters? that way this solution also works for non-compose tests

app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
```diff
@@ -0,0 +1,115 @@
...
+  fun iconOnly() = capture {
+    assumeVariant(BpkTestVariant.Default) // since its only icon, RTL doesn't matter
+    assumeTrue(type == BpkButtonType.Primary) // the layout the same across different button types
+    // icon is bigger on large size, so we need to test this
+
+    BpkButton(icon, "contentDescription", type = type, size = size, onClick = {})
+  }
+
+  private fun capture(content: @Composable () -> Unit) {
+    composed(className = "BpkButton($type, $size)", content = content)
```
marianeum, line 102: spaces in file names should generally be avoided - brackets feel wrong too - can we just use a `_` instead to match the current conventions?

app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
```diff
@@ -0,0 +1,115 @@
...
+import net.skyscanner.backpack.compose.icons.sm.LongArrowRight
+import org.hamcrest.Matchers.isOneOf
+import org.junit.Assume.assumeThat
+import org.junit.Assume.assumeTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@RunWith(Parameterized::class)
+class BpkButtonTest(flavour: Flavor) : BpkSnapshotTest() {
```
marianeum, line 21: some of the large screenshots look wonky because they go over multiple lines - can we increase the size of the screenshots?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButtonIconPosition.kt
```diff
@@ -0,0 +1,8 @@
+package net.skyscanner.backpack.compose.button
+
+enum class BpkButtonIconPosition {
```
marianeum, line 3: i'm a big fan of keeping enum definitions in the same file as where they are used to make it easier to see what the different properties are, rather than having to open lots of little files

marianeum (approved):
very nice! don't forget to update docs (and also include card)

PR #903, opened 2022-02-09
KOA-4827 Update BpkText API

marianeum:
Updated BpkText & related items to use the new typography system. This comes with some breaking changes - mainly around the `textStyle` & `weight` property on `BpkText` when set programmatically as this is only happening in a few instances.
Also updated all components to the new styles - some used no longer supported ones - namely badge & calendar - this also included changes in a few other components that use them, like the tab bar and bar chart

+ [x] `UNRELEASED.md`
+ [x] Tests


docs/view/Text/README.md
```diff
@@ -38,53 +36,13 @@ BpkText(context).apply {
 - `bpkFontFamilyEmphasized`
 - `bpkFontFamilyHeavy`
 
-- `bpkTextBaseAppearance`
```
marianeum, line 41: I removed these as we don't really want anybody to modify these anyways - happy to bring them back (but i would probably only put back the non-deprecated ones)

Backpack/src/main/res/values/attrs.xml
```diff
@@ -39,6 +39,19 @@
       <enum name="hero3" value="10" />
       <enum name="hero4" value="11" />
       <enum name="hero5" value="12" />
+      <enum name="heading1" value="13" />
```
marianeum, line 42: that's the new styles - unfortunately we can't deprecate the old ones in xml as android studio doesn't actually support it :(

bvitaliyg (approved):

PR #912, opened 2022-02-16
KOA-4792 Add lint check for compose component usage

marianeum:
Replicated the lint check we had in place for the existing view system to detect usage of non-compose components

+ [x] `UNRELEASED.md`
+ [x] Tests

### DIFF
--- a/backpack-compose/CHANGELOG.md
+++ b/backpack-compose/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 [Unreleased changes](UNRELEASED.md).
 
+# 0.2.0
+
+**Added:**
+
+- New `BpkButton` component
+- New `focused` property for `BpkCard`
+
 # 0.1.1
 
 **Fixed:**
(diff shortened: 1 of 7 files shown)


backpack-compose/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 [Unreleased changes](UNRELEASED.md).
 
+# 0.2.0
```
marianeum, line 5: looks like this wasn't updated with the last release

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.JavaContext
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.SourceCodeScanner
+import com.intellij.psi.PsiMethod
+import org.jetbrains.uast.UCallExpression
+
+@Suppress("UnstableApiUsage")
+class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
```
marianeum, line 33: basically the same as the existing detector, but much simpler because we only need to look at method calls, rather than lots of different options

bvitaliyg (approved):

PR #916, opened 2022-02-21
KOA-4828 Remove old text API

marianeum:
Removed the old text API and deprecated text styles, as well as the relative theme as it's no longer needed.

+ [x] `UNRELEASED.md`
+ [x] Tests


Backpack/src/main/res/values/backpack.button.xml
```diff
@@ -4,8 +4,4 @@
     <attr name="android:fontFamily" />
     <attr name="android:textSize" />
   </declare-styleable>
-
-  <style name="bpkButtonBase" parent="bpkTextSmEmphasized">
```
marianeum, line 8: I found there is one usage of this in the app that is unnecessary

Backpack/src/main/res/values/backpack.font.xml
```diff
@@ -1,80 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<resources>
-
-  <style name="bpkTextBaseRelative" parent="bpkTextBase">
```
marianeum, line 4: these are no longer needed - all the new styles didn't have these applied already as its set programmatically with the line height now being included

Backpack/src/main/res/values/attrs.xml
```diff
@@ -26,14 +26,6 @@
...
     <attr name="textStyle" format="enum">
-      <enum name="xs" value="0" />
-      <enum name="sm" value="1" />
-      <enum name="base" value="2" />
-      <enum name="lg" value="3" />
-      <enum name="xl" value="4" />
-      <enum name="xxl" value="5" />
-      <enum name="xxxl" value="6" />
-      <enum name="caps" value="7" />
       <enum name="hero1" value="8" />
```
bvitaliyg, line 29: Since we removing the old values, would that make sense to use the numbers starting with 0? That'd would make the future changes easier.

bvitaliyg (approved):

bvitaliyg (approved):

PR #918, opened 2022-02-23
Koa 4886 update button component to match new interim button styles

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] Tests


marianeum (approved):
changes look good, but let's hold off on merging this until we've added the new button styles as the outline button is broken in its current state and it'll block us from releasing any other changes

PR #926, opened 2022-02-25
KOA-4941 Update dependencies

marianeum:
Update all outdated dependencies. Although had to downgrade `activity-compose` as there's a weird conflict with compose where it seems to force an old version.

+ [x] `UNRELEASED.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarDispatchers.kt
```diff
@@ -36,10 +35,12 @@ object CalendarDispatchers {
   val Background: CoroutineDispatcher
     get() = background
 
+  @VisibleForTesting
```
marianeum, line 38: only these two methods are restricted for testing - the others can be used validly in production code

Backpack/src/test/java/net/skyscanner/backpack/util/TestStateMachine.kt
```diff
@@ -46,14 +48,16 @@ class TestStateMachineResultScope<State>(
...
 )
 
+@OptIn(ExperimentalCoroutinesApi::class)
 fun <SM : StateMachine<State, *>, State> testStateMachine(
   creator: CoroutineScope.(CoroutineDispatcher) -> SM,
   block: suspend TestStateMachineScope<SM, State>.() -> TestStateMachineResult
 ) {
-  val coroutineDispatcher = TestCoroutineDispatcher()
-  val coroutineScope = TestCoroutineScope() + coroutineDispatcher
+  val testCoroutineScheduler = TestCoroutineScheduler()
```
marianeum, line 56: we need to run everything on the same scheduler otherwise tests will fail now. also replaced the other bits with a non-deprecated version

build.gradle
```diff
@@ -124,9 +124,9 @@ ext {
   composeUiToolingPreview = "androidx.compose.ui:ui-tooling-preview:$compose_version"
   composeTestJUnit = "androidx.compose.ui:ui-test-junit4:$compose_version"
   composeUiTooling = "androidx.compose.ui:ui-tooling:$compose_version"
-  composeActivity = "androidx.activity:activity-compose:1.3.1"
+  composeActivity = "androidx.activity:activity-compose:1.3.0"
```
marianeum, line 127: for some reason this needs to be 1.3.0 - it seems there's a strict dependency onto this from compose. upgrading it will result in a version conflict

bvitaliyg (approved):

PR #940, opened 2022-03-16
Koa 4893 android update loading button behaviour

bvitaliyg:

+ [ ] `UNRELEASED.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -102,34 +100,23 @@ open class BpkButton(
       applyStyle(type.createStyle(context))
     }
 
-  private var iconDrawable: Drawable? = icon
```
bvitaliyg, line 105: No need to store it as we don't replace it with progress anymore.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -176,6 +164,7 @@ open class BpkButton(
     this.iconPosition = iconPosition
     this.minHeight = resources.getDimensionPixelSize(size.minHeight)
     this.iconSize = resources.getDimensionPixelSize(size.iconSize)
+    this.progress.centerRadius = iconSize / 2f - this.progress.strokeWidth
```
bvitaliyg, line 167: Progress size is calculated differently – to match it with our icon (and Compose impl version size) we need to substract the stroke width.

Backpack/src/main/java/net/skyscanner/backpack/button/internal/ButtonStyle.kt
```diff
@@ -37,12 +38,18 @@ internal class ButtonStyle(
...
-  private val contentColor: ColorStateList,
+  @ColorInt private val contentColor: Int,
+  @ColorInt private val contentPressedColor: Int,
+  @ColorInt private val contentDisabledColor: Int,
   @ColorInt private val rippleColor: Int,
 ) {
 
   fun getContentColor(): ColorStateList =
-    contentColor
+    colorStateList(
```
bvitaliyg, line 48: Basically, this is the bugfix for icon alignment. It fixes when you switch to custom ColorStateList creation instead of fetching from resources.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -144,6 +131,7 @@ open class BpkButton(
     }
 
   private var enabled = isEnabled
+  private var contentColors = textColors
```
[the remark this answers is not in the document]
marianeum: Is this actually needed? we should be able to get the colours based on the type/state

Backpack/src/main/java/net/skyscanner/backpack/button/internal/ButtonStyle.kt
marianeum: that is so weird! any idea why this is happening?

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -102,34 +100,23 @@ open class BpkButton(
...
-    iconDrawable = icon
-  }
-
   private val progress by unsafeLazy {
     CircularProgressDrawable(context).apply {
       setStyle(CircularProgressDrawable.DEFAULT)
-      centerRadius = resources.getDimension(R.dimen.bpkSpacingSm) * 1.3f
       strokeWidth = resources.getDimension(R.dimen.bpkSpacingSm) * 0.5f
-      start()
+      callback = this@BpkButton
```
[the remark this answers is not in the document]
bvitaliyg: It happens automatically -- callback is held by a weak reference and the object exists in the view anyway.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -144,6 +131,7 @@ open class BpkButton(
     }
 
   private var enabled = isEnabled
```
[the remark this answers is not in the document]
bvitaliyg: It's not a problem. It's just the button can be disabled in 2 cases: when it's been disabled explicitly or when it's in a loading state. Once loading is done, it need to switch back to whatever state it was before, that's why we store it.

Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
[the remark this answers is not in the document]
bvitaliyg: Hmm. It may be needed if user customise text color using these methods. Not officially supported, but could be used in theory.

Backpack/src/main/java/net/skyscanner/backpack/button/internal/ButtonStyle.kt
bvitaliyg: Resources has an internal cache for CST, it may be related. But no, I've figured it out by filtering the changes.

bvitaliyg: Indded

frugoman (commented):
docs/view/Button/screenshots/standard.png seems to be wrong? the link icon only button is not showing up in the screanshot

frugoman (approved):

PR #955, opened 2022-03-29
KOA-4973 Setup release pipeline

marianeum:
This separates our workflows into different triggers to ensure we can run different steps depending on the trigger. As all triggers still need to run all the usual checks moved those into a shared workflow that gets triggered by all of them. 

+ [x] `UNRELEASED.md`
+ [x] Tests


.github/workflows/build.yml
```diff
@@ -1,11 +1,7 @@
-name: CI
+name: Build
```
marianeum, line 1: this is a shared workflow that the other workflows call - includes all the tests/builds etc

.github/workflows/pr.yml
```diff
@@ -0,0 +1,14 @@
+name: PR
```
marianeum, line 1: this is the main PR trigger - just calls the other build steps now as no deploy steps were actually ran on PRs

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,42 @@
+name: Deploy
+
+on:
+  push:
+    branches: [main]
+
+concurrency:
```
marianeum, line 7: while i was at it i also added concurrency to cancel out previous builds on the same branch/PR to avoid wait times on CI (didn't do this on release as release should ALWAYS run)

.github/workflows/release.yml
```diff
@@ -0,0 +1,79 @@
+name: Release
```
marianeum, line 1: this is again what was in deploy, but what ran on tag releases - again removing the if conditions now that it only runs when needed

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,42 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+
+  Build:
```
marianeum, line 17: all 3 new workflows still run the same shared workflow

olliecurtis (approved):

.github/workflows/build.yml
```diff
@@ -1,11 +1,7 @@
...
 
 on:
-  push:
-    branches: [main]
-    tags: ['*.*.*']
-  pull_request:
-    branches: [main]
+  workflow_call:
```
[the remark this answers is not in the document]
marianeum: good shout - i'll give it a try!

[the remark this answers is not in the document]
marianeum: seems to work :D

.github/workflows/deploy.yml
marianeum: yep exactly - the `github.workflow` is there to ensure different triggers don't cancel each other out. `head_ref` is available for PRs and `run_id` for everything else

zgeor (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+
+      - name: Trigger backpack.github.io CI
+        run: |
+          curl -f -s -X POST -H "Content-Type: application/json" -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $GITHUB_TOKEN" -d "{\"ref\":\"main\",\"inputs\":{\"repo\":\"backpack-android\"}}" https://api.github.com/repos/Skyscanner/backpack-docs/actions/workflows/ci.yml/dispatches
```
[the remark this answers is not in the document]
marianeum: Not sure - this is just the existing deploy step - maybe @olliecurtis knows more about this

.github/workflows/build.yml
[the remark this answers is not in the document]
marianeum: Just fyi: just had a conversation with @bvitaliyg and we decided to keep the separate trigger to make it clearer what triggers we have and that these are the shared steps

bvitaliyg (approved):

PR #976, opened 2022-04-11
KOA-5011 Add new label 3 typography style

marianeum:
Added the new label 3 style to allow a smaller, bold text.

+ [x] `UNRELEASED.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/text/BpkText.kt
```diff
@@ -60,8 +60,9 @@ open class BpkText @JvmOverloads constructor(
     BodyDefault(12),
     Label1(13),
     Label2(14),
-    Footnote(15),
-    Caption(16),
+    Label3(15),
```
marianeum, line 63: similarly to last time we changed the enum values here - consumers are using the enum name rather than the value, so this is an internal change and changing the order is not a problem

app/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -50,104 +50,120 @@ class BpkButtonTest(flavour: Flavor) : BpkSnapshotTest() {
     )
 
   @Test
-  fun text() = capture {
+  fun text() {
```
marianeum, line 53: this makes sure we're not starting an activity for every variant, even if we don't run them

app/src/androidTest/java/net/skyscanner/backpack/compose/text/BpkTextTest.kt
```diff
@@ -31,22 +35,121 @@ import org.junit.runner.RunWith
...
   }
 
   @Test
-  fun annotated() = composed {
+  fun annotated() = capture {
     AnnotatedTextExample()
   }
+
+  @Test
+  fun bodyDefault() = capture {
```
marianeum, line 60: realised we didn't have tests for the styles in compose

bvitaliyg (approved):

PR #979, opened 2022-04-12
KOA-5024 Add primaryOnDark and linkOnDark button styles

marianeum:
We've introduced new button styles: `primaryOnDark` and `linkOnDark`. As part of this change we've decided to deprecate `BpkButtonLink` and move it into `BpkButton` to simplify things and align with compose.
Also split out the demo/docs pages to separate links as we had too many types now

+ [x] `UNRELEASED.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/button/BpkButton.kt
```diff
@@ -220,6 +208,22 @@ open class BpkButton(
     } else {
       setTextColor(style.getContentColor())
     }
+
+    var paddingHorizontal = paddingHorizontal
```
marianeum, line 212: padding now gets changed depending on the button type

app/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -55,7 +60,7 @@ class BpkButtonTest(flavour: Flavor) : BpkSnapshotTest() {
     // we want to see colors of all types
     // different sizes have different text style
 
-    capture {
+    capture(background = type.rowBackground()) {
```
marianeum, line 63: set the background in tests - this is for the on dark/light buttons to make sure they are visible correctly

app/src/main/java/net/skyscanner/backpack/demo/compose/ButtonsStory.kt
```diff
@@ -53,95 +61,54 @@ fun ButtonsStory(
...
-        onClick = ::load,
-      )
-    }
-
-    ButtonsColumn { type, enabled ->
-      BpkButton(
-        text = stringResource(R.string.button),
-        icon = icon,
-        position = BpkButtonIconPosition.End,
+  LazyColumn(modifier.fillMaxSize()) {
```
marianeum, line 64: separated out the link buttons here and moved back from column based to row based to apply the background for the on dark/light buttons

app/src/main/res/layout/fragment_button_large.xml
```diff
@@ -5,17 +5,16 @@
   android:layout_width="match_parent"
   android:layout_height="match_parent"
   android:gravity="top"
-  android:orientation="horizontal"
-  android:paddingHorizontal="@dimen/bpkSpacingBase"
+  android:orientation="vertical"
```
marianeum, line 8: these changed massively as they're now rendered as rows instead of columns - this makes it easier to apply the background. one day it might be worth making this more dynamic, rather than duplicating rows for each type

bvitaliyg (approved):

PR #990, opened 2022-04-21
KOA-5019 Setup release drafter

marianeum:
This sets up release drafter to automate the changelog creation. This is based on the playground implementation, but still relies on us manually pushing the tag after a release.

+ [x] `UNRELEASED.md`
+ [x] Tests


.github/pull_request_template.md
```diff
@@ -2,13 +2,17 @@
 Thanks for contributing to Backpack :pray:
 
 Please include a description of the changes you are introducing and some screenshots if appropriate.
+
+Please ensure your pull request title is clear as it will be used to generate the changelog.
+
+Add `major`, `minor` or `patch` label depending on the change according to [semver](semver.org) or `skip-changelog` if the change shouldn't be added to the changelog (e.g. a change to a test or documentation)
```
marianeum, line 8: there's a check to ensure the label was added, but also mentioned it here to avoid confusion

.github/pull_request_template.md
```diff
@@ -2,13 +2,17 @@
...
 -->
 
 + [ ] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-android/blob/main/CONTRIBUTING.md)
 
 Remember to include the following changes:
-+ [ ] `UNRELEASED.md`
++ [ ] Component `README.md`
```
marianeum, line 14: unreleased gone! but i realised after we removed readme from here that we still have the readme for the component here.

.github/pull_request_template.md
```diff
@@ -2,13 +2,17 @@
...
 + [ ] Tests
 + [ ] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)
 
-_If you are curious about how we review, please read through the [code review guidelines](https://github.com/Skyscanner/backpack/blob/main/CODE_REVIEW_GUIDELINES.md)_
+If you are curious about how we review, please read through the [code review guidelines](https://github.com/Skyscanner/backpack/blob/main/CODE_REVIEW_GUIDELINES.md)
```
marianeum, line 18: spellcheck was unhappy about this?!?!

.github/workflows/main.yml
```diff
@@ -18,6 +18,14 @@ jobs:
     name: Build
     uses: ./.github/workflows/_build.yml
 
+  ReleaseDraft:
```
marianeum, line 21: drafting the release notes as a separate job to ensure it still gets drafted even if the other workflows on main fail

.github/workflows/release.yml
```diff
@@ -34,6 +34,13 @@ jobs:
           java-version: '11'
           distribution: 'adopt'
 
+      - name: Publish release
```
marianeum, line 37: this publises the latest tag

.github/workflows/main.yml
frugoman: what's the purpose of creating the release notes if the other workflows fail?

.github/workflows/release.yml
```diff
@@ -20,7 +20,6 @@ jobs:
   Deploy:
     name: Deploy
     runs-on: ubuntu-latest
-    needs: [Build]
```
marianeum, line 23: I removed the dependency onto build for now as we currently do all these checks locally (and artifacts published) and sometimes CI can be flaky, but we always want to make sure we publish the latest docs/release etc

.github/workflows/main.yml
marianeum: This runs on main after a PR gets merged. Right not CI is a bit flaky due to the android emulator, but we always want to make sure the release notes get drafted even if main fails for some reason

frugoman (approved):

.github/release-drafter.yml
```diff
@@ -0,0 +1,29 @@
...
+      - 'major'
+  minor:
+    labels:
+      - 'minor'
+  patch:
+    labels:
+      - 'patch'
+      - 'dependencies'
+  default: patch
+template: |
```
zgeor, line 27: It might be good to add $CONTRIBUTORS to the template. It will be quite nice to mention the people who contributed to the change.

.github/workflows/label-check.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  pull_request_target:
+    types: [opened, labeled, unlabeled, synchronize]
+
+jobs:
+  label-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: jesusvasquez333/verify-pr-label-action@v1.4.0
```
olliecurtis, line 12: have we done due diligence on this 3rd party action and verified it is safe to use and not at risk of malicious intent?

.github/release-drafter.yml
marianeum: Actually removed this again, but using the default template. I found if i added this it also added a second list of contributors below the changelog with images etc. Then I noticed that if I just use the default change template it includes the contributor in the change AND below (see https://github.com/marianeum/release-model-playground/releases)

.github/workflows/label-check.yml
marianeum: Yep I had a look and it's doing exactly what I'd expect it to do

marianeum: actually had another look and found this which is done by a company and also supports to ensure only one of the above labels is set - will swap it out https://github.com/agilepathway/label-checker

PR #1005, opened 2022-05-04
Improved Calendar2 accessibility.

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


Backpack/build.gradle
```diff
@@ -48,9 +48,8 @@ dependencies {
   androidTestImplementation rootProject.ext.testRules
   androidTestImplementation rootProject.ext.mockitoKotlin
   androidTestImplementation rootProject.ext.mockitoAndroid
-  testImplementation rootProject.ext.junit
-  testImplementation rootProject.ext.coroutinesTest
-  testImplementation rootProject.ext.threeTenAbp
+  androidTestImplementation rootProject.ext.junit
```
marianeum, line 51: hmm it sucks that we have to change these all from unit tests. maybe we should move the span creation to the ui layer, rather than having it as part of the model to be able to continue using unit tests and have a cleaner separation?

PR #1006, opened 2022-05-04
Exclude js and github actions from changelog

marianeum:
JS & github actions dependencies generally don't affect anything for the consumers, so we shouldn't include them in the changelog. The only exception to this is backpack dependencies which include token changes. these should be part of the changelog, so creating a separate config for them.

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,21 @@ updates:
   versioning-strategy: increase-if-necessary
   allow:
     - dependency-type: "direct"
+  ignore:
+    - dependency-name: "bpk-*"
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "10:00"
+  open-pull-requests-limit: 10
+  versioning-strategy: increase-if-necessary
+  labels:
+    - "bpk"
+    - "depencdencies"
+  allow:
+    - dependency-type: "direct"
+    - dependency-name: "bpk-*"
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,7 +32,7 @@ updates:
   versioning-strategy: increase-if-necessary
   labels:
     - "bpk"
-    - "depencdencies"
+    - "dependencies"
   allow:
     - dependency-type: "direct"
     - dependency-name: "bpk-*"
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -21,6 +21,21 @@ updates:
...
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "10:00"
+  open-pull-requests-limit: 10
+  versioning-strategy: increase-if-necessary
+  labels:
+    - "bpk"
+    - "depencdencies"
```
olliecurtis, line 35:
```suggestion
    - "dependencies"
```

marianeum: whoops, thanks!

marianeum:
argh looks like this is actually not working :(
> The property '#/updates/3' is a duplicate. Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'

.github/workflows/pr.yml
```diff
@@ -17,3 +17,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Fetch Dependabot metadata
+        id: dependabot-metadata
+        uses: dependabot/fetch-metadata@v1.3.1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Add bpk label
+        if: contains('bpk-', steps.dependabot-metadata.outputs.dependency-names[0])
```
marianeum, line 31: is this syntax correct? we will find out...

.github/workflows/pr.yml
```diff
@@ -17,3 +17,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Fetch Dependabot metadata
+        id: dependabot-metadata
+        uses: dependabot/fetch-metadata@v1.3.1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Add bpk label
+        if: contains(steps.dependabot-metadata.outputs.dependency-names[0], 'bpk-')
```
marianeum, line 31: is this syntax correct? we will find out...

frugoman (approved):

PR #1019, opened 2022-05-12
KOA-4973 Move release process to CI

marianeum:
Update the release process to publish on CI, rather than locally for a simpler, safer way of releasing. I've tested this as best as I can in a separate repo, but the only real test will be to actually run the release and see what happens - there may be minor issues in place which we may not find out until the release process was ran.

+ [x] Component `README.md`
+ [x] Tests


.github/workflows/release.yml
```diff
@@ -1,8 +1,8 @@
 name: Release
 
 on:
-  push:
-    tags: ['*.*.*']
+  release:
```
marianeum, line 4: this will only trigger when actually publishing releases - not just if a tag matches to be more explicit

.github/workflows/release.yml
```diff
@@ -33,21 +34,19 @@ jobs:
...
 
-      - name: Get release tag
-        id: tag
-        uses: dawidd6/action-get-tag@v1
-#
-#      - name: Create Sonatype key
-#        run: base64 -d <<< $SONATYPE_SINGING_KEY_BASE64 > sonatype.gpg
-#        env:
-#          SONATYPE_SINGING_KEY_BASE64: ${{ secrets.SONATYPE_SINGING_KEY_BASE64 }}
+      - name: Publish artifacts
```
marianeum, line 42: This is essentially replacing what's currently in the node file

publish-root.gradle
```diff
@@ -10,6 +10,29 @@ if (localProps.exists()) {
   Properties p = new Properties()
   new FileInputStream(localProps).withCloseable { is -> p.load(is) }
   p.each { name, value -> ext[name] = value }
+} else {
+  Properties p = new Properties()
+  p.putAll(System.getenv())
```
marianeum, line 15: read the release properties from env variables - the properties need to match the current naming as they're used by other dependencies

.github/workflows/release.yml
```diff
@@ -33,21 +34,19 @@ jobs:
...
+        run: ./gradlew checkMavenCredentials publishMavenPublicationToSonatypeRepository closeAndReleaseSonatypeStagingRepository -Pversion=${{ github.event.release.tag_name }}
```
bvitaliyg, line 43: Can we split this line in few?

.github/workflows/release.yml
```diff
@@ -33,21 +34,19 @@ jobs:
           java-version: '11'
           distribution: 'adopt'
 
-      - name: Publish release
```
bvitaliyg, line 36: I see no usages of release-drafter actions anymore. It's not required as now the CI is triggered by the release itself?

marianeum: Correct - before the release would be triggered by pushing the tag. Now it's the opposite

.github/workflows/release.yml
marianeum: I was originally considering splitting the credential check out to a separate action, but as all the env variables need to be passed through it makes sense executing it all in one go - not sure how i can split out one execution to separate lines

bvitaliyg: key: >
    long
    string here.

bvitaliyg (approved):
💯

PR #1027, opened 2022-05-16
Added compose BpkCheckbox

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


app/src/main/res/values/strings.xml
```diff
@@ -100,6 +100,8 @@
   <string name="text_hero5">Hero 5</string>
   <string name="button">Button</string>
 
+  <string name="toggle_default">Default</string>
```
bvitaliyg, line 103: This is used for toggleable that doesn't have "fixed" state being interactable

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -73,31 +71,33 @@ fun BpkRadioButton(
...
-    modifier = rowModifier,
     verticalAlignment = Alignment.CenterVertically,
+    horizontalArrangement = Arrangement.spacedBy(BpkDimension.Spacing.Sm),
+    modifier = modifier.applyIf(onClick != null) {
+      selectable(
+        selected = selected,
+        interactionSource = interactionSource,
+        indication = null,
+        role = Role.RadioButton,
+        onClick = onClick!!,
```
bvitaliyg, line 84: Unfortunately the compiler cannot undertsand from the contract that onClick will never be null here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -107,8 +107,8 @@ private fun BpkRadioButtonImpl(
   selected: Boolean,
   onClick: (() -> Unit)?,
   modifier: Modifier = Modifier,
-  enabled: Boolean = true,
-  interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
```
bvitaliyg, line 111: No need to create default params for internal methods if it's expected to be used with passing-though params. I may lead to forgetting pass param and potential errors in the future.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Modifier.kt
```diff
@@ -30,3 +33,12 @@ internal fun Modifier.hideContentIf(hide: Boolean): Modifier =
       }
     }
   }
+
+
+@OptIn(ExperimentalContracts::class)
+internal inline fun Modifier.applyIf(predicate: Boolean, block: Modifier.() -> Modifier): Modifier {
+  contract {
```
bvitaliyg, line 40: I still left the contract here in case the compiler learns how to understand it in the future.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/checkbox/BpkCheckbox.kt
```diff
@@ -0,0 +1,167 @@
...
+  state: ToggleableState,
+  onClick: (() -> Unit)?,
+  modifier: Modifier = Modifier,
+  enabled: Boolean = true,
+  interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+  content: @Composable RowScope.(ToggleableState) -> Unit,
+) {
+  Row(
+    verticalAlignment = Alignment.CenterVertically,
+    horizontalArrangement = Arrangement.SpaceBetween,
```
marianeum, line 115: you could use `Arrangement.spacedBy(BpkDimension.Spacing.Sm)` here and then avoid setting the padding end of the checkbox

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/checkbox/BpkCheckbox.kt
```diff
@@ -0,0 +1,167 @@
...
+  // our design system isn't designed with the minimum touch target in mind at the moment.
+  // Disable the enforcement to avoid the extra padding
+  CompositionLocalProvider(LocalMinimumTouchTargetEnforcement provides false) {
+    TriStateCheckbox(
+      state = state,
+      onClick = onClick,
+      enabled = enabled,
+      modifier = modifier.scale(0.89f).semantics { invisibleToUser() },
+      interactionSource = interactionSource,
+      colors = CheckboxDefaults.colors(
```
marianeum, line 162: looks like we're setting a lot more colours in the view version - are those set correctly by the theme?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Modifier.kt
marianeum: The contract just specifies that it is only called once in place, not that the result is not nullable, so this will never work - you need a different contract/function for this

bvitaliyg: It doesn't work via results. A compiler can understand that the lambda will be called in place and therefore it can analyse that all the smart checks are done before the lambda will be valid. But there's no way to state that a predicate will be true in this case as well for the compiler.

marianeum: this might explain what i mean better: https://www.baeldung.com/kotlin/contracts you need a contract like this ` returns() implies (request != null)` 
the call in place only verifies that it'll be called once and can be used for initialising things - it does not do any further smart checks

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/checkbox/BpkCheckbox.kt
```diff
@@ -0,0 +1,167 @@
...
+      )
+    },
+  ) {
+
+    BpkCheckboxImpl(
+      modifier = Modifier.padding(end = BpkSpacing.Sm),
+      state = state,
+      enabled = enabled,
+      interactionSource = interactionSource,
+      onClick = onClick ?: {},
```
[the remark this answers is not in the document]
bvitaliyg: Yes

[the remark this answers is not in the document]
marianeum: In that case we probably shouldn't make this non-nullable as it'll indicate it is clickable when it is not

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -100,6 +105,11 @@ class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
       webName = "text",
       componentsToReplace = setOf("androidx.compose.material.Text")
     ),
+    TRI_STATE_CHECKBOX(
+      fullName = "net.skyscanner.backpack.compose.checkbox.BpkCheckbox",
+      webName = "checkbox",
+      componentsToReplace = setOf("androidx.compose.material.TriStateCheckbox")
```
marianeum, line 111: you can also provide multiple package names in `CHECKBOX`, rather than creating another enum here (similar to buttons)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/checkbox/BpkCheckbox.kt
```diff
@@ -0,0 +1,173 @@
...
+  interactionSource : MutableInteractionSource,
+) {
+  // our design system isn't designed with the minimum touch target in mind at the moment.
+  // Disable the enforcement to avoid the extra padding
+  CompositionLocalProvider(LocalMinimumTouchTargetEnforcement provides false) {
+    TriStateCheckbox(
+      state = state,
+      onClick = onClick,
+      enabled = enabled,
+      modifier = modifier.scale(0.89f).semantics { invisibleToUser() },
```
marianeum, line 162: can we have a constant for this?

marianeum (approved):

PR #1033, opened 2022-05-18
KOA-4927 Tweak compose RadioButton API and improve accessibility

marianeum:
This utilises the slot approach for the RadioButton, rather than providing a standalone component. Also  improved accessibility by announcing the trait on the parent and not making the RadioButton focusable, as it would be duplicate information.

+ [x] Component `README.md`
+ [x] Tests


app/src/main/java/net/skyscanner/backpack/demo/compose/ThemeStory.kt
```diff
@@ -92,10 +90,6 @@ fun ThemeStory() {
       }
     }
 
-    BpkButton(text = "Primary button", onClick = { })
```
marianeum, line 95: removed our components from the preview

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -22,59 +22,60 @@ import androidx.compose.foundation.clickable
...
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.invisibleToUser
+import androidx.compose.ui.semantics.semantics
 import net.skyscanner.backpack.compose.text.BpkText
 import net.skyscanner.backpack.compose.theme.BpkTheme
 import net.skyscanner.backpack.compose.tokens.BpkColor
 import net.skyscanner.backpack.compose.tokens.BpkDimension
 import net.skyscanner.backpack.compose.utils.dynamicColorOf
 
-@OptIn(ExperimentalMaterialApi::class)
```
marianeum, line 40: moved this to the bottom and private - no changes to the radiobutton itself except for the one called out

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -83,11 +84,42 @@ fun BpkRadioButton(
...
+      LocalTextStyle provides BpkTheme.typography.footnote,
+    ) {
+      content(selected)
+    }
+  }
+}
+
+@OptIn(ExperimentalMaterialApi::class, ExperimentalComposeUiApi::class)
+@Composable
+private fun BpkRadioButtonImpl(
```
marianeum, line 102: so that's the moved function

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -83,11 +84,42 @@ fun BpkRadioButton(
...
+  enabled: Boolean = true,
+  interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+) {
+  // our design system isn't designed with the minimum touch target in mind at the moment.
+  // Disable the enforcement to avoid the extra padding
+  CompositionLocalProvider(LocalMinimumTouchTargetEnforcement provides false) {
+    RadioButton(
+      selected = selected,
+      onClick = onClick,
+      modifier = modifier.semantics { invisibleToUser() },
```
marianeum, line 115: that's the one change to this function - make it invisible to the user - this makes it not focusable in accessibility mode as the whole row has the radiobutton role & handles the functionality (and provides the context around it as well)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -22,59 +22,60 @@ import androidx.compose.foundation.clickable
...
-  text: String,
   onClick: (() -> Unit)?,
   modifier: Modifier = Modifier,
   enabled: Boolean = true,
   interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+  content: @Composable RowScope.(Boolean) -> Unit,
 ) {
   val rowModifier = if (onClick != null) {
-    modifier.clickable(interactionSource = interactionSource, indication = null, onClick = onClick)
+    modifier.clickable(interactionSource = interactionSource, indication = null, role = Role.RadioButton, onClick = onClick)
```
marianeum, line 78: this makes the row the radiobutton as well - the announcement will include the text in the content, as well as the toggle behaviour/state

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -22,59 +22,60 @@ import androidx.compose.foundation.clickable
...
 import net.skyscanner.backpack.compose.theme.BpkTheme
 import net.skyscanner.backpack.compose.tokens.BpkColor
 import net.skyscanner.backpack.compose.tokens.BpkDimension
 import net.skyscanner.backpack.compose.utils.dynamicColorOf
 
-@OptIn(ExperimentalMaterialApi::class)
 @Composable
 fun BpkRadioButton(
   selected: Boolean,
+  text: String,
```
bvitaliyg, line 51: Can we make text the first param? That's the design I used for checkbox.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
bvitaliyg: Does it work okay with VoiceOver?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -83,11 +84,42 @@ fun BpkRadioButton(
...
-    BpkText(
-      text = text,
-      style = BpkTheme.typography.footnote,
-      color = if (enabled) BpkTheme.colors.textPrimary else dynamicColorOf(BpkColor.SkyGrayTint04, BpkColor.BlackTint06),
+    BpkRadioButtonImpl(selected = selected, onClick = onClick, enabled = enabled, interactionSource = interactionSource)
+    val contentColor =
+      if (enabled) BpkTheme.colors.textPrimary else dynamicColorOf(BpkColor.SkyGrayTint04, BpkColor.BlackTint06)
+    CompositionLocalProvider(
+      LocalContentColor provides contentColor,
+      LocalContentAlpha provides 1f,
```
bvitaliyg, line 92: I'm thinking maybe we shouldn't include setting content alpha here. The widgets which override it usually override it for its content and have good reasons for it.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
marianeum: yep - as i said it'll announce the content with the role & call to action to toggle (and announces the state when toggle is changed)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
bvitaliyg: I'd probably inline it and not make it as a function.

app/src/main/java/net/skyscanner/backpack/demo/compose/RadioButtonStory.kt
```diff
@@ -98,3 +101,20 @@ fun DisabledCheckedRadioButtonExample() {
...
+@Composable
+fun CustomContentRadioButtonExample() {
+  BackpackPreview {
+    var selected by remember { mutableStateOf(false) }
+    BpkRadioButton(
+      selected = selected,
+      onClick = { selected = !selected },
+    ) {
+      Column {
+        BpkText(text = stringResource(id = R.string.toggle_custom_title), style = BpkTheme.typography.heading5)
```
bvitaliyg, line 115: This may mislead used to assume that we actually provide title + subtitle implementation. I'd probably use something like:

Custom
Content

bvitaliyg (commented):
Just few ideas/suggestions

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
marianeum: i'd rather keep it as a separate function as it is cleaner. inlining has minimal benefit for this case based on the IDE hint, so don't think that's necessary

bvitaliyg (approved):

PR #1034, opened 2022-05-18
Tweak CI permissions for build steps

marianeum:
We need write permissions on pull requests for screenshot & build steps to be able to report the status correctly/comment on PRs

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -19,9 +19,8 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     timeout-minutes: 30
-
-    strategy:
-      fail-fast: false
+    permissions:
+      pull-requests: write
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


.github/workflows/_build.yml
```diff
@@ -19,9 +19,8 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     timeout-minutes: 30
-
-    strategy:
```
marianeum, line 23: this is only valid in combination with a matrix, so removing

.github/workflows/main.yml
```diff
@@ -16,6 +16,8 @@ jobs:
 
   Build:
     name: Build
+    permissions:
```
marianeum, line 19: the permission needs to be set where the workflow is being called as well, otherwise the steps fail as it won't be allowed to request the permissions

frugoman (approved):

PR #1046, opened 2022-05-23
Don't use MainActivity for screenshot tests

marianeum:
In an attempt to fix screenshot tests for dialogs I stumbled upon this issue of us using `MainActivity` for screenshot tests. This not only inflates unnecessary views, but also provides somewhat inaccurate results in some cases (ie the theming not applied for the snackbar) - some screenshots changed slightly with this change, but the changes are expected and correct. We also were using the whole content view, rather than just the specific compose view for compose tests. This doesn't match the experience with other screenshot tests.

+ [x] Component `README.md`
+ [x] Tests


app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -136,6 +137,22 @@ open class BpkSnapshotTest {
     }
   }
 
+  private fun View.findComposeView(): ComposeView? {
```
marianeum, line 140: this essentially loops through the view tree until we find the right view. we only have 1 compose view as a root for our tests

olliecurtis (approved):
LGTM!

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -136,6 +137,22 @@ open class BpkSnapshotTest {
...
+    if (this !is ViewGroup) return null
+    for (i in 0..childCount) {
+      val child = getChildAt(i)
+      if (child is ComposeView) {
+        return child
+      } else if (child is ViewGroup) {
+        val view = child.findComposeView()
+        if (view != null) {
```
frugoman, line 148: I think you can just return child.findComposeView() since you have the check and return case at the beginning of the method... no?

marianeum: Not quite - this will return null on the first child if it's not a compose view, so it won't go to the other children to check

PR #1063, opened 2022-05-30
KOA-5106 Update checkbox & radiobuttion paddings

marianeum:
These didn't match the design system

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/checkbox/BpkCheckbox.kt
```diff
@@ -78,6 +78,8 @@ open class BpkCheckbox @JvmOverloads constructor(
         intArrayOf(textDisabledColor, textEnabledColor)
       )
     )
+
+    setPaddingRelative(resources.getDimensionPixelSize(R.dimen.bpkSpacingSm), paddingTop, paddingEnd, paddingBottom)
```
marianeum, line 82: the view padding is different to the compose padding because the view one already includes some default padding within the drawable, but it ends up being the same

frugoman (approved):

PR #1064, opened 2022-05-31
OGM-512 Implement whole month calendar selection

ernestkamara:

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -59,23 +60,24 @@ data class CalendarParams(
...
+    object Single : SelectionMode
 
     /**
      * Only a range can be selected.
      * The boundaries of range are always non-disabled dates, but there could be disabled dates within the range.
+     *
+     * @param allowSelectWholeMonth allows a whole month selection.
      */
-    Range,
+    data class Range(val allowSelectWholeMonth: Boolean = false) : SelectionMode
```
sergiomarqmoura, line 80: Not sure if we discussed this before, but just curious about why we have `Range(val allowSelectWholeMonth: Boolean = false)` instead of having another type of the sealed interface, like `Month` (which is also a bit more consistent with the `CalendarSelection` class)?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarMonth.kt
```diff
@@ -43,16 +45,27 @@ internal inline fun CalendarMonth(
   monthsFormatter: SimpleDateFormat,
   weekFields: WeekFields,
   selection: CalendarSelection,
+  selectionMode: CalendarParams.SelectionMode,
+  selectWholeMonthLabel: String?,
   day: (YearMonth, LocalDate) -> CalendarCell.Day,
 ): CalendarMonth {
 
   val firstDay = days.first()
+  val allowWholeMonthSelection = (selectionMode as? CalendarParams.SelectionMode.Range)?.allowSelectWholeMonth ?: false
```
sergiomarqmoura, line 54: If we had a new `Month` selection mode, here would be just `selectionMode is CalendarParams.SelectionMode.Month`

sergiomarqmoura:
It seems like most (if not all) of the screenshots changed, there is an extra padding now (see [here](https://github.com/Skyscanner/backpack-android/pull/1064/files?short_path=6acbf23#diff-6acbf2324d6809f4383fb752a975715901fc349269be716ded5dadde3527c561)), is this expected?

sergiomarqmoura (commented):
Just a few clarifications needed on the calendar parameters, but looking good overall.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
ernestkamara: @sergiomarqmoura, Good point. However, this is the selection mode that describes the selection behaviour. @bvitaliyg and I agreed on this change instead of adding another sealed interface to minimise breaking changes.

Backpack/src/main/res/layout/view_bpk_calendar_header.xml
```diff
@@ -1,17 +1,46 @@
...
-  android:paddingEnd="@dimen/bpkSpacingBase"
-  android:paddingTop="@dimen/bpkSpacingLg"
-  android:paddingBottom="@dimen/bpkSpacingLg"
-  android:singleLine="true"
-  android:gravity="start|center_vertical"
-  app:textStyle="heading4"
-  tools:targetApi="p" />
+  android:clipToPadding="false"
+  android:clipChildren="false"
+  android:paddingTop="@dimen/bpkSpacingMd"
```
marianeum, line 9: it looks like the padding changed here as we have padding md + lg from the text view now, which explains the difference in the screenshots

sergiomarqmoura: Nice one @marianeum, that answers one of my questions

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
marianeum: ah yes I think I understand the distinction now - I think that makes sense then. Follow up question is: should the label be part of the same range data class as it's only available for the range mode as well and linked to the boolean?

ernestkamara: > ah yes I think I understand the distinction now - I think that makes sense then. Follow up question is: should the label be part of the same range data class as it's only available for the range mode as well and linked to the boolean?

@marianeum, after reviewing your comment above -  adding the label to the data class for the range selection should be enough. We don't actually need the boolean since we can check if the label is available.

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar2/data/CalendarRangeSelectionTests.kt
```diff
@@ -35,65 +36,69 @@ import org.junit.Test
...
       }
     }
   }
 
   @Test
   fun when_range_is_closed_selection_is_correct() {
     testCalendarWith(rangeSelection) {
-      stateMachine.onClick(firstDay)
-      stateMachine.onClick(lastDay)
+      stateMachine.onClick(CalendarAction.CalendarDayAction(firstDay))
```
bvitaliyg, line 60: I would still keep it as 2 methods instead of sealed classes here.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarComponent.kt
```diff
@@ -33,6 +34,11 @@ interface CalendarComponent {
    */
   val state: StateFlow<CalendarState>
 
+  /**
+   * The current effects the calendar emmit.
```
bvitaliyg, line 38:
```suggestion
   * The side effects the calendar emits.
```

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -59,23 +59,33 @@ data class CalendarParams(
...
+     * Describes the rangeable selection behaviour
      * The boundaries of range are always non-disabled dates, but there could be disabled dates within the range.
      */
-    Range,
+    sealed interface Rangeable : SelectionMode
+
+    /**
+     * A range of dates can be selected.
+     */
+    object Dates : Rangeable
```
bvitaliyg, line 82: New line?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarStateMachine.kt
```diff
@@ -125,3 +135,17 @@ internal fun CalendarState.dispatchSetSelection(selection: CalendarSelection): C
     cells = CalendarCells(params, selection),
   )
 }
+
+internal fun CalendarState.dispatchClick(data: CalendarCell.Header): CalendarState {
+  if (data.selectWholeMonthLabel.isNullOrEmpty()) return this
+  val selection = when (params.selectionMode) {
+    SelectionMode.Disabled -> selection
```
bvitaliyg, line 142:
```suggestion
    is SelectionMode.Disabled -> selection
```

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarStateMachine.kt
```diff
@@ -125,3 +135,17 @@ internal fun CalendarState.dispatchSetSelection(selection: CalendarSelection): C
...
+    SelectionMode.Single -> selection
```
bvitaliyg, line 143:
```suggestion
    is SelectionMode.Single -> selection
```

Backpack/src/main/res/layout/view_bpk_calendar_header.xml
```diff
@@ -1,17 +1,44 @@
 <?xml version="1.0" encoding="utf-8"?>
-<net.skyscanner.backpack.text.BpkText xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
bvitaliyg, line 2: We can still use LinearLayout here.

bvitaliyg (commented):
Looking good! Just few comments.

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar2/data/CalendarRangeSelectionTests.kt
ernestkamara: @bvitaliyg Initially, I had these two methods, however, the introduction of CalendarAction improves readability as well as makes adding a new action/listener to the calendar in the future easier. Do you still want me to change it?

CC: @sergiomarqmoura & @marianeum, any objection  👆🏽 ?

marianeum: Yeah I also like that it makes the distinction between click events/user interactions and other state changes clearer, rather than a few functions that are connected by the name

marianeum (approved):
thanks for the contribution!

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar2/data/CalendarRangeSelectionTests.kt
bvitaliyg: We can rename the methods to something like dispatchDayClicked(). Introducing Actions as a concept may not work with state machines in general, for instance – in some implementations Actions are used for another, SM-related comcepts.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -59,23 +59,34 @@ data class CalendarParams(
...
+
+    /**
+     * A whole month can be selected, but [Dates] could also be selected if desired.
+     *
+     * @param wholeMonthSelectionLabel the label for the whole month selection button.
+     */
+    data class Month(val wholeMonthSelectionLabel: String) : Rangeable
```
bvitaliyg, line 89: selectWholeMonthLabel?

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCells.kt
```diff
@@ -78,6 +78,7 @@ internal fun CalendarCells(
         monthsFormatter = params.monthsFormatter,
         weekFields = params.weekFields,
         selection = selection,
+        selectWholeMonthLabel = (params.selectionMode as? CalendarParams.SelectionMode.Month)?.wholeMonthSelectionLabel,
```
bvitaliyg, line 81: Like here :)

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarStateMachine.kt
```diff
@@ -75,16 +84,16 @@ internal fun CalendarState.dispatchClick(date: CalendarCell.Day): CalendarState
   if (date.inactive) return this
 
   val selection = when (params.selectionMode) {
-    CalendarParams.SelectionMode.Disabled -> selection
-    CalendarParams.SelectionMode.Single -> CalendarSelection.Single(date.date)
-    CalendarParams.SelectionMode.Range -> {
+    SelectionMode.Disabled -> selection
```
bvitaliyg, line 87: is? I tend to use is in sealed whens because the underlying entities might be converted from objects to classes in the future.

bvitaliyg (commented):
Just few minor things

Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarStateMachine.kt
ernestkamara: These changes existed before my changes. But, I will fix them now.

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar2/data/CalendarRangeSelectionTests.kt
bvitaliyg: Or, we can rename it to something else rather than action – may interaction? CalendarInteraction.DateClicked, CalendarInteraction.SelectMonthClicked?

PR #1076, opened 2022-06-07
Add Flare compose component

marianeum:

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/BpkFlare.kt
```diff
@@ -0,0 +1,59 @@
...
+  insetContent: Boolean = false,
+  contentAlignment: Alignment = Alignment.TopStart,
+  content: @Composable BoxScope.() -> Unit,
+) {
+  Box(
+    modifier = modifier
+      .clip(FlareShape(radius, pointerDirection))
+      .background(background),
+    propagateMinConstraints = true,
+    content = { FlareContent(pointerDirection, insetContent, contentAlignment, content) }
```
bvitaliyg, line 57: We use 2 boxes inside each other. Any reasons why we do this?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/BpkFlare.kt
```diff
@@ -0,0 +1,59 @@
...
+  background: Color = Color.Unspecified,
+  insetContent: Boolean = false,
+  contentAlignment: Alignment = Alignment.TopStart,
+  content: @Composable BoxScope.() -> Unit,
+) {
+  Box(
+    modifier = modifier
+      .clip(FlareShape(radius, pointerDirection))
+      .background(background),
+    propagateMinConstraints = true,
```
bvitaliyg, line 56: Let's add it to parameters as well.

bvitaliyg, line 56: Let's add it to parameters as well.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/BpkFlareImpl.kt
```diff
@@ -0,0 +1,153 @@
...
+) {
+  val flareWidth = scale * FlareVectorWidth
+  val flareX = (size.width - flareWidth) / 2
+  val flareY = when (pointerDirection) {
+    BpkFlarePointerDirection.Up -> flareHeight
+    BpkFlarePointerDirection.Down -> size.height - flareHeight
+  }
+  moveTo(flareX + 4.329f * scale, flareY)
+  cubicTo(
+    flareX + 14.663905f * scale, flareY + 0.409248008f * scale,
```
bvitaliyg, line 132: How did we get these constants?

bvitaliyg, line 132: How did we get these constants?

bvitaliyg (commented):
Just few questions

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/BpkFlareImpl.kt
bvitaliyg: And how iOS get it? :)

marianeum: That code long outlives the current set of iOS engineers, but my assumption would be they got an svg and mapped the svg to code. I tried to do that, but the flare isn't exportable from Figma for some reason. I also tried to convert the android vector to a compose vector and use that, but that contains a lot of relative position changes which aren't available in a simple Path, so this was the simpler approach and seems to match the current Flare

bvitaliyg: It may be also worth exposing it to SVG -> trying to use Vector2Compose to get it done. These shapes are in general generated by such tools.

Otherwise, it looks good, I'd just inline drawFlarePath() method. Also, please consider using 4 different shape object instead of using a new one.

bvitaliyg: It just describes the path and does not draw anything. Why do you think inlining this method decreases the performance?

bvitaliyg: FlareBottom, FlatBottomMediumRadius, FlatTop, FlareTopMediumRadius

marianeum: Not decreasing, but also not improving. It results in a warning
Screenshot 2022-06-08 at 16 31 09

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/FlareShape.kt
```diff
@@ -0,0 +1,111 @@
...
+
+private fun Path.addRect(
+  flareHeight: Float,
+  size: Size,
+  density: Density,
+  radius: BpkFlareRadius,
+  pointerDirection: BpkFlarePointerDirection,
+) {
+  val borderRect = when (pointerDirection) {
+    BpkFlarePointerDirection.Up -> Rect(0f, flareHeight, size.width, size.height + RectOffset)
```
marianeum, line 63: this was the easiest way to do this. add some more offset to the rect on the opposite side to avoid having to manually create the whole path

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/FlareShape.kt
```diff
@@ -0,0 +1,111 @@
...
+import net.skyscanner.backpack.compose.flare.BpkFlarePointerDirection
+import net.skyscanner.backpack.compose.flare.BpkFlareRadius
+import net.skyscanner.backpack.compose.tokens.BpkDimension
+
+internal class FlareShape(
+  private val radius: BpkFlareRadius,
+  private val pointerDirection: BpkFlarePointerDirection,
+) : Shape {
+
+  override fun createOutline(size: Size, layoutDirection: LayoutDirection, density: Density): Outline {
```
bvitaliyg, line 40: =

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/FlareShape.kt
```diff
@@ -0,0 +1,111 @@
...
+    flareX + 213.318f * scale, flareY + 3.345f * scale,
+    flareX + 223.396f * scale, flareY + 0.303f * scale,
+    flareX + 233.724f * scale, flareY + 0f * scale,
+  )
+  lineTo(flareX + 238f * scale, flareY + 0f * scale)
+}
+
+private const val FlareVectorHeight = 53
+private const val FlareVectorWidth = 234
+internal val FlareHeight = 11f.dp
```
bvitaliyg, line 110: Why internal?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/RoundedRectShape.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.geometry.toRect
+import androidx.compose.ui.graphics.Outline
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.LayoutDirection
+import net.skyscanner.backpack.compose.flare.BpkFlareRadius
+import net.skyscanner.backpack.compose.tokens.BpkDimension
+
+internal class RoundedRectShape(
```
bvitaliyg, line 32: We don't need to create a class, just using a function to create existing RoundedRectShape should be fine

bvitaliyg (commented):
Few suggestions

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flare/internal/FlareShape.kt
marianeum: it's also used for setting the content padding

bvitaliyg (approved):

PR #1094, opened 2022-06-17
Updated CONTRIBUTING.md

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,7 @@ Given that you have a compatible environment as stated above you can now set up 
 #### Snapshot testing
 Create an AVD using the following command
 
-> Note: Currently, screenshot testing doesn't work properly on M1 chips.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,8 +50,15 @@ Given that you have a compatible environment as stated above you can now set up
 #### Snapshot testing
 Create an AVD using the following command
 
+> Note: Currently, screenshot testing doesn't work properly on M1 chips.
```
marianeum, line 53: Maybe add something what to do in this case (eg generate screenshot tests on a different machine/check with another engineer if not available)

bvitaliyg: This solution should come as an output of the ticket I created :)

marianeum: yeah but in the meantime we need have the right instructions - particularly with m1s rolled out more widely soon

bvitaliyg: Done

marianeum (approved):

PR #1107, opened 2022-07-01
KOA-5208 Update chip component to match Figma

marianeum:
This also adds the new `onDark` style and `type` property for setting a trailing icon.

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/chip/BpkChip.kt
```diff
@@ -45,53 +45,72 @@ open class BpkChip @JvmOverloads constructor(
       this.isEnabled = !disabled
     }
 
-  private val style: BpkChipStyle
+  private var _style: BpkChipStyle
```
marianeum, line 48: renamed this as this is the internal representation

Backpack/src/main/java/net/skyscanner/backpack/chip/BpkChip.kt
```diff
@@ -45,53 +45,72 @@ open class BpkChip @JvmOverloads constructor(
...
         ?.mutate()
-        ?.apply { setBounds(0, 0, iconSize, iconSize) }
-        ?.also {
-          it.setTintList(textColors)
-          this.setCompoundDrawablesRelative(it, null, null, null)
+        ?.apply {
+          setBounds(0, 0, iconSize, iconSize)
+          setTintList(textColors)
         }
+      updateIcons()
```
marianeum, line 105: we now have two icon options, so had to update setting the drawable to a separate function

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipStyles.kt
```diff
@@ -103,14 +100,90 @@ internal sealed class BpkChipStyles : BpkChipStyle {
...
       )
 
-    init {
-      context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0)
-        .use {
-          backgroundColor = it.getColor(R.styleable.BpkChip_chipBackgroundColor, backgroundColor)
-          selectedBackgroundColor = it.getColor(R.styleable.BpkChip_chipSelectedBackgroundColor, selectedBackgroundColor)
-          disabledBackgroundColor = it.getColor(R.styleable.BpkChip_chipDisabledBackgroundColor, disabledBackgroundColor)
-          textColor = it.getColor(R.styleable.BpkChip_chipTextColor, textColor)
+    internal class Default(context: Context, typedArray: TypedArray?) : Solid(context) {
```
marianeum, line 103: this is split up into default & on dark now

Backpack/src/main/res/values/attrs.xml
```diff
@@ -125,6 +125,15 @@
   <declare-styleable name="BpkChip">
     <attr name="disabled" format="boolean" />
     <attr name="selected" format="boolean" />
+    <attr name="chipsStyle" format="enum">
+      <enum name="normal" value="0" />
```
marianeum, line 129: can't use default because it's a protected keyword

Backpack/src/main/res/values/attrs.xml
```diff
@@ -125,6 +125,15 @@
   <declare-styleable name="BpkChip">
     <attr name="disabled" format="boolean" />
     <attr name="selected" format="boolean" />
+    <attr name="chipsStyle" format="enum">
```
marianeum, line 128: can't use just style or chipStyle as it's used somewhere in android/material

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipStyles.kt
```diff
@@ -38,6 +41,7 @@ internal sealed class BpkChipStyles : BpkChipStyle {
     override var backgroundColor: Int = context.getColor(R.color.bpkBackgroundElevation03)
     override var textColor: Int = context.getColor(R.color.bpkTextPrimary)
     override var disabledBackgroundColor: Int = Color.TRANSPARENT
+    override val style = BpkChip.Style.Default // use default here to avoid breaking changes when `outline` is removed
```
marianeum, line 44: Because `BpkOutlineChip` exists as a separate components, rather than a style there's no nice way of introducing a style while outline still exists. This was the least disruptive way. It does mean that if somebody uses an outline chip and changes the style the style will be no longer outline, but given that outline is deprecated and should no longer be used this is acceptable, as using outline AND using a new property are not compatible.

Backpack/src/main/java/net/skyscanner/backpack/chip/BpkChip.kt
bvitaliyg: Renamed it to BpkChipAppearance

frugoman (approved):

PR #1110, opened 2022-07-06
Fix current month highlighting + whole month selection

sergiomarqmoura:
This PR tackles a few issues on the month selection feature:

### Current month highlighting not working when preselected

Currently, the `Calendar2Story` for [whole month selection](https://github.com/Skyscanner/backpack-android/blob/b7464a3b6b677ff1194b8f80ea6af8c38fe8c078/app/src/main/java/net/skyscanner/backpack/demo/stories/Calendar2Story.kt#L118) sets a `SelectionMode.Month` and then calls `setSelection(2019, January)`. This works (the calendar is correctly highlighted) because the current date is hardcoded as 1-Jan-2019 . If we change the current date to 2-Jan-2019 , it doesn't get highlighted anymore (unless we tap Select whole month again), which seems to be due to [this line](https://github.com/Skyscanner/backpack-android/blob/b7464a3b6b677ff1194b8f80ea6af8c38fe8c078/Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarStateMachine.kt#L124), that ignores a selection whenever its start date is not within the calendar params range. This condition makes total sense for a selection of `Dates` (`Single` or `Range`), but not for a `Month` (in which we would like to have the whole month highlighted - open for discussion).

This PR changes [this line](https://github.com/Skyscanner/backpack-android/blob/b7464a3b6b677ff1194b8f80ea6af8c38fe8c078/Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarStateMachine.kt#L119) to `Dates` instead, and creates a new case in the when {} for the `CalendarSelection.Month`, just to make it exhaustive, with no such restrictions.

### `CalendarParams.SelectionMode` doesn't support showing the "Select whole month" button + selecting a single date

Given that the only way to show the "Select whole month" button now is to set the `SelectionMode` as `SelectionMode.Month`, it means that whenever we want to have this button showing up, this has to be the selection mode we set. However, with the current behaviour, this will automatically mean that single dates are not supported, given that the `SelectionMode.Month` only allows users to input whole month + range selections. What we would really want is for these to not be exclusive - users may select whole month OR a range of dates OR a single date. 

This PR reverts the `SelectionMode` class to how it was before, and adding the `wholeMonthSelectionLabel` to the `CalendarParams` , with a default value of `""` , and in [CalendarCells.kt:L81](https://github.com/Skyscanner/backpack-android/blob/b7464a3b6b677ff1194b8f80ea6af8c38fe8c078/Backpack/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCells.kt#L81) replaces `selectWholeMonthLabel = (params.selectionMode as? CalendarParams.SelectionMode.Month)?.selectWholeMonthLabel` with `selectWholeMonthLabel = params.wholeMonthSelectionLabel` . It also helped reducing the complexity of this code, since there is now less inheritance (no sealed interfaces and classes in the `SelectionMode`) but instead the old `enum class SelectionMode` we had before.

+ [ ] Component `README.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -44,6 +44,7 @@ data class CalendarParams(
   val locale: Locale = Locale.getDefault(),
   val dayOfWeekText: TextStyle = TextStyle.NARROW,
   val now: LocalDate = LocalDate.now(),
+  val wholeMonthSelectionLabel: String = "",
```
marianeum, line 47: Maybe this could be nullable instead? That way it's more clear that this is optional, rather than relying on a less-explicit empty string

marianeum (approved):
much simpler, thanks! just a small suggestion on nullability. and just to clarify: the issue for the select whole month option for a month with disabled dates is going to be a separate PR?

ernestkamara (approved):
LGTM!

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarCellHeaderHolder.kt
```diff
@@ -48,8 +49,17 @@ internal class CalendarCellHeaderHolder(
...
     btnSelectWholeMonth.apply {
-      model.selectWholeMonthLabel?.let { text = it }
-      isVisible = !model.selectWholeMonthLabel.isNullOrEmpty()
+      when (model.monthSelectionMode) {
+        CalendarParams.MonthSelectionMode.Disabled -> {
+          isEnabled = false
+          isVisible = false
+        }
+        is CalendarParams.MonthSelectionMode.SelectWholeMonth -> {
+          isEnabled = true
```
bvitaliyg, line 58: I think there's no need to disable it if we set it to View.GONE. But it needs to be disabled when calendar is disabled and MonthSelectionMode == SelectWholeMonth.

bvitaliyg (changes requested):
Few things related to the disabled logic.

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarCellHeaderHolder.kt
sergiomarqmoura: Well spotted! Just added [another commit](https://github.com/Skyscanner/backpack-android/pull/1110/commits/2a3c5983d60f31dab0c19b1415c7cca2a1724aee) fixing this

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarCellHeaderHolder.kt
```diff
@@ -48,8 +49,16 @@ internal class CalendarCellHeaderHolder(
   override fun bind(model: CalendarCell.Header) {
     month.text = model.title
     btnSelectWholeMonth.apply {
-      model.selectWholeMonthLabel?.let { text = it }
-      isVisible = !model.selectWholeMonthLabel.isNullOrEmpty()
+      when {
```
bvitaliyg, line 52: isVisible = monthSelectionMode is CalendarParams.MonthSelectionMode.SelectWholeMonth
isEnabled = selectionMode !is CalendarParams.SelectionMode.Disabled

bvitaliyg (changes requested):
Just one small request

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarCellHeaderHolder.kt
sergiomarqmoura: Applied 👍

Backpack/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -65,22 +66,24 @@ data class CalendarParams(
...
+    object Disabled : MonthSelectionMode
 
     /**
-     * A whole month can be selected, but [Dates] could also be selected if desired.
-     *
-     * @param selectWholeMonthLabel the label for the whole month selection button.
+     * Only an entire month can be selected, by tapping on the [label] next to its name.
      */
-    data class Month(val selectWholeMonthLabel: String) : Rangeable
+    data class SelectWholeMonth(val label: String) : MonthSelectionMode
```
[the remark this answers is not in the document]
marianeum: I wouldn't do that as that stops us from adding a different type of month selection mode in the future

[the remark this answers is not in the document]
sergiomarqmoura: Agree with @marianeum , and I think we discussed this before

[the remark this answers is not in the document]
bvitaliyg: Maybe "single" then, perhaps? I don't see any other was month selection logic can be extended in the future.

PR #1114, opened 2022-07-07
KOA-5414 Provide list of tokens

marianeum:
Create list for tokens in demo app, so we can easily add a demo for this.

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/data/Tokens.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/data/Tokens.kt
@@ -1,0 +1,52 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.skyscanner.backpack.demo.data
+
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Dp
+import net.skyscanner.backpack.compose.tokens.BpkBorderRadius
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/skyscanner/backpack/demo/data/Tokens.kt
```diff
@@ -0,0 +1,52 @@
...
+import net.skyscanner.backpack.compose.tokens.BpkBorderRadius
+import net.skyscanner.backpack.compose.tokens.BpkColor
+import net.skyscanner.backpack.compose.tokens.BpkColors
+import net.skyscanner.backpack.compose.tokens.BpkElevation
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+import kotlin.reflect.KProperty1
+import kotlin.reflect.full.memberProperties
+import kotlin.reflect.full.starProjectedType
+
+val BpkSpacing.values: List<Token<Dp>>
```
marianeum, line 32: these mainly exist, so we don't have to provide the generics when using them

app/src/main/java/net/skyscanner/backpack/demo/data/Tokens.kt
```diff
@@ -0,0 +1,52 @@
...
+  get() = BpkBorderRadius.values()
+
+val BpkElevation.values: List<Token<Dp>>
+  get() = BpkElevation.values()
+
+val BpkColor.values: List<Token<Color>>
+  get() = BpkColor.values()
+
+val BpkColors.values: List<Token<Color>>
+  get() = BpkColors::class.memberProperties.mapNotNull {
```
marianeum, line 45: colour is a bit of a special case as we also have other functions here. i tried adding the return type check to the generic function as it might be helpful for future proofing anyways, but the check doesn't work with generics and i couldn't find a generic way to do it

build.gradle
```diff
@@ -95,6 +95,7 @@ ext {
   threeTenAbpAndroid = "com.jakewharton.threetenabp:threetenabp:$threeTenAbpVersion"
   picasso = "com.squareup.picasso:picasso:2.71828"
   ktx = "androidx.core:core-ktx:1.8.0"
+  kotlinReflection = "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
```
bvitaliyg, line 98: Any way we can do this without Kotlin reflection?

bvitaliyg (commented):
A question about Kotlin reflection

build.gradle
marianeum: It doesn't really work as Dp is an inline class and java reflection can't really work with that and treats it as a float

bvitaliyg (approved):

PR #1117, opened 2022-07-14
Migrated Calendar2 to single-threaded model

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


marianeum (approved):
have you double checked that the crash no longer happens? you can add selection + scrollToDate in the demo app. worth finding a date that consistently crashes before & no longer after the change to verify

PR #1127, opened 2022-07-25
KOA-5405 Add Compose elevation screen to the demo app

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/compose/ElevationStory.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/compose/ElevationStory.kt
@@ -31,6 +31,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.shadow
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.max
 import net.skyscanner.backpack.compose.text.BpkText
 import net.skyscanner.backpack.compose.tokens.BpkBorderRadius
 import net.skyscanner.backpack.compose.tokens.BpkColor
@@ -69,7 +70,10 @@ private fun ElevationCard(token: Token<Dp>) {
 
 @Composable
 private fun ElevationSample(token: Token<Dp>) {
-  Box(modifier = Modifier.padding(BpkSpacing.Base)) {
+  Box(
+    modifier = Modifier
+      .padding(horizontal = max(BpkSpacing.Base, token.value), vertical = max(BpkSpacing.Base, token.value))
+  ) {
     ElevationCard(token)
   }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/skyscanner/backpack/demo/compose/ElevationStory.kt
```diff
@@ -0,0 +1,73 @@
...
+        text = "${token.name} = ${token.value}",
+        modifier = Modifier.padding(BpkSpacing.Base),
+      )
+    }
+  }
+}
+
+@Composable
+private fun ElevationSample(token: Token<Dp>) {
+  Box(Modifier.padding(horizontal = max(BpkSpacing.Base, token.value), vertical = max(BpkSpacing.Base, token.value))) {
```
marianeum, line 70: as both the horizontal & vertical padding are the same you can also use `Modifier.padding(max(BpkSpacing.Base, token.value))` and it'll be applied on all 4 sides

app/src/main/java/net/skyscanner/backpack/demo/compose/ElevationStory.kt
```diff
@@ -0,0 +1,73 @@
...
+@Composable
+private fun ElevationCard(token: Token<Dp>) {
+  Surface(
+    modifier = Modifier.fillMaxWidth(),
+    shape = RoundedCornerShape(BpkBorderRadius.Md),
+    elevation = token.value,
+  ) {
+    Box(contentAlignment = Alignment.Center) {
+      BpkText(
+        text = "${token.name} = ${token.value}",
```
marianeum, line 61: it is generally good practise to use string resources for any kind of string manipulation. i've also noticed that the same format is used for other tokens, so you can reuse the same string resource format for all of these and it'll be consistent. This is also how Android does localisation - while we currently don't support different languages for the demo app it is generally good practise to use them in case you need support later on.

This is how you use string resources in compose: https://developer.android.com/jetpack/compose/resources - particularly the second example with positional formatting is relevant here.
This is the xml part - https://developer.android.com/guide/topics/resources/string-resource#formatting-strings

Let me know if you have any questions!

app/src/main/java/net/skyscanner/backpack/demo/compose/ElevationStory.kt
jaanainkinen: Thanks, changed this!

app/src/main/java/net/skyscanner/backpack/demo/compose/ElevationStory.kt
jaanainkinen: Thanks, I applied this in both elevation and spacing stories!

marianeum (approved):

bvitaliyg (approved):

PR #1128, opened 2022-07-25
KOA-4934 Added BpkTextField and BpkFieldSet for Compose

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


app/src/main/java/net/skyscanner/backpack/demo/compose/FieldStatusSwitcher.kt
```diff
@@ -0,0 +1,72 @@
...
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.button.BpkButton
+import net.skyscanner.backpack.compose.button.BpkButtonType
+import net.skyscanner.backpack.compose.fieldset.BpkFieldStatus
+
+@Composable
+internal fun FieldStatusSwitcher(
+  modifier: Modifier = Modifier,
+  initialStatus: BpkFieldStatus = BpkFieldStatus.Default,
+  errorText: String = "Error text",
```
marianeum, line 41: can the demo strings be string resources?

app/src/main/java/net/skyscanner/backpack/demo/compose/TextFieldStory.kt
```diff
@@ -0,0 +1,104 @@
...
+
+  FieldStatusSwitcher(
+    verticalArrangement = Arrangement.spacedBy(BpkSpacing.Base),
+    modifier = Modifier
+      .verticalScroll(rememberScrollState())
+      .padding(BpkSpacing.Base),
+  ) { status ->
+    CompositionLocalProvider(LocalTextStyle provides BpkTheme.typography.label2) {
+
+      BpkText(text = "Default")
```
marianeum, line 55: can you split the individual samples into separate functions with a preview annotation?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/internal/Colors.kt
```diff
@@ -38,7 +39,7 @@ internal fun BpkButtonType.disabledBackgroundColor(): Color =
 internal fun BpkButtonType.disabledContentColor(): Color =
   when (this) {
     BpkButtonType.LinkOnDark -> Color.White.copy(alpha = 0.2f)
-    else -> dynamicColorOf(BpkColor.SkyGrayTint04, BpkColor.BlackTint03)
+    else -> BpkTheme.colors.textDisabled
```
marianeum, line 42: this is not the same as other disabled colours (intentionally)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/radiobutton/BpkRadioButton.kt
```diff
@@ -122,7 +121,7 @@ private fun BpkRadioButtonImpl(
       colors = RadioButtonDefaults.colors(
         selectedColor = BpkTheme.colors.primary,
         unselectedColor = BpkTheme.colors.textSecondary,
-        disabledColor = BpkColor.SkyGrayTint04,
+        disabledColor = BpkTheme.colors.textDisabled,
```
marianeum, line 124: is that intended? that's the radiobutton colour that doesn't seem to change colour in night mode based on figma

app/src/main/java/net/skyscanner/backpack/demo/compose/FieldSetStory.kt
```diff
@@ -0,0 +1,73 @@
...
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+
+@Composable
+fun FieldSetStory() {
+  FieldStatusSwitcher(
+    modifier = Modifier.padding(BpkSpacing.Base),
+    verticalArrangement = Arrangement.spacedBy(BpkSpacing.Base),
+  ) { status ->
+    BpkFieldSet(
+      label = "With description",
```
marianeum, line 40: can you use string resources for all the demo strings? particularly as a bunch are duplicated across different usages & files

app/src/main/java/net/skyscanner/backpack/demo/compose/FieldStatusSwitcher.kt
marianeum, line 41: still a few hardcoded strings in this file

app/src/main/res/values/strings.xml
```diff
@@ -138,4 +138,15 @@
   <string name="chip_option">Option</string>
   <string name="chip_select">Select</string>
   <string name="chip_dismiss">Dismiss</string>
+  <string name="default_text">Default</string>
```
marianeum, line 141: maybe prefix these with `input`?

docs/compose/FieldSet/README.md
```diff
@@ -0,0 +1,43 @@
...
+## Installation
+
+Backpack Compose is available through [Maven Central](https://search.maven.org/artifact/net.skyscanner.backpack/backpack-compose). Check the main [Readme](https://github.com/skyscanner/backpack-android#installation) for a complete installation guide.
+
+## Usage
+
+FieldSet is a component which wraps its content (user input or form element) and adds optional title, description and
+error labels around it depending on the field status.
+
+The statuses available is Default, Disabled, Error and Validated. The statuses automatically dispatched to
```
marianeum, line 12: are, rather than is, dispatch rather than dispatched

docs/compose/FieldSet/README.md
```diff
@@ -0,0 +1,43 @@
...
+Example of a FieldSet with Validated status:
+
+```Kotlin
+import net.skyscanner.backpack.compose.fieldset.BpkFieldSet
+import net.skyscanner.backpack.compose.fieldset.BpkFieldStatus
+
+BpkFieldSet(
+  status = BpkFieldStatus.Validated,
+  label = "Title (optional)",
+  description = "Description (optional",
```
marianeum, line 24: closing bracket

docs/compose/FieldSet/README.md
```diff
@@ -0,0 +1,43 @@
...
+Example of a FieldSet with Error status:
+
+```Kotlin
+import net.skyscanner.backpack.compose.fieldset.BpkFieldSet
+import net.skyscanner.backpack.compose.fieldset.BpkFieldStatus
+
+BpkFieldSet(
+  status = BpkFieldStatus.Error("Error text"),
+  label = "Title (optional)",
+  description = "Description (optional",
```
marianeum, line 39: again

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -80,6 +80,7 @@ object DocsRegistry {
     ComposeScreenshot("Dialog - Compose", "warning") { setupComposeDialog(it, ShownDialog.Warning) },
     ComposeScreenshot("Dialog - Compose", "destructive") { setupComposeDialog(it, ShownDialog.Destructive) },
     ComposeScreenshot("Dialog - Compose", "flare") { setupComposeDialog(it, ShownDialog.Flare) },
+    ComposeScreenshot("FieldSet", "default"),
```
marianeum, line 83: can you also add docs screenshots for the other states here?

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -115,8 +116,9 @@ object DocsRegistry {
     ComposeScreenshot("Text - Compose - Hero", "hero"),
     ComposeScreenshot("Text - Compose - Heading", "heading"),
     ComposeScreenshot("Text - Compose - Body", "body"),
-    ViewScreenshot("Text Field - Default", "default"),
-    ViewScreenshot("Text Field - With labels", "labels"),
+    ViewScreenshot("Text Field - View - Default", "default"),
+    ViewScreenshot("Text Field - View - With labels", "labels"),
+    ComposeScreenshot("Text Field - Compose", "default"),
```
marianeum, line 121: same here

app/src/main/res/values/strings.xml
bvitaliyg: We already have "Default" in strings.xml – in fact, in 2 places. Let's make it generic and reusable.

app/src/androidTest/java/net/skyscanner/backpack/compose/fieldset/BpkFieldSetTest.kt
```diff
@@ -0,0 +1,162 @@
...
+class BpkFieldSetTest : BpkSnapshotTest() {
+
+  @Before
+  fun setup() {
+    setDimensions(height = 50, width = 200)
+  }
+
+  @Test
+  fun default() = composed {
+    BpkFieldSet(
```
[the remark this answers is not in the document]
marianeum: we're not testing automatically passing through the status anywhere then, so that might break without us noticing

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -125,6 +125,14 @@ class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
       webName = "text",
       componentsToReplace = setOf("androidx.compose.material.Text")
     ),
+    TEXT_FIELD(
+      fullName = "net.skyscanner.backpack.compose.textfield.BpkTextField",
+      webName = "text-input",
+      componentsToReplace = setOf(
+        "androidx.compose.foundation.text.BasicTextField",
```
marianeum, line 132: probably also good to add the other variants here - particularly outline as that is very similar to our component

docs/compose/FieldSet/README.md
bvitaliyg: being dispatched, perhaps?

marianeum: actually: The statuses are automatically dispatched to

app/src/main/res/values/strings.xml
marianeum: duplication in strings files isn't necessarily a bad thing as they might be used in different cases. if something is generic i'd mark it as generic and prefix it that way (and remove the duplicates at the same time), rather than implicitly making all of the input strings generic

bvitaliyg: In this case, we're working with the demo app. Using XML strings create some complexity itself. Considering that we don't get a lot of value from this (we're using this as an example of proper locale handling) I'd rather not increase the complexity further.

marianeum: let's keep it simple and rename them then. just because it's a demo app doesn't mean we shouldn't follow good practices

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
bvitaliyg: I'm not sure we actually benefit from having all these screenshots on the website.

marianeum: both designers & engineers use the website to see what the components support in code - the screenshots are particularly useful for design. a lot of the functionality isn't shown on the screenshots, and while you can see that there is some support for error, validation etc you can't see what it looks like/what's actually supported

app/src/main/res/values/strings.xml
bvitaliyg: We are following good practices. I'm just saying we need to reuse the strings when possible to not increase the complexity of the strings management. We do this for a lot of other components, the only reason why we can't do this is that 'default' is a reserved Java keyword – hence the suffix.

app/src/androidTest/java/net/skyscanner/backpack/compose/fieldset/BpkFieldSetTest.kt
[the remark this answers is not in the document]
bvitaliyg: Ok updated.

app/src/main/res/values/strings.xml
marianeum: all our strings are prefixed with something right now - again: if they should be generic, let's also indicate that they are and update other usages. otherwise we're just adding another version of the same string with no benefit

bvitaliyg: We can a generic prefix, yes. Which would you prefer?

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
bvitaliyg: Added it!

marianeum (approved):

PR #1169, opened 2022-08-23
Remove new base colours

marianeum:
Removed the new base colours following the foundations update

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkColors.kt
```diff
@@ -57,50 +57,48 @@ public class BpkColors private constructor(
...
-      textDisabled: Color = BpkColor.BlackAlpha20,
-      textError: Color = BpkColor.Berry,
-      textLink: Color = BpkColor.SkyBlue,
-      textPrimary: Color = BpkColor.SkyGray,
-      textPrimaryOnContrast: Color = BpkColor.White,
-      textQuaternary: Color = BpkColor.SkyGrayTint03,
-      textSecondary: Color = BpkColor.SkyGrayTint02,
-      textSecondaryOnContrast: Color = BpkColor.WhiteAlpha80,
-      textTertiary: Color = BpkColor.SkyGrayTint03,
+      background: Color = Color(0xFFFFFFFF),
```
marianeum, line 67: I originally wanted to try and maintain the referred tokens that still exist, however this is actually quite tricky to implement without changing the current token generation quite significantly. It would require keeping a list of all the tokens, as well as the filtered list (or filtering at a later stage) to check if a token exists. As eventually we should no longer have base colours (in a happy future far away when all deprecated tokens have been migrated) this is acceptable i think

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkColors.kt
```diff
@@ -138,23 +139,26 @@ private val ColorClass = ClassName("androidx.compose.ui.graphics", "Color")
...
   namespace: String,
 ): TypeSpec {
 
-  fun BpkColorModel.toProperty() : PropertySpec {
+  fun BpkColorModel.toProperty(): PropertySpec {
 
-    fun String.toHexColor() =
-      "0xFF${uppercase()}"
+    fun String.toComposeStaticName() =
+      CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.UPPER_CAMEL, removePrefix("MARCOMMS_"))
```
olliecurtis, line 157: Does this mean we are making the Marcomms colours available in Android? As these should not be available in code as its for marketing material only

olliecurtis (approved):

PR #1210, opened 2022-09-09
Converted the subcomponent list to Compose

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests


app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
```diff
@@ -1,89 +0,0 @@
...
-    intent.putExtra(ComponentDetailFragment.ARG_ITEM_ID, viewId)
-
-    context.startActivity(intent)
-  }
-
-  override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
-    return if (viewType == 0) {
-      ViewHolder(
-        LayoutInflater.from(parent.context)
-          .inflate(R.layout.component_list_header, parent, false)
```
marianeum, line 47: I think we should be able to delete this one

app/src/main/java/net/skyscanner/backpack/demo/StoryItemDecoration.kt
```diff
@@ -1,69 +0,0 @@
...
-import android.graphics.Rect
-import android.view.View
-import androidx.appcompat.content.res.AppCompatResources
-import androidx.recyclerview.widget.RecyclerView
-import net.skyscanner.backpack.demo.StoriesRecyclerViewAdapter.Companion.STORY_VIEW_TYPE
-import kotlin.math.roundToInt
-
-class StoryItemDecoration(context: Context) : RecyclerView.ItemDecoration() {
-
-  private val divider = AppCompatResources.getDrawable(context, R.drawable.item_divider)!!
```
marianeum, line 32: and this one

app/src/main/java/net/skyscanner/backpack/demo/stories/SubStory.kt
```diff
@@ -36,14 +37,17 @@ open class SubStory : Story() {
   ): View? {
     val stories = arguments?.getStringArray(STORIES) ?: savedInstanceState?.getStringArray(STORIES)
     if (stories != null) {
-      val view = inflater.inflate(R.layout.component_list, container, false)
```
marianeum, line 39: I think the only usage left of `component_list` is in the `ComponentRegistry` as an empty placeholder for the "sneak peek" item. It might be good to actually remove the view tab for that, so we can remove the layout and then we can also delete `component_list_content`

app/src/main/java/net/skyscanner/backpack/demo/StoriesRecyclerViewAdapter.kt
jaanainkinen: Thanks, done!

marianeum (approved):
nice one!

PR #1229, opened 2022-09-20
KOA-5578 Update fab to semantic tokens

marianeum:
The disabled colour will need to be updated when the button disabled colours are available

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/fab/BpkFab.kt
```diff
@@ -59,14 +59,16 @@ open class BpkFab @JvmOverloads constructor(
...
+      context.getColor(R.color.bpkTextDisabled)
+    )
     this.isClickable = isEnabled
     this.stateListAnimator = AnimatorInflater.loadStateListAnimator(context, R.drawable.bpk_button_state_animator)
     this.elevation = resources.getDimensionPixelSize(R.dimen.bpkElevationBase).toFloat()
     this.backgroundTintList = getColorSelector(
       backgroundColour,
-      backgroundColour,
-      context.getColor(R.color.bpkSkyGrayTint06)
+      context.getColor(R.color.__buttonDisabledBackground)
```
marianeum, line 71: once the new button tokens are available we can use the new colours here too

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -298,14 +298,14 @@ object ComponentRegistry {
         TAB_TITLE_COMPOSE composeStory { FlareStory() },
       ),
     ),
+    "Floating Action Button" story NodeData { Story of R.layout.fragment_fab },
```
marianeum, line 301: this was in the wrong position alphabetically

olliecurtis (approved):
LGTM! One concern is are teams going to be aware this big change of colour to the FAB before we release this as looking at the screenshots this is a significant change colour wise from green to blue?

marianeum:
Yep - I had a look at current FAB usages and it looks like almost all of them set a custom background colour - when making the update I'll also apply a custom foreground colour, as we now switch between day/night mode where this wasn't done yet.

PR #1295, opened 2022-10-31
Added BpkNudger for Compose

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


bvitaliyg:
Record snapshots

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,91 @@
...
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+import net.skyscanner.backpack.compose.tokens.Minus
+import net.skyscanner.backpack.compose.tokens.Plus
+
+@Composable
+fun BpkNudger(
+  value: Int,
+  onValueChange: (Int) -> Unit,
+  range: IntRange,
```
marianeum, line 45: As we've got a separate min & max value property on all platforms let's stay consistent with that

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
```diff
@@ -0,0 +1,91 @@
...
+  val trimmedValue = value.coerceIn(range)
+
+  Row(
+    modifier = modifier,
+    verticalAlignment = Alignment.CenterVertically,
+  ) {
+
+    BpkButton(
+      icon = BpkIcon.Minus,
+      contentDescription = "-",
```
marianeum, line 59: For the view version we did this to cheat and add accessibility without making a breaking change in the API. Now that we're starting from fresh let's add a parameter for this. ie web has a `decreaseButtonLabel` and `increaseButtonLabel`
Alternatively I wonder if we can re-use the progress semantics on the whole component to allow increasing it from the slider. but it might be that that only supports percentages, rather than absolute values (similar to the view version)

We also need to announce the state when the value is increased/decreased

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
bvitaliyg: We can, although I think we need to align to the developer experience in the first place and have the range in simply more convenient.

marianeum: Alignment across platforms (AND with Figma) also improves developer platform - that's why we're now starting to spec out the API first in Figma before working on the component to ensure we're all aligned.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
marianeum: +/- is still a bit of a hack and I'm not sure if it works great in all languages either. 

Slider is one option (that's what iOS uses) - the other is simply announcing the state after changing it which the current nudger component does as well. We definitely need to announce _something_, though.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
bvitaliyg: Sliders use Ranges as well. Since most of our widgets will be used by Android (not cross-platform) developers, having API aligned with Compose and BpkSlider seems a better option for me

marianeum: We haven't created slider yet, so what we decide there we can align with here as well. The alignment is something we agreed on as a team, so if there's deviation we should discuss it as a team.

bvitaliyg: It has the same semantic meaning – min and max values, it just handles it in a different way. We still follow the same specs. Deviating for the platform-specific API is fine as long as we keep the same meaning from Figma – and we do that with arguments such as Modifier, InteractionSource, state hoisting etc.

docs/compose/Nudger/README.md
```diff
@@ -0,0 +1,29 @@
...
+[![Source code](https://img.shields.io/badge/Source%20code-GitHub-lightgrey)](https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger)
+
+## Default
+
+| Day                                                                                                                             | Night                                                                                                                                          |
+|---------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------|
+| ![Nudger component](https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/Nudger/screenshots/all.png) | ![Nudger component - dark mode](https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/Nudger/screenshots/all_dm.png) |
```
marianeum, line 11: can you please update this to the new format that's used on main? otherwise the screenshots will be too large

marianeum (approved):

PR #1300, opened 2022-11-02
Added BpkRating for Compose

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


bvitaliyg:
Record snapshots

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/rating/BpkRating.kt
```diff
@@ -0,0 +1,79 @@
...
+      overflow = TextOverflow.Ellipsis,
+    )
+  }
+}
+
+@Composable
+fun BpkRating(
+  value: Float,
+  modifier: Modifier = Modifier,
+  scale: BpkRatingScale? = BpkRatingScale.ZeroToFive,
```
marianeum, line 65: Showing the scale is different from the supported scale. For example the trip advisor example doesn't show the scale, but the logo still has a 0-5 scale, therefore we should clamp it to that. So the `showScale` property still makes sense and making this non-nullable.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/rating/internal/BpkRatingComponents.kt
```diff
@@ -0,0 +1,136 @@
...
+    style = BpkTheme.typography.bodyDefault,
+    color = BpkTheme.colors.textSecondary,
+    maxLines = 1,
+    overflow = TextOverflow.Ellipsis,
+  )
+}
+
+private fun formatValue(value: Float, scale: BpkRatingScale?, format: DecimalFormat): String {
+  val coerced = if (scale != null) value.coerceIn(scale.range) else value
+  return format.format(coerced)
```
marianeum, line 129: how is this rounding? i remember the existing component had a specific rounding we had to use

bvitaliyg: https://github.com/Skyscanner/backpack-android/blob/936cfcb629c53317c407bcd626581bdab6c80c38/Backpack/src/main/java/net/skyscanner/backpack/rating/internal/RatingScore.kt#L38

This is what we use for rounding. Added it to the code

bvitaliyg:
Record snapshots

PR #1301, opened 2022-11-03
Convert final files to Kotlin

marianeum:

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/toast/BpkToast.kt
```diff
@@ -0,0 +1,48 @@
...
+import net.skyscanner.backpack.text.BpkFontSpan
+import net.skyscanner.backpack.text.BpkText
+import net.skyscanner.backpack.text.BpkText.Companion.getFont
+import net.skyscanner.backpack.text.BpkText.FontDefinition
+
+object BpkToast {
+  const val LENGTH_SHORT = Toast.LENGTH_SHORT
+  const val LENGTH_LONG = Toast.LENGTH_LONG
+
+  @JvmStatic
```
marianeum, line 35: This makes sure people can access it the same way they could before

android-configuration.gradle
```diff
@@ -41,6 +41,10 @@ android {
     warningsAsErrors true
     disable 'UnusedResources' // we're exposing resources for consumers
   }
+
+  buildFeatures {
```
marianeum, line 45: This stops generating a `buildConfig` file, which was also generated in java and we don't need it.

backpack-compose/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ apply(from = "${rootProject.projectDir}/android-configuration.gradle")
 android {
   buildFeatures {
     compose = true
+    resValues = false
```
marianeum, line 34: no resources in the compose module either

Backpack/src/main/java/net/skyscanner/backpack/toast/BpkToast.kt
frugoman: is it worth deprecating this api and establishing a new kotlin friendly one? not blocking for this pr necesarily tho

frugoman (approved):
🚢

Backpack/src/main/java/net/skyscanner/backpack/toast/BpkToast.kt
marianeum: so what this does is make sure that java can also still access it correctly without it being very awkward. doesn't mean it's _not_ kotlin friendly.

PR #1308, opened 2022-11-04
Create script for setting up new components

marianeum:
Created a script to setup the basic parts required for creating new components (component, demo app sample, test, readme).

Sample output/steps:
```
Enter the name of the component:
BpkSpaceShip
Enter the package name of the component, or leave empty for spaceship:
space

Create compose component structure for space.BpkSpaceShip?
[Y/n]: y

Creating compose component structure...

Created BpkSpaceShip.kt
Created SpaceShipStory.kt
Created BpkSpaceShipTest.kt
Created README.md

To include the component in the demo app add the following line to the ComponentRegistry.kt COMPONENT list in alphabetic position:
"SpaceShip" composeStory { SpaceShipStory() }

To add it to the screenshot generation add the following line to the DocsRegistry.kt screenshots list in alphabetic position:
ComposeScreenshot("SpaceShip", "default")

Note: If the component already exists in the view system the guide above will need to be adopted slightly - see CONTRIBUTING.md for details
```

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,15 @@ Given that you have a compatible environment as stated above you can now set up 
 +  Open the project in Android Studio
 + If you are a Skyscanner employee, search the internal documentation for _"Guide – Setup Internal Backpack Android Builds"_ and follow the instructions.
 
+## Creating components
+
+To create a new component in Compose you can use the following script to setup the most common required parts of a component:
+
+```
+python3 scripts/generate-component.py NameOfComponent
+```
+
+Note: This currently is not supported for creating components using the view system.
 
 ## Testing
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,10 +46,15 @@ Given that you have a compatible environment as stated above you can now set up 
 
 ## Creating components
 
-To create a new component in Compose you can use the following script to setup the most common required parts of a component:
+To create a new component in Compose you can use the following script to setup the most common required parts of a component.
 
+To run it execute:
 ```
-python3 scripts/generate-component.py NameOfComponent
+./scripts/generate-component.py
+```
+and follow instructions, or provide arguments directly:
+```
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -44,6 +44,15 @@ Given that you have a compatible environment as stated above you can now set up
 +  Open the project in Android Studio
 + If you are a Skyscanner employee, search the internal documentation for _"Guide – Setup Internal Backpack Android Builds"_ and follow the instructions.
 
+## Creating components
+
+To create a new component in Compose you can use the following script to setup the most common required parts of a component:
+
+```
+python3 scripts/generate-component.py NameOfComponent
```
bvitaliyg, line 52: I think we need to do this as a Gradle task. Having 3rd building framework (in addition to Gradle and Node) seems flaky. Besides, I believe Python needs its own environment (sooner or later), and we want the contribution and setup to be as easy as possible.

marianeum: I originally started off writing this using our current token generation system and got something working, but it's not ideal: it would include all the `public` / `Unit` modifiers/return values, so you'd either have this polluted in the code base or need to remove it manually. kotlinpoets is for generated files that don't need to be touched manually.

The problem with gradle is it's not really built for things like this - it's a build system. So even just providing an input to it is not as simple as in python or other scripting languages (both in development terms & usage terms).

Output using a gradle task also isn't that obvious (as you know from the ktlint output), so it'll be harder to use/figure out what manual steps to take and what actually happened in the script.

I also think this script needs to be extended to take step by step inputs from users (with defaults) & confirmation steps. For example to make it clear what format the component name should be in, giving users options to change the default package name/component name (for example for something like fab where you might want to use a shorter name in places). That isn't possible with a gradle task either.

So I definitely understand where you're coming from, but also we need to use the right tool for the right task and Gradle isn't the right tool for this.

frugoman (approved):

PR #1309, opened 2022-11-05
Added BpkCalendar for Compose

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


.github/workflows/_build.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           force-avd-creation: false
           emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: true
-          script: ./gradlew :Backpack:connected${{ env.config }}AndroidTest :backpack-compose:connected${{ env.config }}AndroidTest
+          script: ./gradlew :Backpack:connected${{ env.config }}AndroidTest :backpack-compose:connected${{ env.config }}AndroidTest :backpack-common:connected${{ env.config }}AndroidTest
```
bvitaliyg, line 125: We extracted business logic and its tests to common

Backpack/build.gradle
```diff
@@ -30,9 +30,8 @@ apply from: "$rootProject.projectDir/android-configuration.gradle"
 android {
   namespace 'net.skyscanner.backpack'
   kotlinOptions {
-    freeCompilerArgs += [
-      "-Xuse-experimental=net.skyscanner.backpack.util.ExperimentalBackpackApi",
-    ]
+    freeCompilerArgs += "-Xopt-in=net.skyscanner.backpack.util.ExperimentalBackpackApi"
+    freeCompilerArgs += "-Xopt-in=net.skyscanner.backpack.util.InternalBackpackApi"
```
bvitaliyg, line 34: No opt-ins for InternalBackpackApi for using Backpack API internally

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarCellHeaderHolder.kt
```diff
@@ -51,9 +51,7 @@ internal class CalendarCellHeaderHolder(
     btnSelectWholeMonth.apply {
       isVisible = model.monthSelectionMode is CalendarParams.MonthSelectionMode.SelectWholeMonth
       isEnabled = model.calendarSelectionMode !is CalendarParams.SelectionMode.Disabled
-      if (model.monthSelectionMode is CalendarParams.MonthSelectionMode.SelectWholeMonth) {
-        text = model.monthSelectionMode.label
-      }
+      text = (model.monthSelectionMode as? CalendarParams.MonthSelectionMode.SelectWholeMonth)?.label ?: ""
```
bvitaliyg, line 54: No smart cast is possible since the code lives in another module

app/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTestCases.kt
```diff
@@ -0,0 +1,191 @@
...
+
+import net.skyscanner.backpack.calendar2.extension.toIterable
+import org.threeten.bp.DayOfWeek
+import org.threeten.bp.LocalDate
+import org.threeten.bp.Month
+import org.threeten.bp.Period
+import org.threeten.bp.YearMonth
+import java.util.Locale
+
+object BpkCalendarTestCases {
```
bvitaliyg, line 29: Common test cases to be used both in Compose/View calendars.

app/src/androidTest/java/net/skyscanner/backpack/compose/calendar/BpkCalendarTest.kt
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  @Test
+  fun screenshotTestCalendarWithSameStartAndEndDateSelected() {
+    val controller = createController(BpkCalendarTestCases.Params.WithSameStartAndEndDateSelected)
+
+    snap(controller) {
+
+      it.onAllNodesWithText("17")
+        .onFirst()
```
bvitaliyg, line 145: onFirst=Jan, onLast=Fab. No methods in Compose to click on item with index like in View

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -194,6 +200,28 @@ private fun setupWholeMonthCalendar() {
     }
 }
 
+private fun setupCalendar2(rule: AndroidComposeTestRule<*, *>) {
```
bvitaliyg, line 203: Again, since we cannot select things programmatically in Compose from this point in code, we simulate user interaction

app/src/main/java/net/skyscanner/backpack/demo/data/CalendarStoryType.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.threeten.bp.LocalDate
+import org.threeten.bp.Month
+import org.threeten.bp.Period
+import org.threeten.bp.YearMonth
+import kotlin.math.roundToInt
+
+private val now = LocalDate.of(2019, 1, 1)
+private val range = now..(now + Period.ofYears(2))
+
+enum class CalendarStoryType {
```
bvitaliyg, line 37: Extracted story types from Calendar2 story and created a helper method to generate the params  – this is to be used consistently in both stories.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -47,7 +48,8 @@ data class CalendarParams(
   val monthSelectionMode: MonthSelectionMode = MonthSelectionMode.Disabled,
 ) {
 
-  internal val weekFields = WeekFields.of(locale)
+  @InternalBackpackApi
```
bvitaliyg, line 51: Whenever possible, keep internal. Whenever not, used `@InternalBackpackApi`

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarSelection.kt
```diff
@@ -22,11 +22,12 @@ import net.skyscanner.backpack.calendar2.extension.firstDay
 import net.skyscanner.backpack.calendar2.extension.lastDay
 import org.threeten.bp.LocalDate
 import org.threeten.bp.YearMonth
+import java.io.Serializable
 
 /**
  * Describes the current selection in the calendar
  */
-sealed class CalendarSelection {
+sealed class CalendarSelection : Serializable {
```
bvitaliyg, line 30: This is to save the state. Not ideal, but the alternative is to bring the Parcelize plugin.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
```diff
@@ -24,27 +24,32 @@ import androidx.core.text.buildSpannedString
 import net.skyscanner.backpack.calendar2.CalendarParams
 import net.skyscanner.backpack.calendar2.CalendarSelection
 import net.skyscanner.backpack.calendar2.CellInfo
+import net.skyscanner.backpack.util.InternalBackpackApi
 import org.threeten.bp.LocalDate
 import org.threeten.bp.YearMonth
 
-internal sealed class CalendarCell {
+@InternalBackpackApi
+sealed class CalendarCell {
```
bvitaliyg, line 32: Ideally, we also want to annotate these with `@Stable/@Immutable` as Compose won't be able to calculate it itself. However, that means including Compose Runtime into the dependencies of the module. Need you opinion.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendar.kt
```diff
@@ -0,0 +1,90 @@
...
+    Box(modifier = Modifier.weight(1f)) {
+
+      BpkCalendarGrid(
+        state = state,
+        lazyGridState = controller.lazyGirdState,
+        onClick = controller.stateMachine::onClick,
+        modifier = Modifier.fillMaxSize(),
+      )
+
+      var yearToBadge by remember { mutableStateOf(controller.firstVisibleItemYear) }
```
bvitaliyg, line 70: Do not show the current yeah when animation is already on the way out

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlinx.coroutines.CoroutineScope
+import net.skyscanner.backpack.calendar2.CalendarComponent
+import net.skyscanner.backpack.calendar2.CalendarParams
+import net.skyscanner.backpack.calendar2.CalendarSelection
+import net.skyscanner.backpack.calendar2.data.CalendarStateMachine
+import org.threeten.bp.LocalDate
+import java.io.Serializable
+
+@Stable
+class BpkCalendarController private constructor(
```
bvitaliyg, line 39: It's either State, StateMachine or Controller. However, we already have CalendarState, which is a State from MVI point of view, so it's controller to avid confusion.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
```diff
@@ -0,0 +1,97 @@
...
+@Composable
+fun rememberCalendarController(
+  initialParams: CalendarParams,
+  coroutineScope: CoroutineScope = rememberCoroutineScope(),
+  lazyGridState: LazyGridState = rememberLazyGridState(),
+): BpkCalendarController =
+  rememberSaveable(
+    coroutineScope, lazyGridState,
+    saver = Saver<BpkCalendarController, Serializable>(
+      save = { it.state.value.selection },
```
bvitaliyg, line 87: We only store selection – it's expected the same params will be set by the developers, and lazyGridState is saved itself.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -0,0 +1,217 @@
...
+  val status = model.info.status
+  val style = model.info.style
+  val label = model.info.label
+
+  Column(
+    verticalArrangement = Arrangement.Top,
+    horizontalAlignment = Alignment.CenterHorizontally,
+    modifier = modifier
+      .padding(bottom = BpkSpacing.Lg)
+      .selectable(
```
bvitaliyg, line 71: This also improved accessibility

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -0,0 +1,217 @@
...
+        indication = null,
+        selected = selection != null,
+        enabled = !inactive,
+        onClick = { onClick(model) },
+        interactionSource = remember { MutableInteractionSource() },
+      ),
+  ) {
+
+    Box(
```
bvitaliyg, line 80: In Compose we won't be able to center the text vertically without using a parent container

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -0,0 +1,217 @@
...
+      contentAlignment = Alignment.Center,
+      modifier = Modifier
+        .fillMaxWidth()
+        .cellSelectionBackground(selection),
+    ) {
+
+      Spacer(
```
bvitaliyg, line 87: Also, the background is composite here, with 2 layers.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -0,0 +1,217 @@
...
+        CellStatus.Positive -> BpkTheme.colors.statusSuccessSpot
+        CellStatus.Neutral -> BpkTheme.colors.textSecondary
+        CellStatus.Negative -> BpkTheme.colors.textSecondary
+        CellStatus.Empty -> BpkTheme.colors.textDisabled
+      }
+
+    else -> BpkTheme.colors.textSecondary
+  }
+
+private val StartSemiRect = DeflatedRect(horizontal = 0f..0.5f)
```
bvitaliyg, line 216: Notice the new DeflatedRect

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Shapes.kt
```diff
@@ -0,0 +1,41 @@
...
+
+package net.skyscanner.backpack.compose.utils
+
+import androidx.compose.foundation.shape.GenericShape
+import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.util.lerp
+
+internal fun DeflatedRect(
```
bvitaliyg, line 27: And this is an implementation of easy way to create a semi-filled rects

bvitaliyg (commented):
Added some comments

backpack-common/src/androidTest/java/net/skyscanner/backpack/calendar2/data/CalendarCellsLayoutTests.kt
```diff
@@ -21,15 +21,22 @@ package net.skyscanner.backpack.calendar2.data
...
 import org.junit.Assert.assertTrue
+import org.junit.Before
 import org.junit.Test
 import org.threeten.bp.Month
 
 class CalendarCellsLayoutTests {
 
+  @Before
+  fun setup() {
+    initAndroidThreeTen()
```
bvitaliyg, line 37: This was previously done as a part of Calendar1 tests, that's why the prev tests did't fail

app/src/androidTest/java/net/skyscanner/backpack/compose/calendar/BpkCalendarTest.kt
```diff
@@ -0,0 +1,299 @@
...
+  val composeTestRule = createEmptyComposeRule()
+
+  @Before
+  fun setup() {
+    setDimensions(700, 400)
+    AndroidThreeTen.init(testContext)
+  }
+
+  @Test
+  fun screenshotTestCalendarDefault() {
```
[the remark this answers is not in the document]
bvitaliyg: This was done to make 121 comparisons of the screenshots and make sure it works as expected. This have been done already, so we can rename it to remove reducant info

app/src/androidTest/java/net/skyscanner/backpack/compose/calendar/BpkCalendarTest.kt
```diff
@@ -0,0 +1,299 @@
...
+        .assertIsDisplayed()
+    }
+  }
+
+  private fun snap(controller: BpkCalendarController) =
+    composed {
+      BpkCalendar(controller = controller)
+    }
+
+  private fun snap(controller: BpkCalendarController, content: (ComposeTestRule) -> Unit) {
```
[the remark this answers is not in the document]
bvitaliyg: We have some tests which simulate user behavior (click, etc). Compose test rule needs to be used in place.

app/src/main/java/net/skyscanner/backpack/demo/compose/CalendarStory.kt
```diff
@@ -0,0 +1,68 @@
...
+      CalendarStoryType.SelectionWholeMonth -> controller.setSelection(CalendarStorySelection.WholeMonthRange)
+      CalendarStoryType.PreselectedRange -> controller.setSelection(CalendarStorySelection.PreselectedRange)
+      else -> Unit
+    }
+
+    controller.state
+      .filter { it.selection !is CalendarSelection.None }
+      .collect {
+        if (!automationMode) {
+          BpkToast.makeText(context, it.selection.toString(), BpkToast.LENGTH_SHORT).show()
```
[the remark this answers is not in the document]
bvitaliyg: Again, that was done for comparison of the screens. We can replace it with Snackbar, the only concern is that it may overcomplicate the same code. WDYT?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendar.kt
```diff
@@ -0,0 +1,90 @@
...
+      if (controller.firstVisibleItemYear != state.params.now.year) {
+        yearToBadge = controller.firstVisibleItemYear
+      }
+
+      androidx.compose.animation.AnimatedVisibility(
```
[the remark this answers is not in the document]
bvitaliyg: RowScope has it own AnimatedVisibity which conflicts with generic one useless imported specifically. But we may extract everything to another function to make the imports work.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
```diff
@@ -0,0 +1,97 @@
...
+import net.skyscanner.backpack.calendar2.CalendarComponent
+import net.skyscanner.backpack.calendar2.CalendarParams
+import net.skyscanner.backpack.calendar2.CalendarSelection
+import net.skyscanner.backpack.calendar2.data.CalendarStateMachine
+import org.threeten.bp.LocalDate
+import java.io.Serializable
+
+@Stable
+class BpkCalendarController private constructor(
+  internal val lazyGirdState: LazyGridState,
```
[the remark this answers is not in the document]
bvitaliyg: Thanks!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -0,0 +1,217 @@
...
+        Selection.StartMonth,
+        Selection.Middle,
+        Selection.EndMonth -> background(BpkTheme.colors.surfaceHighlight, CircleShape)
+
+        Selection.Single,
+        Selection.Start,
+        Selection.End -> background(BpkTheme.colors.coreAccent, CircleShape)
+      }
+
+    style == CellStatusStyle.Background && status != null && !inactive ->
```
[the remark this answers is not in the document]
bvitaliyg: I was thinking about that, but I'd rather prefer to do it as one single change and not have some versions to support it, and some not. I was planning to remove it in the next PR as a single breaking change.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
bvitaliyg: If there's a function taking CalendarCell, only this function will be recomposed.

However, the entire calendar is a function from CalendarState. And that changes with every interaction, so it will cause recomposition for the entire calendar.

Compose can skip individual cells, but only if the items are stable or immutable. Compose can also determine in a lot of cases whether the item is immutable. Lists are NOT considered to be immutable by default. Java types (LocalDate) that do not have properties in the constructor are not considered to be immutable either.

Because these rules are known I haven't checked the actual recompositions – there's no way the compiler can consider the state to be immutable at this point.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
bvitaliyg: It's expected that the developers will have the same parameters to be generated or stored for the calendar. When you open the screen, you usually know which range you want to display.
In other cases the state (or controller in this case) is stored in survivable context (for instance, business logic or DI context)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
bvitaliyg: It says whether the items is selected and merges descendants

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Shapes.kt
bvitaliyg: The name is inspired by `Rect.deflate()` method in Compose. It draws only a part of the rect – specified by its range parameters. vertical = 0.5f..1f will only draw the bottom half of the rect.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Shapes.kt
```diff
@@ -0,0 +1,41 @@
...
+  horizontal: ClosedFloatingPointRange<Float> = 0f..1f,
```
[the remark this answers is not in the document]
bvitaliyg: The design is that you specify the "range" of the full rect which is drawn. But maybe we need better naming for DeflatedRect or its params to be more representative without the docs.

docs/compose/Calendar2/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Now the component is ready. You can listen for the selection change using its state:
+
+```Kotlin
+controller
+  .state
+  .map { it.selection }
+  .onEach { selection ->
+    when (selection) {
+      is CalendarSelection.None -> showToast("No date is selected")
```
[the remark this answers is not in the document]
bvitaliyg: It was copied from Calendar2 view demo. It was used there to be representative of reaction of any calendar state change. Do you think it's unnecessary now?

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/data/CalendarCell.kt
marianeum: OK in which case: can you benchmark the performance to see if it's a problem? If it is we need to look at what we can do

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
marianeum: ah I misread where this was applied - makes sense!

app/src/main/java/net/skyscanner/backpack/demo/compose/CalendarStory.kt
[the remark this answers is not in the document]
marianeum: snackbar is also essentially deprecated - floating notification is the part that should be used now.
Maybe we need to have a generic floating notification composable that we can use to show in stories

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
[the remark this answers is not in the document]
marianeum: IMO there's no need to add something that goes away in the next PR

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Shapes.kt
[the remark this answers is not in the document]
marianeum: Yeah I think some docs + better naming is needed. But also I wonder if we need to expose the range as I can't see this being used for anything else than half the rect on either side

docs/compose/Calendar2/README.md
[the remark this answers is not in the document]
marianeum: I meant more the `showToast()` part, but yeah maybe even the when statement isn't really needed

app/src/main/java/net/skyscanner/backpack/demo/compose/CalendarStory.kt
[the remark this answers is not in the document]
bvitaliyg: Sounds great, I'll do this!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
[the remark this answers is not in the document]
bvitaliyg: I guess I just want to stay over cautions, checking the new implementation against the existing one and removing the thing together. This code has already been written, I'd prefer to remove it in the next PR, just to cross-check with the other calendar.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendar.kt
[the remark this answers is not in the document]
bvitaliyg: Done

app/src/main/java/net/skyscanner/backpack/demo/compose/CalendarStory.kt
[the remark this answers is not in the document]
bvitaliyg: Done

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarBadge.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.calendar2.CalendarParams
+import net.skyscanner.backpack.compose.badge.BpkBadge
+import net.skyscanner.backpack.compose.badge.BpkBadgeType
+
+@Composable
+internal fun BpkCalendarBadge(
```
marianeum, line 37: much nicer!

docs/compose/Calendar2/README.md
```diff
@@ -85,7 +85,7 @@ controller
   .map { it.selection }
   .onEach { selection ->
     when (selection) {
-      is CalendarSelection.None -> showToast("No date is selected")
+      is CalendarSelection.None -> Unit // do nothing
```
marianeum, line 88: just {} might be nicer (same for all the other ones in the sample)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Shapes.kt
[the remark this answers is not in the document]
marianeum: ok i looked at a preview of what the shape actually does - now i understand it better! I think the range works in that case. I think we just need to make sure the naming of the props is clear - ie it includes if the range you provided is clipped or not.
And I think some documentation to clarify what it actually does would still be useful as I really didn't understand what this was doing

marianeum (commented):
btw is this actually a major change? the package name stayed the same, so it's just an internal change (as common is exposed as an api dependency)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/Shapes.kt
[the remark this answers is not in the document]
bvitaliyg: Added docs. Note that I have explicitly opted-out from KDoc format to not have the internal function included to out docset.

docs/compose/Calendar2/README.md
bvitaliyg: Done

docs/compose/Calendar2/README.md
```diff
@@ -0,0 +1,119 @@
...
+Now the component is ready. You can listen for the selection change using its state:
+
+```Kotlin
+controller
+  .state
+  .map { it.selection }
+  .onEach { selection ->
+    when (selection) {
+      is CalendarSelection.None -> {} // do nothing
+      is CalendarSelection.Single -> showToast("${selection.date} is selected")
```
marianeum, line 89: toast still showing in the other examples

bvitaliyg:
Record snapshots

PR #1316, opened 2022-11-08
Koa-4929 Created BpkSlider as Compose

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests


app/src/main/java/net/skyscanner/backpack/demo/compose/BpkSliderStory.kt
```diff
@@ -0,0 +1,67 @@
...
+    verticalArrangement = Arrangement.spacedBy(BpkSpacing.Base, Alignment.CenterVertically)
+  ) {
+    DefaultSliderSample()
+    RangeSliderSample()
+  }
+}
+
+@Composable
+fun RangeSliderSample() {
+  Text(text = "Range", fontSize = BpkTheme.typography.label2.fontSize)
```
marianeum, line 51: could you please use string resources for this?

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkSliderStory.kt
```diff
@@ -0,0 +1,67 @@
...
+  var rangeSliderValue by remember { mutableStateOf(0.2f..0.8f) }
+  BpkRangeSlider(
+    values = rangeSliderValue,
+    onValueChange = { newValue -> rangeSliderValue = newValue }
+  )
+}
+
+@Composable
+fun DefaultSliderSample() {
+  Text(text = "Standard", fontSize = BpkTheme.typography.label2.fontSize)
```
marianeum, line 61: and here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/BpkSlider.kt
```diff
@@ -0,0 +1,81 @@
...
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.theme.BpkTheme
+
+@Composable
+fun BpkSlider(
+  value: Float,
+  onValueChange: (Float) -> Unit,
+  modifier: Modifier = Modifier,
+  enabled: Boolean = true,
+  valueRange: ClosedFloatingPointRange<Float> = 0f..1f,
```
marianeum, line 38: To align with other platforms rather than having a range here can you have two properties:
`minValue: Float = 0f`, `maxValue: Float = 1f` and then map that internally to the `valueRange`

We also need to expose the `steps` property. It'd be nice to then have a sample for that as well

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/BpkSlider.kt
```diff
@@ -0,0 +1,81 @@
...
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun BpkRangeSlider(
+  values: ClosedFloatingPointRange<Float>,
+  onValueChange: (ClosedFloatingPointRange<Float>) -> Unit,
+  modifier: Modifier = Modifier,
+  enabled: Boolean = true,
+  valueRange: ClosedFloatingPointRange<Float> = 0f..1f,
```
marianeum, line 62: Same thing for this one. (and the `steps` property too)

marianeum (commented):
Nice! Just a couple of comments on the API

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/BpkSlider.kt
jaanainkinen: I think @bvitaliyg advised not to include the steps property, as there isn't an example of that in Figma, unless I misunderstood. I'm happy to add the property and a sample though if needed!

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -151,6 +151,13 @@ class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
         "androidx.compose.material.TextField",
       )
     ),
+    SLIDER(
```
marianeum, line 154: we should also add a version for the `RangeSlider` here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/BpkSlider.kt
```diff
@@ -0,0 +1,90 @@
...
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun BpkRangeSlider(
+  value: ClosedFloatingPointRange<Float>,
+  onValueChange: (ClosedFloatingPointRange<Float>) -> Unit,
+  modifier: Modifier = Modifier,
+  enabled: Boolean = true,
+  minValue: Float = 0f,
+  maxValue: Float = 1f,
+  valueRange: ClosedFloatingPointRange<Float> = minValue..maxValue,
```
marianeum, line 67: same thing here

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -159,6 +159,14 @@ class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
...
+      fullName = "net.skyscanner.backpack.compose.slider.BpkSlider",
+      webName = "slider",
+      componentsToReplace = setOf(
+        "androidx.compose.material.Slider",
+        "androidx.compose.material.RangeSlider",
```
marianeum, line 167: for range slider the replacement isn't `BpkSlider`, but rather `BpkRangeSlider`. Might be better to have two separate entries here to map to the right replacement. The web name would stay the same, though as the docs are both in the same place

marianeum (commented):
almost there :)

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
jaanainkinen: Fixed this but left `full name` same as BpkSlider, as they are both in the same file. Please let me know if this needs fixing!

app/src/main/res/values/strings.xml
```diff
@@ -207,4 +207,7 @@
   <string name="generic_dismiss">Dismiss</string>
   <string name="rating_title">Excellent</string>
   <string name="rating_subtitle">1532 reviews</string>
+  <string name="generic_range">Range</string>
```
marianeum, line 210: small naming nitpick: these seem specific to the slider component, rather than generic, so maybe pre-fix them with `slider` rather than `generic`?

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -159,6 +159,20 @@ class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
         "androidx.compose.material.TextField",
       )
     ),
+    RANGE_SLIDER(
+      fullName = "net.skyscanner.backpack.compose.slider.BpkSlider",
```
marianeum, line 163: this should be `BpkRangeSlider` as the function that replaces it is that - the file it is in isn't really important for consumers

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkSliderStory.kt
```diff
@@ -0,0 +1,83 @@
...
+  ) {
+    DefaultSliderSample()
+    SteppedSliderSample()
+    RangeSliderSample()
+  }
+}
+
+@Composable
+fun RangeSliderSample() {
+  Text(text = stringResource(R.string.generic_range), fontSize = BpkTheme.typography.label2.fontSize)
```
marianeum, line 54: this should use `BpkText` instead and rather than just setting the font size set the full text style

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkSliderStory.kt
```diff
@@ -0,0 +1,84 @@
...
+  ) {
+    DefaultSliderSample()
+    SteppedSliderSample()
+    RangeSliderSample()
+  }
+}
+
+@Composable
+fun RangeSliderSample() {
+  BpkText(text = stringResource(R.string.slider_range), style = TextStyle(fontSize = BpkTheme.typography.label2.fontSize))
```
marianeum, line 55: you can just use `BpkTheme.typography.label2` directly as that's already a `TextStyle`

jaanainkinen: You mean without `fontSize =`? Just trying removing =different things here but can't seem to make it work.

marianeum: so on `BpkText` you're setting the `style` by creating a `TextStyle` with specifying the `fontSize` from an existing style. You don't need to create a `TextStyle` at all and can just use `label2` directly

jaanainkinen: Ah! Finally got what you mean! Fixed now!

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkSliderStory.kt
```diff
@@ -52,7 +51,8 @@ fun BpkSliderStory() {
 
 @Composable
 fun RangeSliderSample() {
-  BpkText(text = stringResource(R.string.slider_range), style = TextStyle(fontSize = BpkTheme.typography.label2.fontSize))
+  BpkText(text = stringResource(R.string.slider_range))
```
marianeum, line 54: Getting closer! You still need to set the style onto the `BpkText` composable - maybe the docs help https://backpack.github.io/components/text/?platform=compose

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkSliderStory.kt
```diff
@@ -0,0 +1,92 @@
...
+  ) {
+    DefaultSliderSample()
+    SteppedSliderSample()
+    RangeSliderSample()
+  }
+}
+
+@Composable
+fun RangeSliderSample() {
+  BpkText(
```
marianeum, line 54: one more thing, sorry :D I would move the label out of the sample, so the snapshot test only includes the actual slider

jaanainkinen: Oh yeah, makes sense! I've fixed this! :D

marianeum (approved):
nice! just need to record the snapshots then we're good to go!

jaanainkinen:
Record snapshots

jaanainkinen:
Record snapshots

PR #1332, opened 2022-11-14
Attempt to make CI less flaky

marianeum:
CI has been VERY flaky recently. It seems to have started to fail since we merged the calendar PR. We were doing some interactions on snapshot tests, as well as testing setting things programmatically. Moved the interaction tests to normal UI tests to simplify the tests and hopefully improve flakiness. 

+ [x] Component `README.md`
+ [x] Tests


app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -63,10 +63,10 @@ open class BpkSnapshotTest {
       .layout()
   }
 
-  protected fun snap(view: View) {
+  protected fun snap(view: View, tags: List<Any> = emptyList()) {
```
marianeum, line 66: this was implemented in the compose version, but never in the view version. re-added it here

app/src/androidTest/java/net/skyscanner/backpack/button/BpkButtonTest.kt
```diff
@@ -172,23 +169,18 @@ class BpkButtonTest(flavour: Flavor) : BpkSnapshotTest() {
...
-      Box(
-        Modifier
-          .fillMaxSize()
-          .background(background())
-          .padding(BpkSpacing.Md),
-        contentAlignment = Alignment.TopStart
-      ) {
-        AndroidView(factory = { content() })
-      }
+    val wrapper = FrameLayout(testContext).apply {
```
marianeum, line 175: stopped relying on compose here. compose tests are much slower to run, so this should boost tests a little

app/src/androidTest/java/net/skyscanner/backpack/compose/calendar/BpkCalendarTest.kt
```diff
@@ -118,73 +88,14 @@ class BpkCalendarTest : BpkSnapshotTest() {
   }
 
   @Test
-  fun withStartDateSelected() {
```
marianeum, line 121: stopped relying on snapshot tests for any of the interactions. we were testing these twice - once setting things programmatically and once via interaction. left the programatical ones here

marianeum:
Record snapshots

marianeum:
Record snapshots

marianeum:
Record snapshots

.github/workflows/_build.yml
```diff
@@ -208,7 +206,7 @@ jobs:
           retention-days: 5
 
       - name: Notify PR of screenshots failure
-        if: ${{ failure() }}
+        if: ${{ failure() && steps.screenshotTests.outputs.CHANGED_FILES != '' }}
```
marianeum, line 209: Ensure we only add the comment if screenshots actually changed as it's mighty confusing right now

.github/workflows/_build.yml
```diff
@@ -194,7 +192,7 @@ jobs:
           disable-animations: true
           script: |
             sudo pip3 install Pillow==9.0.1
-            ./gradlew :app:record${{ env.flavour }}${{ env.config }}AndroidTestScreenshotTest -Pandroid.testInstrumentationRunnerArguments.variant=${{ matrix.variant }}
+            ./gradlew :app:record${{ env.flavour }}${{ env.config }}AndroidTestScreenshotTest -Pandroid.testInstrumentationRunnerArguments.variant=${{ matrix.variant }} --stacktrace
```
marianeum, line 195: include the stacktrace to get a better error

.github/workflows/_build.yml
```diff
@@ -150,15 +150,13 @@ jobs:
         with:
           java-version: '11'
           distribution: 'adopt'
+          cache: gradle
```
marianeum, line 153: attempt to use a different gradle cache

.github/workflows/_build.yml
```diff
@@ -194,6 +188,10 @@ jobs:
           disable-animations: true
           script: |
             sudo pip3 install Pillow==9.0.1
+            adb logcat -c
```
marianeum, line 191: we didn't have access to the logs before, so this will give us some more insight into CI failures

.github/workflows/_build.yml
```diff
@@ -11,7 +11,7 @@ defaults:
 env:
   flavour: oss
   config: debug
-  api-level: 24
+  api-level: 26
```
marianeum, line 14: switched to a newer api level to see if that stabilises things. unfortunately this means all the snapshots change as the font rendering is slightly different

.github/workflows/_build.yml
```diff
@@ -46,13 +46,11 @@ jobs:
         with:
           java-version: '11'
           distribution: 'adopt'
+          cache: gradle
```
marianeum, line 49: changed the way caching was done as the java step has gradle caching built in

.github/workflows/_build.yml
```diff
@@ -103,7 +99,6 @@ jobs:
         if: steps.avd-cache.outputs.cache-hit != 'true'
         uses: reactivecircus/android-emulator-runner@v2
         with:
-          target: google_apis
```
marianeum, line 106: stopped using the google apis version as we don't need it and the non-google one is meant to be more stable. i tried this with the old api level, but the emulator wasn't available

.github/workflows/_build.yml
```diff
@@ -150,14 +144,15 @@ jobs:
...
+          cache: gradle
 
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
           python-version: '3.x'
-
-      - name: Gradle cache
-        uses: gradle/gradle-build-action@v2
+          cache: 'pip'
```
marianeum, line 153: also use a python cache

.github/workflows/_build.yml
```diff
@@ -185,15 +179,16 @@ jobs:
...
-          target: google_apis
           profile: Nexus 4
           sdcard-path-or-size: 512M
           api-level: ${{ env.api-level }}
           force-avd-creation: false
           emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: true
           script: |
-            sudo pip3 install Pillow==9.0.1
+            touch logcat.log
```
marianeum, line 189: we didn't have logs for failures which made it difficult to debug issues. this adds the logs

Backpack/src/androidTest/java/net/skyscanner/backpack/calendar2/BpkCalendarTest.kt
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.threeten.bp.LocalDate
+import java.util.Locale
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(AndroidJUnit4::class)
+class BpkCalendarTest {
```
marianeum, line 46: we had a bunch of snapshot tests for calendar that were doing UI interactions. this is not something we usually do and cover by UI tests instead. so extracted these tests here. the actual UI was tested with other snapshots anyways by setting the state.

build.gradle
```diff
@@ -91,22 +91,25 @@ ext {
   androidXLifecycleViewmodel = "androidx.lifecycle:lifecycle-viewmodel:$lifecycleVersion"
   androidXLifecycleViewmodelKtx = "androidx.lifecycle:lifecycle-viewmodel-ktx:$lifecycleVersion"
 
+  threeTenAbpVersion = "1.4.3"
```
marianeum, line 94: i noticed that we had two separate junit versions due to versions being updated in separate PR and thought this might be the issue. i created a property for libraries that should maintain the same version to avoid this in the future

backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import net.skyscanner.backpack.compose.calendar.internal.CALENDAR_GRID_TEST_TAG
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import org.junit.Assert.*
+import org.junit.Rule
+import org.junit.Test
+import org.threeten.bp.LocalDate
+import java.util.*
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class BpkCalendarTest {
```
marianeum, line 44: same here - just extracted some UI tests

.github/workflows/_build.yml
olliecurtis: Is it a good and ok way for font rednering or is there something that needs to be fixed?

.github/workflows/_build.yml
```diff
@@ -115,7 +110,6 @@ jobs:
       - name: Android Tests
         uses: reactivecircus/android-emulator-runner@v2
         with:
-          target: google_apis
           profile: Nexus 4
```
olliecurtis, line 113: Not part of this but do we need to change the profile to a newer device given the Nexus 4 even in physical device is no longer supported?

olliecurtis (approved):

.github/workflows/_build.yml
marianeum: nope it's just that the rendering changed in this OS version, so this is expected

.github/workflows/_build.yml
marianeum: good question. it doesn't _really_ matter what device we're using - the OS version is the important part. everything else doesn't really change much aside from resolution etc which is a wild west anyways

PR #1346, opened 2022-11-28
KOA-4913 Added BpkBarChart for Compose

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


app/src/main/java/net/skyscanner/backpack/demo/data/BpkBarChartData.kt
```diff
@@ -0,0 +1,73 @@
...
+
+  fun generateModel(): BpkBarChartModel =
+    BpkBarChartModel(
+      items = createMonth(Month.JANUARY) +
+        createMonth(Month.FEBRUARY) +
+        createMonth(Month.MARCH) +
+        createMonth(Month.APRIL) +
+        createMonth(Month.MAY),
+      legend = BpkBarChartModel.Legend(
+        selectedTitle = "Selected",
```
marianeum, line 43: can you use string resources please?

app/src/main/java/net/skyscanner/backpack/demo/data/BpkBarChartData.kt
```diff
@@ -0,0 +1,73 @@
...
+    create: (LocalDate) -> BpkBarChartModel.Item = { createBar(it) },
+  ): List<BpkBarChartModel.Item> =
+    YearMonth.of(year, month)
+      .let { LocalDate.of(it.year, it.month, 1)..LocalDate.of(it.year, it.month, it.lengthOfMonth()) }
+      .toIterable()
+      .map(create)
+
+  fun createBar(
+    date: LocalDate,
+    badge: String = "£" + random.nextInt(100),
```
marianeum, line 60: same

backpack-common/src/main/java/net/skyscanner/backpack/barchart/BpkBarChartModel.kt
```diff
@@ -0,0 +1,83 @@
...
+   * @param value the value of the bar itself, should be a range between 0.0f and 1.0f. If it exceeds, it'll be clamped
+   */
+  @Immutable
+  data class Item(
+    @Stable val key: Any,
+    val title: String,
+    val subtitle: String,
+    val badge: String,
+    val group: String,
+    val inactive: Boolean,
```
marianeum, line 55: When it comes to the API i believe this is something that is incorrect right now (and also comparing it to iOS): on android we can have inactive (which indicates no price), but have a value. however the inactive state just means that it has no value and always shows a fixed height (you can also see that in the figma designs)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
```diff
@@ -57,18 +56,36 @@ fun BpkBadge(
...
+    backgroundColor = type.backgroundColor,
+    contentColor = type.contentColor,
+    borderColor = type.borderColor,
+    modifier = modifier,
+    icon = icon,
+  )
+}
+
+@Composable
+internal fun BpkBadge(
```
marianeum, line 71: there's a new branded badge type that we can add instead of this and kill 2 birds with 1 stone

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/BpkBarChartModel.kt
```diff
@@ -0,0 +1,80 @@
...
+   * @param value the value of the bar itself, should be a range between 0.0f and 1.0f. If it exceeds, it'll be clamped
+   */
+  @Immutable
+  data class Item(
+    @Stable val key: Any,
+    val title: String,
+    val subtitle: String,
+    val badge: String? = null,
+    val group: String,
+    @FloatRange(from = 0.0, to = 1.0) val value: Float,
```
marianeum, line 54: we treat the badge text missing as inactive, but an inactive item has no value either - could the two be a separate model that is nullable? otherwise it'll be confusing that you can set a value, but it has no effect

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/internal/BarChartTitle.kt
```diff
@@ -0,0 +1,78 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.barchart.BpkBarChartModel
+import net.skyscanner.backpack.compose.text.BpkText
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.utils.ScrollingDirection
+import net.skyscanner.backpack.compose.utils.lastScrollingDirection
+
+@OptIn(ExperimentalAnimationApi::class)
```
marianeum, line 39: what's experimental here? we should generally avoid using experimental apis as they'll likely change in the future. this is unavoidable in some cases where the material component isn't stable yet, but in an animation case relying on experimental apis is risky

docs/compose/BarChart/README.md
```diff
@@ -0,0 +1,72 @@
...
+  selected = selectedItem,
+  onSelectionChange = { selectedItem = it },
+)
+```
+
+To render data you need to create a model for `BarChart` in code.
+Detailed documentation on each property of the model is
+available [here](https://backpack.github.io/android/Backpack/net.skyscanner.backpack.compose.barchart).
+
+Here's some an example which generates a model for one single month in Kotlin:
```
marianeum, line 42: remove the some?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/internal/BarChartTitle.kt
bvitaliyg: AnimatedContent() is experimental. It's also the way the content animation should be performed according to this diagram: https://developer.android.com/jetpack/compose/animation

A lot of animation APIs are experimental in Compose. It doesn't mean the code is buggy or not ready to be used – it just means the API might be changed. In most of the cases it'll be deprecated first, so we'll have enough time to migrate away. We already use this animation in some other places, I wouldn't be worried about that.

Besides, in worth case scenario this specific usage can just be replaced with Crossfade or have no animation at all.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/BpkBarChartModel.kt
bvitaliyg: It's expected from the design that no badge makes the columns to have a fixed height. However, I think we can do this – to have better API and less confusion.

app/src/main/java/net/skyscanner/backpack/demo/data/BpkBarChartData.kt
```diff
@@ -77,16 +80,13 @@ object BpkBarChartData {
   fun createBar(
     date: LocalDate,
     locale: Locale = Locale.getDefault(),
-    badge: String? = null,
-    value: Float = 0.5f,
-    inactive: Boolean = false,
+    values: BpkBarChartModel.Values? = null,
```
marianeum, line 83: singular value? it's still referring to a single value, represented in two different forms

app/src/androidTest/java/net/skyscanner/backpack/compose/badge/BpkBadgeTest.kt
```diff
@@ -102,4 +102,12 @@ class BpkBadgeTest : BpkSnapshotTest() {
       BpkBadge(text = BpkBadgeType.Outline.toString(), type = BpkBadgeType.Outline)
     }
   }
+
+  @Test
+  fun brand() {
```
marianeum, line 107: can you add that to the story & docs as well?

marianeum: the docs screenshot needs regenerating then

app/src/main/java/net/skyscanner/backpack/demo/data/BpkBarChartData.kt
```diff
@@ -80,14 +80,24 @@ object BpkBarChartData {
...
+    return BpkBarChartModel.Item(
       key = date,
-      title = date.dayOfWeek.getDisplayName(TextStyle.SHORT, locale),
-      subtitle = date.dayOfMonth.toString(),
+      title = dayOfWeek,
+      subtitle = dayOfMonth,
       group = date.month.getDisplayName(TextStyle.FULL, locale),
       values = values,
+      accessibilityLabel = if (values != null) {
+        resources.getString(R.string.bar_chart_accessibility_label, fullDayOfWeek, dayOfMonth, values.text)
```
marianeum, line 97: rather than building the date string ourselves we can use the same logic we apply in calendar to build a date

bvitaliyg: I think it's an overkill for a sample app. It's important to show the principle, but using resources, currency, and date formatters make the code harder to read (which is most important for a sample) and maintain

docs/compose/BarChart/README.md
```diff
@@ -0,0 +1,72 @@
...
+Here's an example which generates a model for one single month in Kotlin:
+
+```kotlin
+import net.skyscanner.backpack.compose.barchart.BpkBarChartModel
+import org.threeten.bp.LocalDate
+import org.threeten.bp.YearMonth
+import org.threeten.bp.format.TextStyle
+import java.util.Locale
+
+fun createBarChartModel(locale: Locale, yearMonth: YearMonth) =
```
marianeum, line 51: this needs updating

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -133,6 +136,36 @@ open class BpkSnapshotTest {
...
+    size: IntSize = IntSize(width, height),
+    background: Color = Color.Unspecified,
+    vararg providers: ProvidedValue<*>,
+    content: @Composable () -> Unit,
+  ) {
+    Assume.assumeFalse(BpkTestVariant.current == BpkTestVariant.Themed)
+
+    val asyncScreenshot = prepareForAsyncTest()
+
+    val scenario = launchActivity<AppCompatActivity>()
```
marianeum, line 149: the compose test rule already has launched an activity - does this need to be launched again?

app/src/androidTest/java/net/skyscanner/backpack/compose/barchart/BpkBarchartTests.kt
```diff
@@ -31,6 +36,9 @@ import kotlin.math.roundToInt
 @RunWith(AndroidJUnit4::class)
 class BpkBarchartTests : BpkSnapshotTest() {
 
+  @get:Rule
+  val composeTestRule = createAndroidComposeRule<AppCompatActivity>()
```
marianeum, line 40: by using a rule here this activity always gets started, no matter if the test actually gets run by the `assumeVariant` check. `createEmptyComposeRule` allows you to start the activity manually later

app/src/androidTest/java/net/skyscanner/backpack/compose/barchart/BpkBarchartTests.kt
```diff
@@ -31,6 +36,9 @@ import kotlin.math.roundToInt
 @RunWith(AndroidJUnit4::class)
 class BpkBarchartTests : BpkSnapshotTest() {
```
marianeum, line 37: nitpick on the file name. `Test` - singular

app/src/androidTest/java/net/skyscanner/backpack/compose/barchart/BpkBarchartTests.kt
```diff
@@ -131,14 +139,15 @@ class BpkBarchartTests : BpkSnapshotTest() {
...
-      val model = BpkBarChartModel(caption = "Bar chart", items = createMonth(Month.JANUARY))
+    val model = BpkBarChartModel(caption = "Bar chart", items = createMonth(Month.JANUARY))
+    composeTestRule.setScreenshotContent {
       BpkBarChart(
         model = model,
         selected = model.items[10],
         onSelectionChange = {},
       )
     }
+    composeTestRule.onNodeWithText(model.items[10].values?.text!!).assertIsDisplayed()
```
marianeum, line 150: how come it's only this test that needs to wait for the content?

bvitaliyg: We're waiting for the initial composition to get anchor position plus there's animation involved

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/internal/BarChartColumn.kt
```diff
@@ -63,7 +63,7 @@ internal fun BarChartColumn(
   Column(
     horizontalAlignment = Alignment.CenterHorizontally,
     modifier = modifier
-      .semantics { contentDescription = model.accessibilityLabel }
+      .semantics { contentDescription = model.accessibilityLabel.toString() }
```
marianeum, line 66: as you're just changing this back to `toString` this does nothing. how the `TtsSpan` works (at least in the view world) is that you include the accessibility content as part of the text on a view and it'll read out the extra accessibility content as part of reading the content text. so if your text is August 11 the `TtsSpan` will include a span for this part saying that this is a date.

bvitaliyg: Currently, there's no Tts support (TtsAnnotation) in Compose at all. I've changed the API to AnnotatedString to support it in the future, but for now yes, just `.toString()` method. Also, there's no way to use TtsSpan from View and no direct access to View accessibility node from Compose.

marianeum: yeah what i'm saying is: using an annotated string here won't work for the content description (as that's just a string) - the `TtsSpan` needs to be part of the content text directly, not a separate content description, so changing the content description to an annotated string won't help - we'd have to change the content text to be an `AnnotatedString` and not have an `accessibilityLabel` at all.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/internal/BarChartTitle.kt
```diff
@@ -0,0 +1,78 @@
...
+@Composable
+internal fun BarChartTitle(
```
marianeum, line 41: Is the title announced for screenreaders when it changes?

bvitaliyg: In most cases, the title will be the name of the month. The same name will be pronounced each time user selects a cell if accessibility is set up correctly. Pronouncing just the title whenever the list scrolls will simply make things more confusing.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/internal/BarChartColumn.kt
bvitaliyg: We don't have content text. We only got the title, subtitle, and value text. If we make any of these parameters AnnotatedString it will make the API more complicated. Also, it's not clear why it's the title (and not subtitle, or value) that should play the main role as an accessibility label provider.

When TttsAnnotation will be available we can inside set it as a text to any component we want. For instance, we can add TtsAnnotation to the title, but keep the title text the same as it was. But those are the implementation detail

app/src/androidTest/java/net/skyscanner/backpack/compose/barchart/BpkBarchartTest.kt
```diff
@@ -0,0 +1,168 @@
...
+          items = createMonth(Month.JANUARY),
+        ),
+        selected = null,
+        onSelectionChange = {},
+      )
+    }
+  }
+
+  @Test
+  fun selected() {
```
marianeum, line 140: looks like there's an issue with RTL

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/barchart/BpkBarChartModel.kt
```diff
@@ -42,7 +41,8 @@ data class BpkBarChartModel(
    * @param title a primary text placed just below the bar itself.
    * @param subtitle a secondary text placed just below the title.
    * @param group name of the group in which the item exists. The name is rendered above the bars and updated as the chart scrolls horizontally.
-   * @param accessibilityLabel the label to be used for screen readers to read the values (e.g. "Monday 7, price is £100")
+   * @param contentDescription the content description to be used for screen readers to read the values (e.g. "Monday 7, price is £100")
```
marianeum, line 44: can you update the example to provide what needs to be included?

marianeum: to clarify: i meant the "Monday 7, price is £100" - also needs to include the month/group name here

bvitaliyg: Done

marianeum (approved):

PR #1349, opened 2022-12-01
SONIC-1167 Add custom icon to BpkButton-Compose

mineschan:
Add drawable icon support to Compose version of BpkButton 

+ [x] Component `README.md`
+ [x] Tests


mineschan:
Record snapshots

app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
```diff
@@ -44,6 +45,7 @@ class BpkButtonTest(flavour: Flavor) : BpkSnapshotTest() {
   private val type: BpkButtonType = flavour.first
   private val size: BpkButtonSize = flavour.second
   private val icon = BpkIcon.LongArrowRight
+  private val iconDrawableRes = R.drawable.button_sample_icon
```
marianeum, line 48: can you use a different sample icon please? this might cause copyright issues.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButton.kt
```diff
@@ -133,6 +135,66 @@ fun BpkButton(
   }
 }
 
+@Composable
+fun BpkButton(
+  @DrawableRes iconResource: Int,
```
marianeum, line 140: rather than fixing this to the `iconResource` (and therefore creating a dependency on the existing resource system) could this be an `icon` composable instead?

marianeum:
Btw the record snapshots comment is no longer the appraoch to take - the `Record snapshots` commit is correct (we had to change it to a commit following some security concerns). Just make sure you wait for that run to succeed before making other commits.

app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
xo0ps: changed to backpack's icon

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButton.kt
xo0ps: changed to use `Painter` instead

app/src/androidTest/java/net/skyscanner/backpack/compose/button/BpkButtonTest.kt
```diff
@@ -96,7 +99,7 @@ class BpkButtonTest(flavour: Flavor) : BpkSnapshotTest() {
     // icon is bigger on large size, so we need to test this
 
     capture {
-      BpkButton("Button", icon, BpkButtonIconPosition.Start, type = type, size = size, onClick = {})
+      BpkButton("Button", icon, BpkButtonIconPosition.End, type = type, size = size, onClick = {})
```
marianeum, line 102: why was this changed to end?

marianeum: ah you're right! i got confused by the github formatting as it didn't include the function title

marianeum (approved):

PR #1355, opened 2022-12-07
KOA-5680 Compose-Material is not a transitive dependency anymore

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
```diff
@@ -52,30 +55,31 @@ fun BpkTheme(
...
   val typography = BpkTypography(defaultFontFamily = fontFamily)
   val colors = if (isSystemInDarkTheme()) BpkColors.dark() else BpkColors.light()
+  val shapes = BpkShapes()
 
-  CompositionLocalProvider(
-    LocalBpkTypography provides typography,
-    LocalBpkColors provides colors,
-    LocalContentColor provides colors.textPrimary,
-    LocalElevationOverlay provides null,
+  MaterialTheme(
```
marianeum, line 60: what's the reason for swapping these around?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/divider/BpkDivider.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.compose.material.Divider
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import net.skyscanner.backpack.compose.theme.BpkTheme
+
+@Composable
+fun BpkDivider(
```
marianeum, line 30: did you talk to design about this? we might want to limit colours here to line and line on dark and have a style, as well as for thickness, rather than just exposing the props as they are

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
```diff
@@ -52,30 +55,31 @@ fun BpkTheme(
...
+    typography = typography.toMaterialTypography(fontFamily),
```
marianeum, line 61: do we still need the typography/colour/shape matching? in our components we always set these etc and as material is no longer exposed the mapping shouldn't be necessary anymore (or at least very minimal)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/divider/BpkDivider.kt
bvitaliyg: Not yet, it's still WIP

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/theme/BpkTheme.kt
marianeum: Having the full compatibility might set expectations that using material is supported by us, however we do not - minimal support would work in our favour here. It's relatively easy for us to test if any of the mapping is required internally - just remove it and see if any snapshots change. 
It'll be harder to remove the mapping the longer we keep it

marianeum (approved):

PR #1360, opened 2022-12-12
SONIC-1202 BpkCardButton

xo0ps:
Added `BpkSaveButton` and `BpkShareButton`

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/BpkCardButton.kt
```diff
@@ -0,0 +1,45 @@
...
+  OnDark
+}
+
+enum class BpkCardButtonSize {
+  Default,
+  Small
+}
+
+@Composable
+fun BpkCardButton(
```
marianeum, line 24: rather than having a type here it might be better to have two separate composable functions - one `BpkSaveButton` and another `BpkShareButton` and have the different params. that way the components are a little easier to use

app/src/main/java/net/skyscanner/backpack/demo/compose/CardButtonsStory.kt
```diff
@@ -0,0 +1,148 @@
...
+        var checked1 by remember { mutableStateOf(false) }
+        BpkCardButton(
+          type = BpkCardButtonType.Save(checked = checked1, onCheckedChange = { checked1 = !checked1 }),
+          size = size,
+          style = BpkCardButtonStyle.OnDark,
+          contentDescription = "",
+        )
+        var checked2 by remember { mutableStateOf(true) }
+        BpkCardButton(
+          type = BpkCardButtonType.Save(checked = !checked2, onCheckedChange = { checked2 = !checked2 }),
```
marianeum, line 99: is the `!checked` needed in the initial state? looking at the screenshots both saved & unsaved options are unchecked

app/src/androidTest/java/net/skyscanner/backpack/compose/cardbutton/BpkCardButtonTest.kt
```diff
@@ -0,0 +1,98 @@
...
+@RunWith(Parameterized::class)
+class BpkCardButtonTest(flavor: Flavor) : BpkSnapshotTest() {
+
+  private val type: BpkCardButtonType = flavor.type
+  private val size: BpkCardButtonSize = flavor.size
+  private val style: BpkCardButtonStyle = flavor.style
+
+  @Test
+  fun default() {
+    assumeTrue(style == BpkCardButtonStyle.Default)
```
marianeum, line 29: looks like all three tests always assume their style. does the style need to be part of the paramiterization?

app/src/main/java/net/skyscanner/backpack/demo/compose/CardButtonsStory.kt
```diff
@@ -0,0 +1,148 @@
...
+  size: BpkCardButtonSize,
+  modifier: Modifier = Modifier,
+) {
+  LazyColumn(
+    modifier
+      .fillMaxSize()
+      .padding(top = BpkDimension.Spacing.Md)
+  ) {
+    item {
+      Row(
```
marianeum, line 59: could you extract this into a separate function with the style + background colour as a param and reuse to avoid having the three buttons defined multiple times per style

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/BpkCardButton.kt
xo0ps: done

app/src/androidTest/java/net/skyscanner/backpack/compose/cardbutton/BpkCardButtonTest.kt
xo0ps: fixed it

app/src/main/java/net/skyscanner/backpack/demo/compose/CardButtonsStory.kt
xo0ps: done

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/internal/BpkCardButtonImpl.kt
```diff
@@ -0,0 +1,151 @@
...
+    contentAlignment = Alignment.Center
+  ) {
+    Box(
+      modifier = Modifier
+        .clip(shape = CircleShape)
+        .size(if (size == BpkCardButtonSize.Default) BpkSpacing.Xxl else BpkSpacing.Xl)
+        .background(
+          color = when (style) {
+            BpkCardButtonStyle.Contained -> BpkTheme.colors.surfaceDefault.copy(alpha = 0.5F)
+            else -> BpkTheme.colors.textOnDark.copy(alpha = 0.0F)
```
marianeum, line 89: you can also just use `Color.Transparent`

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/internal/BpkCardButtonImpl.kt
```diff
@@ -0,0 +1,151 @@
...
+@Composable
+fun BpkSaveCardButtonImpl(
+  checked: Boolean,
+  contentDescription: String,
+  style: BpkCardButtonStyle,
+  size: BpkCardButtonSize,
+  onCheckedChange: (Boolean) -> Unit
+) {
+  var state by remember { mutableStateOf(BpkCardButtonState.Default) }
+  val scaleAnimation by animateFloatAsState(
```
marianeum, line 64: the animation looks a little different to the spec in figma - it might be worth sitting with adam to tweak this

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/internal/BpkCardButtonImpl.kt
```diff
@@ -0,0 +1,151 @@
...
+
+@Composable
+fun BpkSaveCardButtonImpl(
+  checked: Boolean,
+  contentDescription: String,
+  style: BpkCardButtonStyle,
+  size: BpkCardButtonSize,
+  onCheckedChange: (Boolean) -> Unit
+) {
+  var state by remember { mutableStateOf(BpkCardButtonState.Default) }
```
marianeum, line 63: something's not right with this - i can sometimes get it into a state where i can't toggle the button

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/internal/BpkCardButtonImpl.kt
xo0ps: spent some time and tweaked it. now it looks more similar IMO.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/internal/BpkCardButtonImpl.kt
xo0ps: could you please let me know how you ended up in that state? I couldn't reproduce it on my device and emulator and compose interactive mode.

marianeum: I can't get 100% steps to reproduce, but it sometimes happens if i tap it again around the time the animation should finish. I also managed to just get it into a state (with the latest version) where the animation doesn't run at all and just switches between the states and it's permanently stuck in this state

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardbutton/internal/BpkCardButtonImpl.kt
marianeum: looking at the spec in figma it's still a bit different (the spec is written out on the right in figma)

marianeum (approved):
nice one, thanks!

PR #1361, opened 2022-12-13
KOA-5780 Added detekt 

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


.idea/compiler.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="11" />
+    <bytecodeTargetLevel target="17" />
```
marianeum, line 4: is that intentional?

Backpack/src/main/java/net/skyscanner/backpack/bottomsheet/BpkBottomSheetBehaviour.kt
```diff
@@ -56,7 +56,7 @@ class BpkBottomSheetBehaviour<V : View>(context: Context, attrs: AttributeSet? =
       override fun onSlide(bottomSheet: View, slideOffset: Float) {
         updateBackground(slideOffset)
       }
-    })
+    },)
```
marianeum, line 59: there's some of these that feel wrong - can we also enforce putting the ) onto the next line?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/dialog/BpkDialog.kt
```diff
@@ -30,6 +30,7 @@ import net.skyscanner.backpack.compose.dialog.internal.Dialog
 import net.skyscanner.backpack.compose.icon.BpkIcon
 
 @Composable
+@Suppress("ModifierMissing")
```
marianeum, line 33: didn't we want to add the modifier as it also causes the lint warning on the app side when using the dialog and having a composable for it?

build.gradle
```diff
@@ -47,30 +49,23 @@ allprojects {
...
-  ktlint {
-    version = "0.43.2"
-    enableExperimentalRules = false
-    reporters {
-      reporter "checkstyle"
-    }
+  detekt {
+    config = files("$rootDir/.detekt.yml", "$rootDir/.detekt-compose.yml")
+    buildUponDefaultConfig = true
+    autoCorrect = true
```
marianeum, line 62: Is this going to automatically fix issues when running detekt? how's that working on CI - particularly with the snapshot testing checking for any files changing when recording snapshots.
also in general i prefer explicitly running a fix command - that way it's easier to look at the issues and potentially manually fixing something (ie automatic line breaks/formatting can sometimes cause very odd formatting

.detekt.yml
```diff
@@ -0,0 +1,584 @@
+build:
```
marianeum, line 1: can you clarify what's different to the app config here?

.detekt-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+  ComposableParamOrder:
+    active: true
+  PreviewNaming:
+    active: true
+  PreviewPublic:
+    active: true
+  RememberMissing:
+    active: true
+  UnstableCollections:
+    active: false
```
marianeum, line 31: why is this one disabled?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/dialog/BpkDialog.kt
bvitaliyg: Yes! But I think it should be in the next ticket, this is an API change.

build.gradle
bvitaliyg: It will only fix the issues on the 2nd run, so CI is fine. Double-running to fix it is an approach used widely in the industry and it's generally safe. And faster, more convinient – if your commit fails, just run it again, instead of opening terminal, running the command and try to commit again (or trying to manually resolve the issues from the commit log).

marianeum: Are you sure this is what it's doing? Afaik it'll fix the errors, but fail the task to make you aware of it. Next time you run it it succeeds because the errors were fixed in the previous run. 
Maybe instead of setting `autoCorrect` here we can change the git hook to use the `--auto-correct` property if we want the behaviour of auto fixing for that, but keep CI separate. That way also if somebody wants to just run detekt without auto correcting that is still possible.

Backpack/src/main/java/net/skyscanner/backpack/bottomsheet/BpkBottomSheetBehaviour.kt
bvitaliyg: I found no way to do that so far

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+  UndocumentedPublicProperty:
+    active: false
+
+complexity:
+  active: false
+  ComplexCondition:
+    active: false
+  ComplexInterface:
+    active: false
+  CyclomaticComplexMethod:
```
bvitaliyg, line 44: ComplexMethod was deprecated in faviour of CyclomaticComplexMethod

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+  FunKeywordSpacing:
+    active: false
+  FunctionTypeReferenceSpacing:
+    active: false
+  ImportOrdering:
+    active: false
+  Indentation:
+    active: true
+    autoCorrect: true
+    indentSize: 2
```
bvitaliyg, line 186: Changed from 4 to 2

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+  NoWildcardImports:
+    active: true
+    packagesToUseImportOnDemandProperty: 'java.util.*,kotlinx.android.synthetic.**'
+  PackageName:
+    active: true
+    autoCorrect: true
+  ParameterListWrapping:
+    active: true
+    autoCorrect: true
+    maxLineLength: 125
```
bvitaliyg, line 242: Changed from 160 to 125

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+  SpacingAroundUnaryOperator:
+    active: false
+  SpacingBetweenDeclarationsWithAnnotations:
+    active: false
+  SpacingBetweenDeclarationsWithComments:
+    active: false
+  StringTemplate:
+    active: true
+    autoCorrect: true
+  TrailingCommaOnCallSite:
```
bvitaliyg, line 282: There was one deprecated rule used, it's now 2

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+  TrailingCommaOnDeclarationSite:
+    active: true
+  TypeArgumentListSpacing:
+    active: false
+  UnnecessaryParenthesesBeforeTrailingLambda:
+    active: false
+  Wrapping:
+    active: false
+
+libraries:
```
bvitaliyg, line 294: These rules was deprecated in their previous category and moved to library

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+    active: false
+  NonBooleanPropertyPrefixedWithIs:
+    active: false
+  ObjectPropertyNaming:
+    active: false
+  PackageNaming:
+    active: false
+  TopLevelPropertyNaming:
+    active: true
+    constantPattern: '[A-Z][_A-Z0-9a-z]*'
```
bvitaliyg, line 340: Changed from `'[A-Z][_A-Z0-9]*'` to match Compose properties, according to detekt gudelines

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+    propertyPattern: '[A-Za-z][_A-Za-z0-9]*'
+    privatePropertyPattern: '_?[A-Za-z][_A-Za-z0-9]*'
+  VariableMaxLength:
+    active: false
+  VariableMinLength:
+    active: false
+  VariableNaming:
+    active: false
+
```
bvitaliyg, line 349: No DuplicateCaseInWhenExpression, MissingWhenCase, RedundantElseInWhen rules – thouse were deprecated as complier performs this checks natively

.detekt.yml
bvitaliyg: Done

build.gradle
bvitaliyg: Done

.detekt.yml
```diff
@@ -0,0 +1,584 @@
...
+  MagicNumber:
+    active: false
+  MandatoryBracesIfStatements:
+    active: false
+  MandatoryBracesLoops:
+    active: false
+  MaxChainedCallsOnSameLine:
+    active: false
+  MaxLineLength:
+    maxLineLength: 125
```
bvitaliyg, line 479: Also changed from 160 to 125

.detekt.yml
marianeum: might be worth creating a follow up ticket to move that to 4 as that's the app & industry standard for android projects

.detekt.yml
marianeum: similar here - maybe we align that to 160 as well - not really a reason to differentiate that from the app

marianeum (approved):

PR #1380, opened 2022-12-29
Koa 5678 Autmate screenshots recording

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


.github/workflows/_build.yml
```diff
@@ -256,3 +256,107 @@ jobs:
...
+          cache: gradle
+
+      - name: AVD cache
+        uses: actions/cache@v3
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-${{ env.api-level }}
```
marianeum, line 292: as this AVD will use a different api level to the other ones you'll need to use a different key here otherwise the cache is useless and it'll cache the wrong emulator

CONTRIBUTING.md
```diff
@@ -120,7 +120,7 @@ For some more complex components with many different types you may want to make
 After adding new snapshot tests or making UI changes, run
 
 ```
- ./scripts/record_screenshot_tests.sh
+ ./scripts/record_snapshot_tests.sh
```
marianeum, line 123: can you also rename the verify script so they're consistent?

.github/workflows/_build.yml
```diff
@@ -256,3 +256,107 @@ jobs:
...
+            var body = `### Screenshots were updated. Please verify the changes match the expected output. \n\n>'${{ steps.screenshotsRecording.outputs.CHANGED_FILES }}'`;
+
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: body
+            })
+
+      - name: Commit changes
```
marianeum, line 350: right now our docs screenshots aren't 100% reproducable - I believe the (old) calendar one changes depending on the date still. we need to fix that first before we enable this otherwise it will always pollute our PRs

marianeum (changes requested):

marianeum (commented):
you should be able to test the behaviour on this PR too on CI - this only didn't work with the comment trigger

.github/workflows/_build.yml
```diff
@@ -256,3 +256,107 @@ jobs:
           git add app/screenshots/*
           git diff-index --quiet HEAD || git commit -m "Updated snapshots for '${{ matrix.variant }}'"
           git push
+
+  Screenshot:
+    name: Screenshots recording
+    runs-on: macos-latest
+    needs: CheckCommitMessage
+    if: ${{ needs.CheckCommitMessage.outputs.head-commit-message == 'Record screenshots' }}
```
marianeum, line 264: we should probably do this the same way we run the snapshot tests: always run them and check for changes and only commit with the trigger (and possibly use the same trigger as well to avoid having to wait twice). otherwise we risk those getting out of sync again

PR #1387, opened 2023-01-03
Fix snapshot flakiness

marianeum:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
+++ b/app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
@@ -50,9 +50,6 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
     @Dimension(unit = DP) width: Int? = null,
     @Dimension(unit = DP) height: Int? = null,
     @Dimension(unit = DP) padding: Int = testContext.resources.getDimensionPixelSize(R.dimen.bpkSpacingMd),
-    // wrapView should almost always be true - in some rare cases (like Snackbar) wrapping the view may break.
-    // padding + background won't be applied if used
-    wrapView: Boolean = true,
   ) {
     runOnUi {
       view.layoutDirection = testContext.resources.configuration.layoutDirection
@@ -61,34 +58,27 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
       view = view,
       background = background,
       padding = padding,
-      wrapView = wrapView,
     )
     wrappedView.measure(measureSpec(width), measureSpec(height))
     compareScreenshot(
       view = wrappedView,
-      widthInPx = wrappedView.measuredWidth,
-      heightInPx = wrappedView.measuredHeight,
+      widthInPx = width?.let { dpToPx(it) } ?: wrappedView.measuredWidth,
+      heightInPx = height?.let { dpToPx(it) } ?: wrappedView.measuredHeight,
       name = screenshotName(tags),
     )
   }
 
   private fun measureSpec(size: Int?): Int {
-    val dimenSize = if (size == null) {
-      0
-    } else {
-      (size * testContext.resources.displayMetrics.density).toInt()
-    }
+    val dimenSize = size?.let { dpToPx(it) } ?: 0
     return makeMeasureSpec(dimenSize, if (size == null) View.MeasureSpec.UNSPECIFIED else View.MeasureSpec.EXACTLY)
   }
 
-  private fun wrapMeasuredViewWithBackground(view: View, background: Int, padding: Int, wrapView: Boolean): View {
-    if (!wrapView) {
-      return view
-    }
+  private fun wrapMeasuredViewWithBackground(view: View, background: Int, padding: Int): View {
     val result = FrameLayout(view.context)
     if (view.parent != null) {
       runOnUi {
         (view.parent as ViewGroup).removeView(view)
+        view.visibility = View.VISIBLE
       }
     }
     result.addView(view)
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -50,9 +50,6 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
     @Dimension(unit = DP) width: Int? = null,
     @Dimension(unit = DP) height: Int? = null,
     @Dimension(unit = DP) padding: Int = testContext.resources.getDimensionPixelSize(R.dimen.bpkSpacingMd),
-    // wrapView should almost always be true - in some rare cases (like Snackbar) wrapping the view may break.
```
marianeum, line 53: this was a workaround for the snackbar tests which is now no longer needed

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -61,34 +58,27 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
...
       background = background,
       padding = padding,
-      wrapView = wrapView,
     )
     wrappedView.measure(measureSpec(width), measureSpec(height))
     compareScreenshot(
       view = wrappedView,
-      widthInPx = wrappedView.measuredWidth,
-      heightInPx = wrappedView.measuredHeight,
+      widthInPx = width?.let { dpToPx(it) } ?: wrappedView.measuredWidth,
```
marianeum, line 65: always use the exact size if specified, otherwise measured one. this shouldn't change any tests, but is more consistent with what we're doing

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -61,34 +58,27 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
...
     )
   }
 
   private fun measureSpec(size: Int?): Int {
-    val dimenSize = if (size == null) {
-      0
-    } else {
-      (size * testContext.resources.displayMetrics.density).toInt()
-    }
+    val dimenSize = size?.let { dpToPx(it) } ?: 0
```
marianeum, line 72: simplify this and use the helper function

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -61,34 +58,27 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
...
     return makeMeasureSpec(dimenSize, if (size == null) View.MeasureSpec.UNSPECIFIED else View.MeasureSpec.EXACTLY)
   }
 
-  private fun wrapMeasuredViewWithBackground(view: View, background: Int, padding: Int, wrapView: Boolean): View {
-    if (!wrapView) {
-      return view
-    }
+  private fun wrapMeasuredViewWithBackground(view: View, background: Int, padding: Int): View {
```
marianeum, line 76: remove the `wrapView` boolean

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -61,34 +58,27 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
...
     val result = FrameLayout(view.context)
     if (view.parent != null) {
       runOnUi {
         (view.parent as ViewGroup).removeView(view)
+        view.visibility = View.VISIBLE
```
marianeum, line 81: turns out that was the root cause of the snackbar tests not working with a wrapped view. the snackbar is set to invisible first for some reason. we never want to test invisble views anyways

app/src/androidTest/java/net/skyscanner/backpack/BpkSnapshotTest.kt
```diff
@@ -97,4 +87,8 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
 
     return result
   }
+
+  private fun dpToPx(@Dimension dp: Int): Int {
```
marianeum, line 91: we have that helper function already in the backpack module, but it's internal. rather than exposing for the app declare it again here

app/src/androidTest/java/net/skyscanner/backpack/bottomsheet/BpkBottomSheetTest.kt
```diff
@@ -58,18 +59,22 @@ class BpkBottomSheetTest : BpkSnapshotTest() {
 
   @Test
   fun default() {
-    snap(setupBottomSheet(), width = 200, height = 200, padding = 0)
+    capture(STATE_COLLAPSED)
```
marianeum, line 62: this test was sometimes flaky (also with fb testing before) - always use the convoluted way of testing as the expanded one was consistent before too. will monitor further. if it's still flaky i have another workaround, but that requires a production code change as well, so want to avoid

app/src/androidTest/java/net/skyscanner/backpack/snackbar/BpkSnackbarTests.kt
```diff
@@ -132,40 +113,9 @@ class BpkSnackbarTests : BpkSnapshotTest() {
...
-        rule.scenario.waitForActivity()
-      }
-
-    snap(snackbar!!.rawSnackbar.view, wrapView = false, width = 350)
-  }
-}
-
-fun waitForOpaque(timeout: Long): ViewAction {
-  return object : ViewAction {
+    val snackbar: BpkSnackbar = what()
```
marianeum, line 116: simplified the snackbar test a lot. no more requirement on the activity 🎉

marianeum (commented):
i re-ran the tests a few times to verify flakiness. seem to run stable now.

frugoman (approved):

PR #1398, opened 2023-01-10
KOA-5642 Cleanup automirror logic

marianeum:

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/compose/IconsStory.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/compose/IconsStory.kt
@@ -18,7 +18,6 @@
 
 package net.skyscanner.backpack.demo.compose
 
-import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/skyscanner/backpack/demo/compose/IconsStory.kt
```diff
@@ -102,14 +100,6 @@ fun IconsStoryCompose() {
     val context = LocalContext.current
     val clipboardManager = LocalClipboardManager.current
 
-    val autoMirrorBackground = BpkTheme.colors.statusSuccessSpot
```
marianeum, line 105: no longer highlighting that an icon can be automirrored in the sample. As that's the only leftover usage for it it doesn't feel important enough as it's quite visible which ones rotate anyways. The automirror property was also technically visible to consumers which shouldn't really be the case

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkIcons.kt
```diff
@@ -60,9 +59,6 @@ data class BpkIcon(
               }
               .filterValues { it != null }
               .let { it as Map<Type, String> },
-            autoMirror = files
-              .filter { it.nameWithoutExtension.startsWith(name) }
-              .any { it.readText().contains("android:automirrored=\"true\"", ignoreCase = true) },
```
marianeum, line 65: the icons are already automirrored - we removed the logic to auto mirror them again based on this prop (in the past auto mirroring wasn't applied correctly in compose) - this removes the generation part as well

frugoman (approved):

PR #1400, opened 2023-01-10
KOA-5678 Automate screenshots recording

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


app/src/screenshots/res/values/themes.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <style name="AppThemeBase" parent="BpkTheme">
+    <item name="bpkFontFamilyBase">@font/skyscanner_relative_android_book</item>
+    <item name="bpkFontFamilyEmphasized">@font/skyscanner_relative_android_bold</item>
+    <item name="android:lineSpacingMultiplier">1</item>
+    <item name="android:overScrollMode">never</item>
```
bvitaliyg, line 7: This is the most important stuff for screenshots consistency

buildSrc/build.gradle.kts
```diff
@@ -14,4 +14,10 @@ dependencies {
   implementation("com.squareup:kotlinpoet:1.12.0")
   implementation("com.google.guava:guava:31.1-jre")
   implementation("com.google.code.gson:gson:2.10")
+
+  implementation(platform("org.http4k:http4k-bom:4.35.3.0"))
```
bvitaliyg, line 18: For the server

buildSrc/build.gradle.kts
```diff
@@ -14,4 +14,10 @@ dependencies {
...
+  implementation("org.http4k:http4k-core")
+  implementation("org.http4k:http4k-server-undertow")
+  implementation("org.http4k:http4k-client-apache")
+  implementation("dev.mobile:dadb:1.2.6")
```
bvitaliyg, line 22: Adb will not always be available in terminal – we use a library here instead

buildSrc/src/main/kotlin/net/skyscanner/backpack/screenshots/ScreenshotTestsServer.kt
```diff
@@ -0,0 +1,95 @@
...
+      val outFile = File(folder, "$file.png")
+      if (!outFile.exists()) {
+        outFile.createNewFile()
+      }
+
+      // 20ms delay to make sure the buttons are unpressed and states is updated
+      Thread.sleep(20L)
+
+      adb.requireShell("screencap -p $tmpRemoteFile")
+      adb.pull(outFile, tmpRemoteFile)
```
bvitaliyg, line 69: This is actually faster the exec-out, which is not supported by the library

buildSrc/src/main/kotlin/net/skyscanner/backpack/screenshots/ScreenshotTestsServer.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+  private fun Dadb.setup(): Unit {
+    requireShell("settings put global sysui_demo_allowed 1")
+    requireShell("settings put global window_animation_scale 0.0")
+    requireShell("settings put global transition_animation_scale 0")
+    requireShell("settings put global animator_duration_scale 0")
+    requireShell("am broadcast -a com.android.systemui.demo -e command clock -e hhmm 1000")
+    requireShell("am broadcast -a com.android.systemui.demo -e command battery -e plugged false -e level 100")
+    requireShell("am broadcast -a com.android.systemui.demo -e command network -e wifi hide -e level 4 -e mobile show -e datatype none")
```
bvitaliyg, line 85: Wifi was glitchy. We won't show it anymore, but we may add Lte for mobile

app/build.gradle
```diff
@@ -67,19 +70,26 @@ android {
       dimension "version"
       versionNameSuffix "-internal"
     }
+    screenshots {
```
bvitaliyg, line 73: I'm also going to add task which will copy fonts automatically from internal to screnshots

app/src/screenshots/res/values/backpack.animation.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<integer name="bpkAnimationDurationXs">0</integer>
```
bvitaliyg, line 3: But button pressed animations

app/build.gradle
```diff
@@ -94,6 +104,29 @@ android {
...
+    server.close()
+  }
+}
+
+tasks.create("recordScreenshots") {
+  dependsOn(
+    "startScreenshotsServer",
+    "PixelScreenshotsDebugAndroidTest",
+  )
+  finalizedBy("stopScreenshotsServer")
```
bvitaliyg, line 127: Finalisers are not transitive in Gradle

app/src/main/java/net/skyscanner/backpack/demo/stories/DialogStory.kt
```diff
@@ -190,7 +190,7 @@ class DialogStory : Story() {
     BpkDialog(requireContext(), BpkDialog.Type.Flare).apply {
       title = getString(R.string.dialog_title)
       description = getString(R.string.dialog_text)
-      Picasso.get().load("file:///android_asset/dialog_sample.jpg").into(image)
+      Picasso.get().load("file:///android_asset/dialog_sample.jpg").noFade().into(image)
```
bvitaliyg, line 193: This was causing inconsistency in screenshot

app/src/screenshots/java/net/skyscanner/backpack/demo/BackpackDemoTheme.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.compose.ui.text.font.Font
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.font.FontWeight
+import net.skyscanner.backpack.compose.theme.BpkTheme
+
+@Composable
+fun BackpackDemoTheme(
+  content: @Composable () -> Unit,
+) {
+  BpkTheme(
```
bvitaliyg, line 31: The same as in internal

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -61,9 +61,6 @@ object DocsRegistry {
     ComposeScreenshot("Button - Compose - Large", "large"),
     ComposeScreenshot("Button - Compose - Link", "link"),
     ComposeScreenshot("Button - Compose - Drawable Icon", "drawable-icon"),
-    ViewScreenshot("Calendar - Default", "range") { setupCalendar() },
```
bvitaliyg, line 64: We don't show it on website anymore, and it was causing inconsistency

app/build.gradle
```diff
@@ -94,6 +104,29 @@ android {
   }
 }
 
+def server = new ScreenshotTestsServer(rootProject.file("docs"))
```
bvitaliyg, line 107: I'm going to extract it to another gradle file

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -183,8 +183,6 @@ object ComponentRegistry {
         TAB_TITLE_COMPOSE composeStory { net.skyscanner.backpack.demo.compose.BarChartStory() },
       ),
     ),
-    "Bottom Nav" story NodeData { BottomNavStory of R.layout.fragment_bottom_nav },
```
bvitaliyg, line 186: A merge error was causing the screenshots to fail. Also, the screenshots were taken on the wrong emulator (you can see it in the diff).

buildSrc/src/main/kotlin/net/skyscanner/backpack/screenshots/ScreenshotTestsServer.kt
```diff
@@ -0,0 +1,95 @@
...
+    try {
+      server.start()
+    } catch (t: BindException) {
+      // port is already taken: gradle didn't finalise the previous run properly
+      // if we just send an empty ping request, the server will close itself due the request being invalid
+      JavaHttpClient().invoke(Request(Method.GET, "http://localhost:$serverPort")).close()
+      // slleping for 3 seconds to make sure the server is closed
+      Thread.sleep(3000L)
+      // another attempt
+      require(Dadb.list().isEmpty()) { "No device or emulator should be running before the tests" }
```
bvitaliyg, line 35: I don't want to use recursion here deliberately

app/src/screenshots/res/values/themes.xml
marianeum: This was consistent before this change by using the `waitForIdleSync` for the nav bar. I still think something's not right here with how things are set up that cause these issues to happen as that used to be an issue in the past, but is no longer happening.

app/build.gradle
```diff
@@ -94,6 +104,29 @@ android {
...
+  finalizedBy("stopScreenshotsServer")
+}
+
+tasks.create("stopScreenshotsServer") {
+  doLast {
+    server.close()
+  }
+}
+
+tasks.create("recordScreenshots") {
```
marianeum, line 122: please don't forget to update the contributing guide

app/build.gradle
```diff
@@ -67,19 +70,26 @@ android {
       dimension "version"
       versionNameSuffix "-internal"
     }
+    screenshots {
+      dimension "version"
+      versionNameSuffix "-screenshots"
+      testInstrumentationRunnerArgument "notClass", "net.skyscanner.backpack.*"
```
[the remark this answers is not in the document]
bvitaliyg: There's no way to provide arguments when you running tests using gradle-managed device for now

buildSrc/src/main/kotlin/net/skyscanner/backpack/screenshots/ScreenshotTestsServer.kt
```diff
@@ -0,0 +1,95 @@
...
+
+      val outFile = File(folder, "$file.png")
+      if (!outFile.exists()) {
+        outFile.createNewFile()
+      }
+
+      // 20ms delay to make sure the buttons are unpressed and states is updated
+      Thread.sleep(20L)
+
+      adb.requireShell("screencap -p $tmpRemoteFile")
```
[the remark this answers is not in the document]
bvitaliyg: No, we call it whenever we init adb – it's lazy

app/build.gradle
bvitaliyg: Will do

app/src/screenshots/res/values/backpack.animation.xml
bvitaliyg: It does. A lot of APIs ignore that flag.
In our case, the cause is in ButtonStyle.getButtonBackground() method when we manually set enter and exit fade durations from these tokens. It seems it only respects the system values when defaults are not overridden.

I tested this on an emulator with disabled animations and I can confirm that the pressed animation still exists there, unless we override the tokens or don't use it at all.

app/src/screenshots/res/values/themes.xml
bvitaliyg: It was not. The screenshots were changing all the time.

`waitForIdleSync()` only waited until the scrolling was done – but not the overscroll effect. The amount of overscroll effect (which depends on time, velocity, and direction) was always slightly different and usually, it was only the pixels at the overscroll effect border which was slightly different.

app/src/main/java/net/skyscanner/backpack/demo/stories/DialogStory.kt
bvitaliyg: I tested it manually on an emulator with completely disabled animations – it still appears with fade.

app/build.gradle
bvitaliyg: Done

app/build.gradle
bvitaliyg: Done

app/build.gradle
bvitaliyg: Done

frugoman (approved):

PR #1414, opened 2023-01-17
Fix flaky bottomsheet test

marianeum:
Bottom sheet tests were both flaky and complex. I also think the current implementation wasn't working correctly if the state was set initially. This change should fix both of those issues.

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/bottomsheet/BpkBottomSheetBehaviour.kt
```diff
@@ -61,6 +61,23 @@ class BpkBottomSheetBehaviour<V : View>(context: Context, attrs: AttributeSet? =
     )
   }
 
+  override fun onLayoutChild(parent: CoordinatorLayout, child: V, layoutDirection: Int): Boolean {
+    return super.onLayoutChild(parent, child, layoutDirection).apply { child.background = background }
```
marianeum, line 65: the background was only set when the state changed, meaning if the state was set initially it might not be applied correctly. This fixes that

Backpack/src/main/java/net/skyscanner/backpack/bottomsheet/BpkBottomSheetBehaviour.kt
```diff
@@ -61,6 +61,23 @@ class BpkBottomSheetBehaviour<V : View>(context: Context, attrs: AttributeSet? =
...
+  }
+
+  override fun setState(state: Int) {
```
marianeum, line 68: similarly when the state was set before layout the background wasn't set correctly - this fixes that issue

app/src/androidTest/java/net/skyscanner/backpack/bottomsheet/BpkBottomSheetTest.kt
```diff
@@ -69,27 +66,12 @@ class BpkBottomSheetTest : BpkSnapshotTest() {
...
-    }
-    val callback = Callback(bottomSheetBehaviour)
-    IdlingRegistry.getInstance().register(callback)
-
-    InstrumentationRegistry.getInstrumentation().waitForIdleSync()
-    var bottomSheet: View? = null
-    Espresso.onView(withId(TEST_ID)).check { view, _ -> bottomSheet = view }
-    snap(bottomSheet!!, width = 200, height = 200, padding = 0)
-    IdlingRegistry.getInstance().unregister(callback)
+    val root = setupBottomSheet(state)
```
marianeum, line 69: makes the tests much simpler

PR #1420, opened 2023-01-20
KOA-5828 Remove static colours in compose

marianeum:
Colours are only accessed semantically in compose - we can clean up the static colours to avoid accidental usage. Also fixed two tokens being accidentally treated as dynamic, although they were static.

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkColors.kt
```diff
@@ -64,10 +64,6 @@ public class BpkColors private constructor(
   public val surfaceDefault: Color,
   public val surfaceElevated: Color,
   public val surfaceHighlight: Color,
-  @Deprecated(DEPRECATION_MESSAGE)
```
marianeum, line 67: these are actually old, static colours that were treated as dynamic due to the detection for what's a dynamic token was wrong

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkColors.kt
```diff
@@ -48,7 +48,7 @@ object BpkColor {
 
     override fun invoke(source: Map<String, Any>): BpkColors =
       parseColors(source, resolveReferences = false) {
-        it.name == it.defaultReference && !it.isMarcomms()
+        !it.isSemanticColor() && !it.isMarcomms()
```
marianeum, line 51: this is where the bug was

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkColors.kt
bvitaliyg: I think we need to clean up the foundations and remove all these custom tricks eventually

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkColors.kt
```diff
@@ -178,29 +173,6 @@ private fun deprecationProperty(): PropertySpec =
 @OptIn(ExperimentalStdlibApi::class)
 private fun String.toHexColor() = uppercase().run { "0x${substring(7)}${substring(1, 7)}" }
 
-private fun toStaticCompose(
```
bvitaliyg, line 181: So, we're not generating static compose colors at all, right?

bvitaliyg (approved):
Some quesntions

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkColors.kt
marianeum: nope - we're meant to only use semantic ones anyways. they're also not used in the app at all. with the view system we've got historic colour usage, but in compose we've started from scratch

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/tokens/BpkColors.kt
marianeum: eventually. but right now we still have usages of these (in the view system)

PR #1426, opened 2023-01-23
Koa-4932 Create BpkStarRating as Compose

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/starrating/BpkStarRating.kt
```diff
@@ -0,0 +1,151 @@
...
+    },
+    modifier = modifier,
+    rating = rating,
+    contentDescription = contentDescription,
+  )
+}
+
+@Composable
+fun BpkHotelRating(
+  rating: Float,
```
marianeum, line 65: as the hotels version only supports full stars should this be an integer? otherwise it might look like it supports half stars

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,35 @@
...
+Backpack Compose is available through [Maven Central](https://search.maven.org/artifact/net.skyscanner.backpack/backpack-compose). Check the main [Readme](https://github.com/skyscanner/backpack-android#installation) for a complete installation guide.
+
+## Usage
+
+Example of a BpkStarRating with full star rating:
+
+```Kotlin
+import net.skyscanner.backpack.compose.starrating.BpkStarRating
+
+BpkStarRating (rating = 2f, contentDescription = { _, _ -> "" })
```
marianeum, line 24: I think this needs to be updated to the latest API for the content description
We should also have an example for rounding + size + the hotels component

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,175 @@
...
+  }
+}
+
+@Composable
+private fun HotelRatingSample(context: Context, size: BpkStarRatingSize, text: String) {
+  BpkText(
+    text = text,
+    style = BpkTheme.typography.heading3,
+  )
+  BpkHotelRating(
```
marianeum, line 106: a small improvement here might be to reduce code duplication and loop from 1 to 5 and create the component like that, rather than duplicating this 5 times

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,175 @@
...
+  )
+}
+
+@Composable
+private fun StaticRatingSample(context: Context, size: BpkStarRatingSize, text: String) {
+  BpkText(
+    text = text,
+    style = BpkTheme.typography.heading3,
+  )
+  BpkStarRating(
```
marianeum, line 139: similarly here you could have a float list of the ratings you want to display and loop through them to avoid repetition

app/src/androidTest/java/net/skyscanner/backpack/compose/starrating/BpkStarRatingTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import net.skyscanner.backpack.compose.BpkSnapshotTest
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class BpkStarRatingTest : BpkSnapshotTest() {
+
+  @Test
+  fun ratingFullStar() = snap {
```
marianeum, line 31: can you include some tests for the rounding, size + hotels one as well? might be worth looking at the view version to see what is tested there

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -140,8 +140,10 @@ object DocsRegistry {
     ViewScreenshot("Snackbar", "icon") { setupSnackbarIconAction() },
     ViewScreenshot("Spinner - View", "default"),
     ComposeScreenshot("Spinner - Compose", "default"),
-    ViewScreenshot("Star Rating - Default", "default"),
+    ViewScreenshot("Star Rating - View - Default", "default"),
+    ComposeScreenshot("Star Rating - Compose", "default"),
     ViewScreenshot("Star Rating Interactive", "default"),
+    ComposeScreenshot("Star Rating - Compose", "default"),
```
marianeum, line 146: Looks like this one's duplicated

marianeum (commented):
looking good - just a couple of minor comments. one more thing: could you also update the star rating example in `RatingStory` that currently uses the view version

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,175 @@
...
+        )
+        Row() {
+          Column(modifier = Modifier.padding(end = BpkSpacing.Xl)) {
+            HotelRatingSample(
+              context = context,
+              size = BpkStarRatingSize.Large,
+              text = stringResource(R.string.icons_large),
+            )
+          }
+          Column() {
```
bvitaliyg, line 87: We don't need to have the brackets if there's no arguments in it and it's followed by trailing lambda

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,175 @@
...
+    size = size,
+  )
+  BpkHotelRating(
+    rating = 2f,
+    contentDescription = context.getString(R.string.star_rating_accessibility_status, 2f, 5),
+    size = size,
+  )
+  BpkHotelRating(
+    rating = 3f,
+    contentDescription = context.getString(R.string.star_rating_accessibility_status, 3f, 5),
```
bvitaliyg, line 118: What happened with the lambda for contentDescription? It seems we still duplicating the way we calculate the logic here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/starrating/BpkStarRating.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+    StarType.Half -> {
+      BpkIcon(
+        icon = BpkIcon.StarHalf,
+        contentDescription = null,
+        size = iconSize,
+        tint = BpkTheme.colors.statusWarningSpot,
+      )
+    }
+    StarType.Full -> {
```
bvitaliyg, line 130: No need to have brackets here as well

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/starrating/BpkStarRating.kt
```diff
@@ -0,0 +1,151 @@
...
+  val roundedRating = when (rounding) {
+    RoundingType.Down -> floor(coercedRating * 2) / 2
+    RoundingType.Up -> ceil(coercedRating * 2) / 2
+    RoundingType.Nearest -> round(coercedRating * 2) / 2
+  }
+  Row(modifier = modifier.semantics { this.contentDescription = contentDescription }) {
+
+    for (item in 0 until maxRating) {
+      val value = (roundedRating - item).coerceIn(0f, 1f)
+      when {
```
bvitaliyg, line 102: This can be a part of when statement: ```when (val value = (roundedRating - item).coerceIn(0f, 1f)) {```

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
jaanainkinen: I think @marianeum suggested a lambda was not necessary here and we came to this solution. By duplicating do you mean setting the rating in the rating attribute but also setting it inside the contentDescription?

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+@Composable
+private fun HotelRatingSample(context: Context, size: BpkStarRatingSize, text: String) {
+  BpkText(
+    text = text,
+    style = BpkTheme.typography.heading3,
+  )
+  var count = 1
```
marianeum, line 106: rather than having a separate `count` property you can use (or rename the `i` in the for loop as you're already looping from 1 to 5

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Default
+
+| Day                                                                                                                                                                  | Night                                                                                                                                                                               |
+|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/StarRating/screenshots/all.png" alt="Star Rating component" width="375" /> | <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/StarRating/screenshots/all_dm.png" alt="Star Rating component - dark mode" width="375" /> |
```
marianeum, line 11: i think the `all` needs to be renamed as the screenshot name is `default` (same for the dm one)

marianeum (approved):
two tiny things

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
jaanainkinen: Haha not sure why I complicated it like that. Should be fixed now!

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Example of a `BpkStarRating` with half star rating:
+
+```Kotlin
+import net.skyscanner.backpack.compose.starrating.BpkStarRating
+
+BpkStarRating(
+    rating = 3.5f,
+    contentDescription = "",
```
bvitaliyg, line 39: I think we need to show how to use content description exactly in these samples

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
bvitaliyg: Yes exactly. We can use string placeholders instead of lambda here, as we do with View version

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,127 @@
...
+  Box(
+    modifier = modifier.fillMaxHeight(),
+    contentAlignment = Alignment.Center,
+  ) {
+    Column(modifier = Modifier.padding(start = BpkSpacing.Lg)) {
+      BpkText(
+        text = stringResource(R.string.static_default_rating_title),
+        style = BpkTheme.typography.heading2,
+      )
+      Row() {
```
bvitaliyg, line 54: You still need to remove these brackets

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,127 @@
...
+          )
+        }
+        Column() {
+          StaticRatingSample(
+            size = BpkStarRatingSize.Small,
+            text = stringResource(R.string.icons_small),
+          )
+        }
+      }
+      Spacer(modifier = Modifier.padding(top = BpkSpacing.Lg))
```
bvitaliyg, line 68: Instead of spaces, you can use arrangement for the column

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,127 @@
...
+      contentDescription = { value, max ->
+        stringResource(R.string.star_rating_accessibility_status, value, max)
+      },
+      size = size,
+    )
+  }
+}
+
+@Composable
+private fun StaticRatingSample(size: BpkStarRatingSize, text: String) {
```
bvitaliyg, line 111: Instead of wrapping each call of this function in Column, please use Column here inside (same can be allied to HotelRatingSample). Also, please provide modifiers to these functions

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/starrating/BpkStarRating.kt
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+private enum class StarType {
+  Empty,
+  Half,
+  Full,
+}
+
+enum class RoundingType {
```
bvitaliyg, line 143: Can we have this enum in the beginning of the file, in the same way, we do for Size? Also, let's keep the naming consistent: BpkRatingRounding

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```Kotlin
+import net.skyscanner.backpack.compose.starrating.BpkStarRating
+
+BpkStarRating(
+    rating = 2f,
+    contentDescription = { value, max ->
+      context.getString(R.string.star_rating_accessibility_status, value, max)
+    },
+    rounding = RoundingType.Up,
```
bvitaliyg, line 29: Don't forget to update the name in the samples later. Also, it's not context.getString() now, but stringResource() instead

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,61 @@
...
+)
+```
+
+Example of a `BpkHotelRating`:
+
+```Kotlin
+import net.skyscanner.backpack.compose.starrating.BpkStarRating
+
+BpkHotelRating(
+    rating = 3,
```
bvitaliyg, line 55: 3f

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,61 @@
...
+    contentDescription = { value, max ->
+      context.getString(R.string.star_rating_accessibility_status, value, max)
+    },
+    size = BpkStarRatingSize.Small,
```
bvitaliyg, line 59: BpkStarRatingSize also needs to be imported

docs/compose/StarRating/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```Kotlin
+import net.skyscanner.backpack.compose.starrating.BpkStarRating
+
+BpkStarRating(
+    rating = 3.5f,
+    contentDescription = { value, max ->
+      context.getString(R.string.star_rating_accessibility_status, value, max)
+    },
+    rounding = RoundingType.Down,
```
bvitaliyg, line 44: RoundingType and BpkStartRating size needs to be imported

docs/compose/StarRating/README.md
marianeum: it's an integer for hotels :)

bvitaliyg (commented):
Few minor comments, otherwise it's good

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+      }
+      Spacer(modifier = Modifier.padding(top = BpkSpacing.Lg))
+      Column {
+        BpkText(
+          text = stringResource(R.string.hotel_star_rating_title),
+          style = BpkTheme.typography.heading2,
+        )
+        Row {
+          Column(modifier = Modifier.padding(end = BpkSpacing.Xl)) {
```
bvitaliyg, line 75: We can also use the arrangement for Row here

app/src/main/java/net/skyscanner/backpack/demo/compose/BpkStarRatingStory.kt
```diff
@@ -0,0 +1,127 @@
...
+
+@Composable
+fun BpkStarRatingStory(
+  modifier: Modifier = Modifier,
+) {
+  Box(
+    modifier = modifier.fillMaxHeight(),
+    contentAlignment = Alignment.Center,
+  ) {
+    Column(modifier = Modifier.padding(start = BpkSpacing.Lg)) {
```
bvitaliyg, line 49: Why do we need this padding in the first place?

jaanainkinen: I added this to bring the whole column a bit to the right, so it's not glued to the left edge.

docs/compose/StarRating/README.md
jaanainkinen: Is this not meant to be 3, because it's an Int? I changed it to Int in the case of BpkHotelRating because Maria suggested that having it as Float might give the impression that this component also supports half stars (Maria's comment further above).

bvitaliyg (approved):
💯

PR #1429, opened 2023-01-25
KOA-5824 Move icon generation to Kotlin

marianeum:
Moved icon generation to Kotlin - our last token generation type 🎉 
Best to review with excluding xml + png files
The icons changed - I noticed that the path positions have an extra decimal in some cases, but otherwise are the same. Looking through the snapshots there are no visual differences (none that you can notice as a human), so that matches the difference in accuracy and is expected when changing tools.

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/tokens.gradle.kts
```diff
@@ -167,12 +167,14 @@ tasks {
 
   val generateIcons by creating {
     this.group = group
-    project.androidFileOf("backpack-common", "src/main/res/drawable-nodpi")
-      .readAs(BpkFormat.Folder)
-      .parseAs(BpkIcon.Parser)
-      .transformTo(BpkIcon.Format.Compose(rClass))
-      .saveTo(BpkOutput.KotlinExtensionFile(src, tokensPackage, "BpkIcon"))
-      .execute()
+    doLast {
```
marianeum, line 170: this was missing - i noticed it when my task was failing on CI due to node not being installed on most jobs and noticed it was missing here too

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkIcons.kt
```diff
@@ -147,5 +176,30 @@ private fun toCompose(
...
+        val suffix = if (type == BpkIcon.Type.Sm) "_sm" else ""
+        "bpk_$it$suffix"
+      }
+
+  fun BpkIcon.fileContent(metadata: Map<String, String?>) =
+    value
+      .replace("android:fillColor=\"#FF000000\"", "android:fillColor=\"@color/bpkTextPrimary\"")
+      .let {
+        if (metadata.containsKey(name) && metadata[name] == "true") {
+          it.replaceFirst("\n", "\n    android:autoMirrored=\"true\"\n")
```
marianeum, line 195: we need to add the auto mirror property to the vector. This was the cleanest way of doing it without adding an xml parser

frugoman (approved):
🚀 🚀 🚀

PR #1431, opened 2023-01-27
Bump Kotlin & Compose

marianeum:
For some reason dependabot didn't raise these, so bumped manually (after having done in the Skyscanner app).
Also moved to the bom as there's various point releases on some of the dependencies, so this ensures we always stay in sync.

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,10 +17,10 @@
  */
 
 buildscript {
-  ext.kotlin_version = '1.7.20'
-  ext.compose_compiler_version = '1.3.2'
+  ext.kotlin_version = '1.8.0'
+  ext.compose_compiler_version = '1.4.0'
   ext.detekt_version = '1.22.0'
-  ext.dokka = "org.jetbrains.dokka:dokka-gradle-plugin:$kotlin_version"
+  ext.dokka = "org.jetbrains.dokka:dokka-gradle-plugin:1.7.20"
   repositories {
     google()
     mavenCentral()
--- a/build.gradle
+++ b/build.gradle
@@ -120,18 +120,19 @@ ext {
   lintApi = "com.android.tools.lint:lint-api:$lintVersion"
   lintTests = "com.android.tools.lint:lint-tests:$lintVersion"
 
-  composeVersion = "1.3.1"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -17,10 +17,10 @@
  */
 
 buildscript {
-  ext.kotlin_version = '1.7.20'
-  ext.compose_compiler_version = '1.3.2'
+  ext.kotlin_version = '1.8.0'
+  ext.compose_compiler_version = '1.4.0'
   ext.detekt_version = '1.22.0'
-  ext.dokka = "org.jetbrains.dokka:dokka-gradle-plugin:$kotlin_version"
+  ext.dokka = "org.jetbrains.dokka:dokka-gradle-plugin:1.7.20"
```
marianeum, line 23: looks like dokka is only somewhat related to the kotlin releases. there's no 1.8 version out there yet

build.gradle
```diff
@@ -120,18 +120,19 @@ ext {
...
-  composeUi = "androidx.compose.ui:ui:$composeVersion"
-  composeFoundation = "androidx.compose.foundation:foundation:$composeVersion"
-  composeRuntime = "androidx.compose.runtime:runtime:$composeVersion"
-  composeMaterial = "androidx.compose.material:material:$composeVersion"
-  composeUiToolingPreview = "androidx.compose.ui:ui-tooling-preview:$composeVersion"
-  composeTestJUnit = "androidx.compose.ui:ui-test-junit4:$composeVersion"
-  composeUiTooling = "androidx.compose.ui:ui-tooling:$composeVersion"
-  composeUiTestManifest = "androidx.compose.ui:ui-test-manifest:$composeVersion"
+  composeVersion = "2023.01.00"
+  composeBom = dependencies.platform("androidx.compose:compose-bom:$composeVersion")
```
marianeum, line 124: switched to the bom, so we can keep all the point releases in sync with the latest version, rather than having to define the version separately for each of the dependencies

frugoman: nice one!

frugoman (approved):

PR #1435, opened 2023-01-30
KOA-5874 Fix chip crash when using theming

marianeum:
This crash happened when programmatically setting a style + using theming due to recycling the `TypedArray` twice.

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
@@ -56,12 +56,14 @@ internal sealed class BpkChipAppearances : BpkChipAppearance {
         disabled = context.getColor(R.color.bpkTextDisabled),
       )
 
-    internal class Default(context: Context, typedArray: TypedArray?) : Solid(context) {
+    internal class Default(context: Context, private val typedArray: TypedArray?) : Solid(context) {
       constructor(
         context: Context,
         attrs: AttributeSet?,
         defStyleAttr: Int,
-      ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0))
+      ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0)) {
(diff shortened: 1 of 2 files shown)


Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
```diff
@@ -56,12 +56,14 @@ internal sealed class BpkChipAppearances : BpkChipAppearance {
...
 
-    internal class Default(context: Context, typedArray: TypedArray?) : Solid(context) {
+    internal class Default(context: Context, private val typedArray: TypedArray?) : Solid(context) {
       constructor(
         context: Context,
         attrs: AttributeSet?,
         defStyleAttr: Int,
-      ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0))
+      ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0)) {
+        typedArray?.recycle()
```
marianeum, line 65: recycling it here where we crate it, rather than within the function. otherwise the passed in typed array might be recycled twice

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
```diff
@@ -80,17 +82,18 @@ internal sealed class BpkChipAppearances : BpkChipAppearance {
...
     }
 
-    internal class OnDark(context: Context, typedArray: TypedArray?) : Solid(context) {
+    internal class OnDark(context: Context, private val typedArray: TypedArray?) : Solid(context) {
       constructor(
         context: Context,
         attrs: AttributeSet?,
         defStyleAttr: Int,
-      ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0))
+      ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0)) {
```
marianeum, line 94: same here

olliecurtis (approved):

PR #1436, opened 2023-01-30
SONIC-1362 Fix BpkTopNavBar scrollable elevation

xo0ps:
[SONIC-1362](https://gojira.skyscanner.net/browse/SONIC-1362)

This PR uses `Modifier.zIndex()` to bring toolbar to the front of all components in the screen.
I added a `BpkCard` to the list in sample to demonstrate how it works when there are other components with elevation in the screen.

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/navigationbar/internal/BpkTopNavBarImpl.kt
```diff
@@ -64,7 +65,7 @@ internal fun BpkTopNavBarImpl(
     contentColor = BpkTheme.colors.textPrimary,
     elevation = animateDpAsState(targetValue = if (fraction == 0f) BpkDimension.Elevation.Sm else 0.dp).value,
     shape = RectangleShape,
-    modifier = modifier,
+    modifier = modifier.zIndex(Float.MAX_VALUE),
```
marianeum, line 68: only concern with using `max` here is there might be cases where we DO want to show something on top of a toolbar (can think of a dropdown menu for example from the nav bar unless they're drawn in a separate window)

marianeum: Let's use 1 for now - we can always make it configurable/change in case it's needed

marianeum (approved):

PR #1440, opened 2023-02-01
KOA-5796 Update chip component to latest designs

marianeum:
This comes with a bunch of breaking changes:
- In compose the types are now separate composable function to align with the `dismissible` state not being toggleable/disableable.
- In the view system the `disabled` property was removed in favour of the system one to avoid bugs/inconsistencies between the two flags
- In both the `selectable` type was replaced with the `dropdown` type

+ [x] Component `README.md`
+ [x] Tests


Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
```diff
@@ -27,126 +27,194 @@ import android.util.AttributeSet
...
+      ),
+      border = chipColors(
+        selected = Color.TRANSPARENT,
+        default = strokeColor,
+        pressed = pressedStrokeColor,
+        disabled = Color.TRANSPARENT,
+      ),
+    )
+
+  override val dismissibleBackground: Drawable
```
marianeum, line 53: a new dismissible background as it behaves a bit different

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
```diff
@@ -27,126 +27,194 @@ import android.util.AttributeSet
...
+
+  override val text: ColorStateList
+    get() = chipColors(
+      selected = selectedTextColor,
+      default = textColor,
+      pressed = pressedTextColor,
+      disabled = context.getColor(R.color.bpkTextDisabled),
+    )
+
+  override val dismissibleText: ColorStateList
```
marianeum, line 73: new prop here as well

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -80,8 +80,10 @@ object DocsRegistry {
     ComposeScreenshot("Card Button - Small", "small"),
     ViewScreenshot("Chip - View - Default", "all"),
     ViewScreenshot("Chip - View - On Dark", "on-dark"),
-    ViewScreenshot("Chip - View - With icon", "with-icon"),
-    ComposeScreenshot("Chip - Compose", "default"),
+    ViewScreenshot("Chip - View - On Image", "on-image"),
+    ComposeScreenshot("Chip - Compose - Default", "default"),
```
marianeum, line 84: had to split this out into multiple screens as it would no longer fit

Backpack/src/main/java/net/skyscanner/backpack/chip/BpkChip.kt
```diff
@@ -105,59 +92,85 @@ open class BpkChip @JvmOverloads constructor(
...
 
   internal open fun provideAppearance(context: Context, attrs: AttributeSet?, defStyleAttr: Int): BpkChipAppearance =
-    BpkChipAppearances.Solid.fromAttrs(context, attrs, defStyleAttr)
+    BpkChipAppearances.fromAttrs(context, attrs, defStyleAttr)
 
   private fun updateStyle() {
-    this.background = appearance.background
-    setTextColor(appearance.text)
-    compoundDrawableTintList = appearance.text
+    elevation = if (style == Style.OnImage) resources.getDimension(R.dimen.bpkElevationSm) else 0f
```
olliecurtis, line 104: Silly non Android dev question is `0f` a thing or is this something else?

olliecurtis (approved):

Backpack/src/main/java/net/skyscanner/backpack/chip/BpkChip.kt
olliecurtis: Spoken to Maria in person, yes its a thing :) and not a mistake it means 'float' for anyone not familiar with it!

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
```diff
@@ -27,126 +27,194 @@ import android.util.AttributeSet
...
+  internal class OnDark(context: Context, private val typedArray: TypedArray?) : BpkChipAppearances(context) {
+    constructor(
+      context: Context,
+      attrs: AttributeSet?,
+      defStyleAttr: Int,
+    ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0)) {
+      typedArray?.recycle()
+    }
+
+    override var selectedBackgroundColor: Int = context.getColor(R.color.__privateChipOnDarkOnBackground)
```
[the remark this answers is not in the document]
frugoman: would it make sense for this properties to be `val` instead of var? i dont think we're mutating them

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
```diff
@@ -27,126 +27,194 @@ import android.util.AttributeSet
...
+  internal class OnImage(context: Context, private val typedArray: TypedArray?) : BpkChipAppearances(context) {
+    constructor(
+      context: Context,
+      attrs: AttributeSet?,
+      defStyleAttr: Int,
+    ) : this(context, context.theme.obtainStyledAttributes(attrs, R.styleable.BpkChip, defStyleAttr, 0)) {
+      typedArray?.recycle()
+    }
+
+    override var selectedBackgroundColor: Int = context.getColor(R.color.bpkCorePrimary)
```
frugoman, line 162: same question as above

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chip/BpkChip.kt
```diff
@@ -67,120 +44,63 @@ fun BpkChip(
...
+    enabled = true,
+    style = style,
+    icon = icon,
+    type = BpkChipType.Dismiss,
+    interactionSource = interactionSource,
+    modifier = modifier.applyIf(onClick != null) {
+      clickable(
+        interactionSource = interactionSource,
+        indication = LocalIndication.current,
+      ) { onClick!!.invoke() }
```
frugoman, line 103: could this be replaced by 
```kt
modifier = onClick?.let {
   // ... clickable modifier
   // { onClick?.let { it.invoke()) }
}
```

not sure if this works in compose due to how contexts work here

Backpack/src/main/java/net/skyscanner/backpack/chip/internal/BpkChipAppearances.kt
marianeum: same here

bvitaliyg (approved):

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chip/BpkChip.kt
marianeum: not quite! the modifier will need to be applied always - the `applyIf` _might_ be wrongly named, though as it is not the modifier that is not being applied if the condition matches, but the `clickable` part on the modifier

PR #1466, opened 2023-02-13
BD-6667: Added attribute elevation to BpkDividedCard component

jacobhuang0121:
https://gojira.skyscanner.net/browse/BD-6667

The new attribute elevation is added in the BpkDividedCard component, which is used to control whether the BpkDividedCard component needs shadow, and the default value is shadow.

|Light mode|Dark mode|
|-----|------|
|Screenshot_20230213_153527|Screenshot_20230213_153553|

+ [ ] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/dividedcard/BpkDividedCard.kt
```diff
@@ -39,13 +40,15 @@ fun BpkDividedCard(
   secondaryContent: @Composable () -> Unit,
   modifier: Modifier = Modifier,
   onClick: (() -> Unit)? = null,
+  isElevated: BpkCardElevation = BpkCardElevation.Default,
```
marianeum, line 43: rather than `isElevated` can you name this `elevation` to be consistent with the rest of the component please?

marianeum:
From a label perspective this is a minor change as there's a default property, so it's not a breaking change

marianeum (approved):

PR #1468, opened 2023-02-14
Skyscanner/sonic 1203 bpk carousel

mineschan:
https://gojira.skyscanner.net/browse/SONIC-1203
|Light mode|Dark mode|
|-----|------|
|default|dm|

+ [x] Component `README.md`
+ [x] Tests


app/src/androidTest/java/net/skyscanner/backpack/compose/carousel/BpkCarouselTest.kt
```diff
@@ -0,0 +1,84 @@
...
+  fun multipleImageDarkMode() = snap {
+    BpkCarousel(
+      count = 2,
+      imageContent = {
+        Image(painter = painterResource(id = imageResAtIndex(it)), contentDescription = "")
+      }
+    )
+  }
+
+  private fun imageResAtIndex(index: Int) = when (index) {
```
marianeum, line 77: can you also use a different image for the tabs here, so the `setCurrentImage` test will ensure the image is updated?

app/src/androidTest/java/net/skyscanner/backpack/compose/carousel/BpkCarouselTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    0 -> net.skyscanner.backpack.demo.R.drawable.canadian_rockies_canada
```
marianeum, line 78: can you add an import for the R file please?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+      LaunchedEffect(it) {
+        snapshotFlow { it.currentPage }.distinctUntilChanged().collect { index ->
+          val page = getModNumber(index, count)
+          setSelectedPage(page)
+          onImageChanged?.invoke(page)
+        }
+      }
+    }
+
+    HorizontalPager(
```
marianeum, line 61: does this handle accessibility internally? ie allow switching between pages, announce the page change, count etc? how does it handle the `max_value` case?

marianeum:
Did you use `--no-verify` or did the pre-commit hooks not install correctly? 🤔 (follow up if not: did you have a recent main base?)

app/src/androidTest/java/net/skyscanner/backpack/compose/carousel/BpkCarouselTest.kt
mineschan: Good catch! I was supposed to update them but left it behind.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
mineschan: Unfortunately no, as currently the talkback would ignore the Carousel component

mineschan:
I have re-commit with detekt check on and fixes the errors.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
marianeum: Ok let me check what other platforms are doing re accessibility for this

build.gradle
```diff
@@ -135,6 +135,7 @@ ext {
   composeActivity = "androidx.activity:activity-compose:1.6.1"
 
   accompanistFlowLayout = "com.google.accompanist:accompanist-flowlayout:0.28.0"
+  accompanistPager = "com.google.accompanist:accompanist-pager:0.28.0"
```
marianeum, line 138: can you add an `accompanistVersion` property and use it for both accompanist libraries to ensure they stay in sync?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
marianeum: Just checked and the reason it is ignored is because we're not providing a content description for the items in the sample (the `image` content description is null) - can you set a content description please? this brings up a different issue where after each page change it switches out of the gallery. i wonder if that has to do with the infinite scroll logic

build.gradle
mineschan: added👍🏻

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
mineschan: I've tested removing different logic and indeed the count with MAX_VALUE causes this weird behaviour. 🙁
However, this is the suggested way to add infinite scroll the the Pager, I'm not sure if we could actually fix that.
https://github.com/google/accompanist/blob/main/sample/src/main/java/com/google/accompanist/sample/pager/HorizontalPagerLoopingSample.kt

marianeum: I'm gonna page @bvitaliyg here who knows a bit more about this than I.
I think in general we might need to change this a little more:
We have the `currentPage` in the API as per spec, but really it is treated like a `initialPage` property as the state is handled within the component. However I'm not sure how in this component this could be moved to the usage with the infinite scroll behaviour

mineschan: btw any idea why the label-check is stuck in the pipeline?

marianeum: good question! i noticed the same happened with another PR the other week. it looks like it didn't trigger. i just added/removed another label and that did it

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
```diff
@@ -0,0 +1,41 @@
...
+package net.skyscanner.backpack.compose.carousel
+
+import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.carousel.internal.BpkCarouselImpl
+
+@Composable
+fun BpkCarousel(
+  imageContent: @Composable (BoxScope.(Int) -> Unit),
```
bvitaliyg, line 28: It's usually named content and it's the last param in arguments list

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
```diff
@@ -0,0 +1,41 @@
...
+  count: Int,
```
bvitaliyg, line 29: Do we have these params name from Figma?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
```diff
@@ -0,0 +1,41 @@
...
+  modifier: Modifier = Modifier,
+  currentImage: Int = 0,
+  onImageChanged: ((Int) -> Unit)? = null,
```
bvitaliyg, line 32: I assume we always want to react on changing the image as it's the only way to update the content. If so, let's make it non-nullable

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+  count: Int,
+  currentImage: Int,
+  modifier: Modifier = Modifier,
+  onImageChanged: ((Int) -> Unit)? = null,
+) {
+  Box(modifier = modifier) {
+    val initialPageIndex = (Int.MAX_VALUE / 2) + currentImage
+    val (selectedPage, setSelectedPage) = rememberSaveable { mutableStateOf(currentImage) }
+
+    val pagerState = rememberPagerState(initialPageIndex).also {
```
bvitaliyg, line 51: Why pager state is internal? The devs might want to access it from the business logic, for instance to perform scrolls to position

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+internal fun BpkCarouselImpl(
+  imageContent: @Composable (BoxScope.(Int) -> Unit),
+  count: Int,
+  currentImage: Int,
+  modifier: Modifier = Modifier,
+  onImageChanged: ((Int) -> Unit)? = null,
+) {
+  Box(modifier = modifier) {
+    val initialPageIndex = (Int.MAX_VALUE / 2) + currentImage
+    val (selectedPage, setSelectedPage) = rememberSaveable { mutableStateOf(currentImage) }
```
bvitaliyg, line 49: Why do we remember this internally? It seems these props need to be injected and be a part of the component API

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    val pagerState = rememberPagerState(initialPageIndex).also {
+      LaunchedEffect(it) {
+        snapshotFlow { it.currentPage }.distinctUntilChanged().collect { index ->
```
bvitaliyg, line 53: LaunchedEffect(pagerState.currentPage) {
val page = getModNumber ...

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
bvitaliyg: Also, we just need to use by keyword with remembers here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+          .testTag("pageIndicator"),
+        totalIndicators = count,
+        currentIndex = selectedPage,
+        style = BpkPageIndicatorStyle.OverImage,
+      )
+    }
+  }
+}
+
+private fun getModNumber(index: Int, count: Int) = (index - (Int.MAX_VALUE / 2)).floorMod(count)
```
bvitaliyg, line 84: So what's the purpose of this function?

mineschan: It's from this sample from accompanist. It calculates with modulo operation methods to get the real index of the current image, after we add the passed index on top of a large number (Int.MAX_VALUE / 2) for infinite looping use.
https://github.com/google/accompanist/blob/main/sample/src/main/java/com/google/accompanist/sample/pager/HorizontalPagerLoopingSample.kt

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
mineschan: it should be the callback AFTER the pager is scrolled to a different page, not the callback to get the content from?
https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=8335%3A38592&t=kaBaEYMePlbY8sWh-0

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
mineschan: the reason we need count here is we don't pass in an array of "images" into the composable, but to supply the composable with the imageContent callback from the pager content callback, to do so we need to let the Carousel know how many items are there.

And the reason to not supply images is we need external library (e.g. GlideImage) to load the image and this should not be part of the Backpack implementation.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
mineschan: I can add a function to calculate the initialPage for the pager which takes the "currentImage" as it cannot be assigned later if it is hoisted at an upper level. So instead of passing the currentImage to BpkCarousel, an optional PagerState will be needed. wdyt?

```
@ExperimentalPagerApi
@Composable
fun rememberCarouselPagerState(
  @IntRange(from = 0) currentImage: Int = 0,
): PagerState {
  val initialPageIndex = (Int.MAX_VALUE / 2) + currentImage
  return rememberPagerState(initialPageIndex)
}

```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
mineschan: I remove this variable completely and use the currentPage from the pagerstate

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
marianeum: the name is specified in the api, so should stay consistent

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
```diff
@@ -0,0 +1,86 @@
...
+): BpkCarouselState {
+  val initialPage = (Int.MAX_VALUE / 2) + currentImage
+  val pagerState = rememberPagerState(initialPage = initialPage)
+  return remember(pagerState) {
+    BpkCarouselStateImpl(delegate = pagerState, totalImages = totalImages)
+  }
+}
+
+@OptIn(ExperimentalPagerApi::class)
+fun BpkCarouselState(
```
marianeum, line 69: when should this be used vs the `remember` function?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
```diff
@@ -0,0 +1,86 @@
...
+  totalImages: Int,
+  currentImage: Int = 0,
+): BpkCarouselState {
+  val initialPage = (Int.MAX_VALUE / 2) + currentImage
+  return BpkCarouselStateImpl(delegate = PagerState(currentPage = initialPage), totalImages = totalImages)
+}
+
+@OptIn(ExperimentalPagerApi::class)
+internal interface BpkCarouselInternalState : BpkCarouselState {
```
marianeum, line 78: rather than the casting & separate interface could there just be internal props on `BpkCarouselState`?

marianeum (commented):
again i think @bvitaliyg has more context on this but there are the two questions around this

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
```diff
@@ -0,0 +1,86 @@
...
+
+@OptIn(ExperimentalPagerApi::class)
+@Composable
+fun rememberBpkCarouselState(
+  totalImages: Int,
+  currentImage: Int = 0,
+): BpkCarouselState {
+  val initialPage = (Int.MAX_VALUE / 2) + currentImage
+  val pagerState = rememberPagerState(initialPage = initialPage)
+  return remember(pagerState) {
```
bvitaliyg, line 63: remember(pagerState, totalImages)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
bvitaliyg: This can be used if you want to store your state in the business logic.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
bvitaliyg: No internal props are allowed for the interfaces

bvitaliyg: To BpkCarouselStateImpl?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+@OptIn(ExperimentalPagerApi::class)
+@Composable
+internal fun BpkCarouselImpl(
+  state: BpkCarouselInternalState,
+  modifier: Modifier = Modifier,
+  onImageChanged: ((Int) -> Unit)? = null,
+  imageContent: @Composable (BoxScope.(Int) -> Unit),
+) {
+  Box(modifier = modifier) {
+    LaunchedEffect(state.currentPage) {
```
bvitaliyg, line 46: LaunchedEffect(state.currentPage) {
onImageChanged?.invoke(state.currentPage)
} 
should be just fine

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/internal/BpkCarouselImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+          .testTag("pageIndicator"),
+        totalIndicators = state.pageCount,
+        currentIndex = state.currentPage,
+        style = BpkPageIndicatorStyle.OverImage,
+      )
+    }
+  }
+}
+
+internal fun getModdedPageNumber(index: Int, count: Int) = (index - (Int.MAX_VALUE / 2)).floorMod(count)
```
bvitaliyg, line 76: Can we make this functions as part of state impl?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
mineschan: This approach was referenced from TopNavBarState and TopNavBarInternalState in BpkTopNavBar.

app/src/main/java/net/skyscanner/backpack/demo/compose/CarouselStory.kt
```diff
@@ -0,0 +1,112 @@
...
+    CarouselSample(totalImages = 3, currentImage = currentIndex, showCurrentPageLabel = true)
+  }
+}
+
+@OptIn(ExperimentalPagerApi::class)
+@Composable
+private fun CarouselSample(
+  totalImages: Int,
+  modifier: Modifier = Modifier,
+  currentImage: Int = 0,
```
bvitaliyg, line 80: intialImage

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
bvitaliyg: Please name it to content, this is what Compose guidelines suggest

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
```diff
@@ -0,0 +1,90 @@
...
+        totalIndicators = internalState.pageCount,
+        currentIndex = internalState.currentPage,
+        style = BpkPageIndicatorStyle.OverImage,
+      )
+    }
+  }
+}
+
+@Composable
+fun BpkCarousel(
```
bvitaliyg, line 69: We can remove this function entierly

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
```diff
@@ -0,0 +1,116 @@
...
+    @IntRange(from = 0) page: Int,
+    @FloatRange(from = -1.0, to = 1.0) pageOffset: Float = 0f,
+  )
+}
+
+@OptIn(ExperimentalPagerApi::class)
+@Composable
+fun rememberBpkCarouselState(
+  totalImages: Int,
+  currentImage: Int = 0,
```
bvitaliyg, line 58: initialImage. Also, please make it the 1st argument

docs/compose/Carousel/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Example of starting in a different image:
+
+```Kotlin
+import net.skyscanner.backpack.compose.carousel.BpkCarousel
+import net.skyscanner.backpack.compose.carousel.rememberBpkCarouselState
+
+// show carousel with image index 1
+val state = rememberBpkCarouselState(totalImages = totalImages, currentImage = currentImage)
```
bvitaliyg, line 58: Please don't forget to update the namings here as well

bvitaliyg (commented):
Some small comments

app/src/main/java/net/skyscanner/backpack/demo/compose/CarouselStory.kt
```diff
@@ -0,0 +1,112 @@
...
+  totalImages: Int,
+  modifier: Modifier = Modifier,
+  initialImage: Int = 0,
+  showCurrentPageLabel: Boolean = false,
+) {
+  val pagerState = rememberBpkCarouselState(totalImages = totalImages, initialImage = initialImage)
+
+  if (showCurrentPageLabel) {
+    BpkText(
+      text = stringResource(id = R.string.carousel_current_image).format(pagerState.currentPage),
```
bvitaliyg, line 87: You can just use stringResource(R.string.carousel_current_image, pagerState.currentPage)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarousel.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.google.accompanist.pager.HorizontalPager
+import net.skyscanner.backpack.compose.pageindicator.BpkPageIndicator
+import net.skyscanner.backpack.compose.pageindicator.BpkPageIndicatorStyle
+
+@OptIn(ExperimentalPagerApi::class)
+@Composable
+fun BpkCarousel(
+  state: BpkCarouselState,
+  modifier: Modifier = Modifier,
+  imageContent: @Composable (BoxScope.(Int) -> Unit),
```
bvitaliyg, line 38: Please rename it to content

docs/compose/Carousel/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+BpkCarousel(state = state) { index ->
```
bvitaliyg, line 60: NIT: BpkCarousel(state)

docs/compose/Carousel/README.md
bvitaliyg: NIT: rememberBpkCarouselState(totalImages, currentImage) (or 1 as you comment says)

bvitaliyg (commented):
One request and couple of NITs

bvitaliyg (approved):
💯

PR #1478, opened 2023-02-24
Koa 5892 Implement support of custom KSP processor for screenshots

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/app/src/androidTest/java/net/skyscanner/backpack/docs/GenerateScreenshots1.kt
+++ b/app/src/androidTest/java/net/skyscanner/backpack/docs/GenerateScreenshots1.kt
@@ -20,26 +20,19 @@ package net.skyscanner.backpack.docs
 
 import android.content.Intent
 import androidx.appcompat.app.AppCompatDelegate
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.test.junit4.AndroidComposeTestRule
 import androidx.test.rule.ActivityTestRule
-import net.skyscanner.backpack.compose.floatingnotification.BpkFloatingNotification
-import net.skyscanner.backpack.compose.floatingnotification.rememberBpkFloatingNotificationState
-import net.skyscanner.backpack.compose.navigationbar.BpkTopNavBar
-import net.skyscanner.backpack.compose.navigationbar.NavIcon
-import net.skyscanner.backpack.demo.BackpackDemoTheme
 import net.skyscanner.backpack.demo.BpkBaseActivity
-import net.skyscanner.backpack.demo.compose.LocalAutomationMode
-import net.skyscanner.backpack.demo.compose.LocalFloatingNotification
+import net.skyscanner.backpack.demo.meta.StoriesRepository
 import net.skyscanner.backpack.demo.meta.Story
-import net.skyscanner.backpack.demo.meta.all
+import net.skyscanner.backpack.demo.ui.BpkScaffold
+import net.skyscanner.backpack.demo.ui.StoryScreen
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
 
+// todo: rename it when we all the stories migrated
 @RunWith(Parameterized::class)
 open class GenerateScreenshots1(
   private val story: Story,
@@ -49,7 +42,7 @@ open class GenerateScreenshots1(
 
     @JvmStatic
     @Parameterized.Parameters(name = "{0} Screenshot")
-    fun data(): List<Story> = Story.all().filter { it.isScreenshot }
+    fun data(): List<Story> = StoriesRepository.getInstance().screenshotStories()
 
   }
 
@@ -75,23 +68,8 @@ open class GenerateScreenshots1(
     val intent = Intent()
     activityRule.launchActivity(intent)
     composeTestRule.setContent {
-      BackpackDemoTheme {
-        val floatingNotificationState = rememberBpkFloatingNotificationState()
-        CompositionLocalProvider(
-          LocalAutomationMode provides true,
-          LocalFloatingNotification provides floatingNotificationState,
-        ) {
-          Box {
-            Column {
-              BpkTopNavBar(
-                navIcon = NavIcon.Back("back", {}),
-                title = story.component.name,
-              )
-              story.content()
-            }
-            BpkFloatingNotification(state = floatingNotificationState)
-          }
-        }
+      BpkScaffold(automationMode = true) {
+        StoryScreen(story = story, onBack = {})
       }
     }
     takeScreenshot(suffix)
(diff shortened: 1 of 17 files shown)


app/src/main/java/net/skyscanner/backpack/demo/meta/ViewStory.kt
```diff
@@ -30,5 +29,4 @@ import javax.annotation.concurrent.Immutable
 annotation class ViewStory(
   val name: String = "Default",
   val screenshot: Boolean = true,
-  @LayoutRes val layoutId: Int = 0,
```
bvitaliyg, line 33: If you try to use it with Android Studio, you'll get the following warning: "Resource IDs will be non-final by default in Android Gradle Plugin version 8.0, avoid using them as annotation attributes". Therefore I'm removing this attribute (and this also simplifies the Visitors)

app/src/main/java/net/skyscanner/backpack/demo/meta/StoriesRepository.kt
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+package net.skyscanner.backpack.demo.meta
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import org.jetbrains.annotations.TestOnly
+
+interface StoriesRepository {
```
bvitaliyg, line 26: This becomes necessary if we want to have the test stories.

app/src/androidTest/java/net/skyscanner/backpack/meta/StoriesTests.kt
```diff
@@ -18,28 +18,32 @@
...
 
-import net.skyscanner.backpack.demo.meta.Story
-import net.skyscanner.backpack.demo.meta.all
+import net.skyscanner.backpack.demo.meta.StoriesRepository
 import org.junit.Assert.assertTrue
-import org.junit.Ignore
 import org.junit.Test
 
-@Ignore("These tests are ignored because we need to migrate the stories first")
 class StoriesTests {
```
marianeum, line 25: now that we've got the right stories can you make these tests a bit more useful - ie test the expected properties are set etc, rather than just "they exist"

app/src/main/java/net/skyscanner/backpack/demo/ui/BpkScarldfold.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.floatingnotification.BpkFloatingNotification
+import net.skyscanner.backpack.compose.floatingnotification.rememberBpkFloatingNotificationState
+import net.skyscanner.backpack.demo.BackpackDemoTheme
+import net.skyscanner.backpack.demo.compose.LocalAutomationMode
+import net.skyscanner.backpack.demo.compose.LocalFloatingNotification
+
+@Composable
+fun BpkScaffold(
```
marianeum, line 32: the file name has a typo. can we rename this to `DemoScaffold` or similar to avoid confusion with a potential `BpkScaffold` component?

app/src/main/java/net/skyscanner/backpack/demo/meta/ViewStory.kt
marianeum: how are view stories defined then? the sample specifies it's a view story, but it's actually a compose story.
also with this gone, maybe we just have a `Story` annotation which has a type for view vs compose?

app/src/main/java/net/skyscanner/backpack/demo/meta/StoriesRepository.kt
```diff
@@ -0,0 +1,103 @@
...
+      ?.sortedBy { it.name }
+      ?: emptyList()
+
+  override fun isComposeSupportedFor(component: Component): Boolean =
+    map[component]?.any { it.isCompose } ?: false
+
+  override fun testStories() = testStories
+}
+
+@Component(name = "TestComponent", link = "test")
```
marianeum, line 74: what is the `link` used  for again?

meta-processor/src/main/kotlin/net/skyscanner/backpack/ksp/writer/StoriesWriter.kt
```diff
@@ -64,23 +64,17 @@ fun writeListOfStories(stories: List<StoryDefinition>, output: XFiler) {
...
-    .addStatement("isCompose = %L,", story.isCompose)
-    .writeComponent("component", story.component)
-    .writeContentBlock("content", story)
-    .unindent()
-    .addStatement("),")
-
-private val AndroidLayout = ClassName("net.skyscanner.backpack.demo.ui", "AndroidLayout")
-
-private fun CodeBlock.Builder.writeContentBlock(name: String, story: StoryDefinition) =
   when {
```
marianeum, line 67: what's the when statement for?

app/src/androidTest/java/net/skyscanner/backpack/docs/GenerateScreenshots1.kt
```diff
@@ -0,0 +1,92 @@
...
+  companion object {
+
+    @JvmStatic
+    @Parameterized.Parameters(name = "{0} Screenshot")
+    fun data(): List<Story> = StoriesRepository.getInstance().screenshotStories()
+
+  }
+
+  @get:Rule
+  var activityRule = ActivityTestRule(BpkBaseActivity::class.java, true, false)
```
marianeum, line 50: did you try and see if we can switch to `ActivityScenario` for this now?

app/src/main/java/net/skyscanner/backpack/demo/meta/ViewStory.kt
bvitaliyg: Having separate annotations seems to be an easier way to use rather than specifying the type of each type.

And all the stories are defined as Composables, it's just you use AndoidView interop for view.

app/src/androidTest/java/net/skyscanner/backpack/meta/StoriesTests.kt
bvitaliyg: But we do test the properties – e.g. whether it's a compose vs view story, as well as component definitions

app/src/main/java/net/skyscanner/backpack/demo/meta/StoriesRepository.kt
bvitaliyg: For instance, for lint rules in the future

app/src/androidTest/java/net/skyscanner/backpack/docs/GenerateScreenshots1.kt
bvitaliyg: Nope, not as a part of this ticket for now

app/src/androidTest/java/net/skyscanner/backpack/meta/StoriesTests.kt
marianeum: these still just test for "is there any story that has x" rather than "is this story the expected result"?

marianeum (approved):

PR #1485, opened 2023-02-28
Adding FlightLeg compose component

athilahs:
Adding the new FlightLeg component. Specifications can be found [in Figma](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=2336%3A17803&t=7wtdDERsRnMkXu6s-0)

The screenshot below showcases the component being applied in the Skyscanner app:

Screenshot 2023-02-27 at 17 58 59

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,266 @@
...
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.tokens.Aircraft
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+fun BpkFlightLeg(
+  departureArrivalTime: String,
+  description: AnnotatedString,
+  stopsInfo: AnnotatedString,
```
athilahs, line 53: I decided to make `stopsInfo` an `AnnotatedString` as opposed to a `String` as defined in the API in Figma to make it more consistent with the other externally-stylable text in the component (description) and give more freedom for clients to style it as they need.
This also means it is easier to integrate the component when formatting texts are returned from a BFF.
For example, in Save To List feature, the BFF would return this field with styling tags (e.g.: `stopsInfo: <style0>2 stops</style0>`). 

We already have code to read these tags and return an `AnnotatedString` out of it for various style tags we support. By making this field an AnnotatedString, we can reuse that logic and reduce the use of bespoke code (this case would have to be treated as an exception to the style-tags mechanism)

Please let me know if there are any concerns

marianeum: The reason for the API proposal was to avoid exposing styling to consumers. We need to do this for the other options given the current backend responses, but ultimately this should be more an exception than a norm - this is also why there's two tables of API design with the second being the preferred option, but due to the current limitation in APIs we're not able to do this.
If we change approach this should be done across all platforms, rather than a change just for Android, so please raise it in the flightleg channel with your concerns.

app/src/androidTest/java/net/skyscanner/backpack/compose/flightleg/BpkFlightLegTest.kt
```diff
@@ -0,0 +1,93 @@
...
+import net.skyscanner.backpack.compose.icon.BpkIcon
+import net.skyscanner.backpack.compose.theme.BpkTheme
+import net.skyscanner.backpack.compose.tokens.Aircraft
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+import org.junit.Test
+
+class BpkFlightLegTest : BpkSnapshotTest() {
+
+  @Test
+  fun simpleLegTest() = snap {
```
marianeum, line 39: could you just reuse the samples from the stories for these tests, rather than redefining essentially a copy here"

app/src/androidTest/java/net/skyscanner/backpack/compose/flightleg/BpkFlightLegTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    BpkFlightLeg(
+      modifier = Modifier.background(BpkTheme.colors.canvas),
```
marianeum, line 41: this should already be the default background

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,266 @@
...
+  duration: String,
+  modifier: Modifier = Modifier,
+  nextDayArrival: String? = null,
+  operatedBy: String? = null,
+  warning: String? = null,
+  contentDescription: String? = null,
+  carrierLogo: @Composable (BoxScope.() -> Unit)? = null,
```
marianeum, line 60: can you add the `content` suffix to follow compose API standards

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/internal/BpkAccessoryTitleDetails.kt
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Composable
+fun BpkAccessoryTitleDetails(
```
marianeum, line 37: I'm not sure about this abstraction - IMO it makes it harder to understand how the component is structured. title, description & accessory view don't really make it clear how they're arranged and I find it easier to look at the whole component and define rows/columns within, rather than abstracting this away. 
Also this is a public function, meaning it'd be exposed as a component, but it is not a defined component. 
I know this is an abstraction you're using in sonic, but it's not really used within backpack or wider.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,266 @@
...
+      details = listOf(
+        AnnotatedString(duration),
+      ),
+    )
+  }
+}
+
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO, showBackground = true, backgroundColor = 0xffffff)
+@Composable
+private fun LightModeSimpleDirect() {
```
marianeum, line 126: we generally don't provide previews within the component. also there's a new `BpkPreviews` annotation for use to provide the default, dm + rtl combination - let me share that in the compose channel :) )

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
athilahs: thanks! raised the discussion there

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/internal/BpkAccessoryTitleDetails.kt
athilahs: yeah, I wasn't too sure about this either. I'll remove it and move everything to the FlightLeg component

app/src/main/java/net/skyscanner/backpack/demo/compose/FlightLegStory.kt
```diff
@@ -0,0 +1,108 @@
...
+@Preview(showBackground = true)
+fun BasicFlightLegSample() {
+  BpkFlightLeg(
+    departureArrivalTime = "19:50 - 22:45",
+    description = AnnotatedString("LHR-SIN,SwissAir"),
+    stopsInfo = "Direct",
+    duration = "7h 55m",
+    carrierLogoContent = {
+      Box(
+        modifier = Modifier.padding(top = BpkSpacing.Sm),
```
marianeum, line 70: should this padding be applied within the component? ie will this be needed for every icon?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,143 @@
...
+                  baselineShift = BaselineShift.Superscript,
+                ),
+              ) {
+                append(it)
+              }
+            }
+          },
+          color = BpkTheme.colors.textPrimary,
+        )
+        mutableListOf<AnnotatedString>().apply {
```
marianeum, line 97: compose generally isn't very compatible with mutable types. rather than creating spans here could you just have separate `BpkText` composables? that way you can also apply the full text style, rather than just the font size (that might be something we need to figure out how we can solve actually for spans, but that seems outside of the scope of this PR)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/internal/BpkAccessoryTitleDetails.kt
marianeum: can you remove this please?

athilahs: 🤦.
Thanks! forgot to remove it

app/src/main/java/net/skyscanner/backpack/demo/compose/FlightLegStory.kt
athilahs: good call. better leave this for the clients to provide

marianeum: this is in the sample, so i'm asking if it should be done in the component (i believe it does looking at figma, but i might be wrong). as it's currently done in the sample this may lead to inconsistencies of icon positioning

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
athilahs: sure.
what do you mean by 
> apply the full text style, rather than just the font size
?
I'd have multiple BpkText with style `caption` and color `textSecondary`. Is that what you'd expect?

marianeum: yep that's what i mean. this is different to the span where we just apply the text size of the style due to limitations (which we may need to figure out how to solve within backpack, but that's on us)

app/src/main/java/net/skyscanner/backpack/demo/data/ComponentRegistry.kt
```diff
@@ -361,6 +362,7 @@ object ComponentRegistry {
         TAB_TITLE_COMPOSE composeStory { FlareStory() },
       ),
     ),
+    "FlightLeg" composeStory { FlightLegStory() },
```
marianeum, line 365: massive nitpick: can you add a space here? (also need to update the `DocsRegistry` then)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,138 @@
...
+  departureArrivalTime: String,
+  description: AnnotatedString,
+  stopsInfo: String,
+  duration: String,
+  modifier: Modifier = Modifier,
+  nextDayArrival: String? = null,
+  highlightStopsInfo: Boolean = false,
+  operatedBy: String? = null,
+  warning: String? = null,
+  contentDescription: String? = null,
```
marianeum, line 57: I wonder if it's better to not make this have a default value to make sure people are explicitly setting either a content description or null, rather than missing it and nothing will be announced

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,138 @@
...
+          BpkText(
+            text = it,
+            style = BpkTheme.typography.caption,
+            color = BpkTheme.colors.textSecondary,
+          )
+        }
+        warning?.let {
+          BpkText(
+            text = it,
+            style = BpkTheme.typography.caption,
```
marianeum, line 116: this should be `label3` (as bold)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,138 @@
...
+          content = it,
+        )
+        Spacer(modifier = Modifier.width(BpkSpacing.Base))
+      }
+      Column(
+        horizontalAlignment = Alignment.Start,
+      ) {
+        BpkText(
+          text = buildAnnotatedString {
+            append(departureArrivalTime)
```
marianeum, line 86: `departureArrivalTime` should be `heading5`

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,138 @@
...
+            nextDayArrival?.let {
+              withStyle(
+                SpanStyle(
+                  color = BpkTheme.colors.textPrimary,
+                  fontSize = BpkTheme.typography.label3.fontSize,
```
marianeum, line 91: this should be `caption`

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
```diff
@@ -0,0 +1,138 @@
...
+        )
+        Spacer(modifier = Modifier.width(BpkSpacing.Base))
+      }
+      Column(
+        horizontalAlignment = Alignment.Start,
+      ) {
+        BpkText(
+          text = buildAnnotatedString {
+            append(departureArrivalTime)
+            nextDayArrival?.let {
```
marianeum, line 87: looking at figma the text is a lot higher here than there. it also looks like the icon + direct text alignment is off when the +1 text is visible

marianeum (commented):
noticed a few inonsistencies with figma - can you double check that everything looks as expected please?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg/BpkFlightLeg.kt
athilahs: thanks! changed it to `label3`
where did you see it specifying as bold? Visually it does look bold but I couldn't find anywhere in figma specifying it to be bold. In fact, the `CSS` section of the page shows `font-style: normal;`.
For educational purposes, how are font weights specified in these designs in Figma?

athilahs:
@marianeum thanks for your comments. I did a spot-check and I could not find any other inconsistencies, apart from the `+1` text inheriting some properties from the `heading5` style (if you look closely, it doesn't look as thin as in Figma. It looks a bit bold). But I believe this is due to what you mentioned before: we cannot specify `fontStyle` on `SpanStyle`'s, only `fontSize`.
Please let me know if you have any suggestions on how to fix this (or if you spot any other inconsistencies) or if we're happy to proceed with these changes

marianeum:
Maybe instead of using a span we could have a separate `BpkText` and align it to the top of the other one? i'm not sure if that would align it correctly, but worth a try.

app/src/main/java/net/skyscanner/backpack/demo/compose/FlightLegStory.kt
```diff
@@ -0,0 +1,101 @@
...
+    CompleteFlightLegSample()
+  }
+}
+
+@Composable
+@BpkPreviews
+fun BasicFlightLegSample() {
+  BpkFlightLeg(
+    departureArrivalTime = "19:50 - 22:45",
+    description = AnnotatedString("LHR-SIN,SwissAir"),
```
marianeum, line 64: looks like a space is missing here

docs/compose/FlightLeg/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+[![Maven Central](https://img.shields.io/maven-central/v/net.skyscanner.backpack/backpack-compose)](https://search.maven.org/artifact/net.skyscanner.backpack/backpack-compose)
+[![Class reference](https://img.shields.io/badge/Class%20reference-Android-blue)](https://backpack.github.io/android/backpack-compose/net.skyscanner.backpack.compose.flightleg)
+[![Source code](https://img.shields.io/badge/Source%20code-GitHub-lightgrey)](https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/flightleg)
+
+## Default
+
+| Day | Night |
+| --- | --- |
+| <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/FlightLeg/screenshots/default.png" alt="FlightLeg component" width="375" /> | <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/FlightLeg/screenshots/default_dm.png" alt="FlightLeg component - dark mode" width="375" /> |
```
marianeum, line 11: I think the screenshots need updating following the latest changes

app/src/main/java/net/skyscanner/backpack/demo/compose/FlightLegStory.kt
```diff
@@ -0,0 +1,101 @@
...
+}
+
+@Composable
+@BpkPreviews
+fun CompleteFlightLegSample() {
+  BpkFlightLeg(
+    departureArrivalTime = "19:50 - 22:45",
+    description = buildAnnotatedString {
+      withStyle(
+        SpanStyle(
```
marianeum, line 81: I think we should have this style available from backpack, so it stays consistent in case we change it. maybe `BpkAirportHighlightStyle` 
Also make it clear within the readme that only this should be used vs other styling

marianeum (commented):
two more small comments. alignment looks much better now, thanks :)

marianeum (approved):
thanks - looks great!

PR #1490, opened 2023-03-02
KOA-5923 Ensure detekt runs as expected on all files

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -184,7 +184,7 @@ formatting:
     active: true
     autoCorrect: true
     indentSize: 2
-    excludes: [ '**/tokens/**' ]
+    excludes: [ '**/compose/tokens/**' ]
   KdocWrapping:
     active: false
   MaximumLineLength:
@@ -199,11 +199,11 @@ formatting:
   NoBlankLineBeforeRbrace:
     active: true
     autoCorrect: true
-    excludes: [ '**/tokens/**', '**/check/BpkDeprecatedTokens.kt' ]
+    excludes: [ '**/compose/tokens/**', '**/check/BpkDeprecatedTokens.kt' ]
   NoConsecutiveBlankLines:
     active: true
     autoCorrect: true
-    excludes: [ '**/tokens/**' ]
+    excludes: [ '**/compose/tokens/**' ]
   NoEmptyClassBody:
     active: true
     autoCorrect: true
@@ -265,7 +265,7 @@ formatting:
   SpacingAroundParens:
     active: true
     autoCorrect: true
-    excludes: [ '**/tokens/**' ]
+    excludes: [ '**/compose/tokens/**' ]
   SpacingAroundRangeOperator:
     active: true
     autoCorrect: true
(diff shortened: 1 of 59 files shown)


.detekt.yml
```diff
@@ -184,7 +184,7 @@ formatting:
     active: true
     autoCorrect: true
     indentSize: 2
-    excludes: [ '**/tokens/**' ]
+    excludes: [ '**/compose/tokens/**' ]
```
bvitaliyg, line 187: To make sure it only skips the generated files

build.gradle
```diff
@@ -61,6 +61,7 @@ subprojects {
   detekt {
     config = files("$rootDir/.detekt.yml", "$rootDir/.detekt-compose.yml")
     buildUponDefaultConfig = true
+    source = files("src", "$rootDir/buildSrc/src")
```
bvitaliyg, line 64: By default, "detekt" task only runs for src/main/java, src/test/java, src/main/kotlin and src/test/kotlin. See the docs here: https://detekt.dev/docs/gettingstarted/gradle

By using "src", we always run detekt regardless tests/main, flavors, etc.

There's "detektTest" task, but it doesn't support auto-correct. Also, it increases the build time to run 2 tasks instead of one.

Also, buildSrc is a separate project, and all we set up in the root build.gradle is not applicable there.
Therefore, in order to run detekt there, you need to set up the same stuff in buildSrc/build.gradle, and run it like this: "./gradlew -p buildSrc :detekt".

Or, since detekt is a static analyser, it's enough just to include "buildSrc/src" :)

The only potential downside is that the check will be run for each module, but it's quite small and doesn't increase the check time to any noticeable way.

app/src/androidTest/java/net/skyscanner/backpack/docs/DocsRegistry.kt
```diff
@@ -89,16 +88,12 @@ object DocsRegistry {
...
-    ComposeScreenshot("Dialog - Compose", "warning") { setupComposeDialog(it, ShownDialog.Warning) },
-    ComposeScreenshot("Dialog - Compose", "destructive") { setupComposeDialog(it, ShownDialog.Destructive) },
-    ComposeScreenshot("Dialog - Compose", "flare") { setupComposeDialog(it, ShownDialog.Flare) },
-    ComposeScreenshot("Dialog - Compose", "image-start-alignment") {
-      setupComposeDialog(it, ShownDialog.ImageStartAlignment)
-    },
-    ComposeScreenshot("Dialog - Compose", "image-end-alignment") {
-      setupComposeDialog(it, ShownDialog.ImageEndAlignment)
-    },
+    ComposeScreenshot("Dialog - Compose", "success"),
```
marianeum, line 91: something snuck in here?

bvitaliyg: Yes, a part of the screenshots experiment! Good catch :D

PR #1502, opened 2023-03-10
[KOA-5959] Automate Supernova docs publish

anambl:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,3 +93,9 @@ jobs:
           file: app/build/outputs/apk/oss/release/app-${{ env.flavour }}-release.apk
           silent: true
           releaseNotes: Version ${{github.event.release.tag_name}}
+
+  SupernovaPublish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update Supernova docs
+        run: supernova publish-documentation --apiKey=${{ secrets.SUPERNOVA_API_KEY }} --designSystemid=${{ secrets.SUPERNOVA_DESIGN_SYSTEM_ID }}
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -220,6 +220,8 @@ Please see the [code review guidelines](https://github.com/Skyscanner/backpack/b
  - Ensure CI runs the release workflow successfully
  - Once released verify the artifacts are available
 
+ The release workflow will also trigger our design docs publish. If successful, you should see your component changes at [skyscanner.design](https://skyscanner.design).
```
marianeum, line 223: maybe add something like: 
> Note! Don't forget that new components need to be added manually!

marianeum (approved):

.github/workflows/release.yml
```diff
@@ -96,6 +96,7 @@ jobs:
 
   SupernovaPublish:
     runs-on: ubuntu-latest
+    environment: Publishing
```
marianeum, line 99: it might be worth moving it to the release workflow otherwise we need to approve two bits manually. Also you'll have to move the environmental variables to the publishing environment as well (not sure if you did that)

anambl: Yep, moved the secrets to the publishing env.
Not sure I understand what you mean - it is part of the release workflow, no? or did you mean part of the Deploy job?
The thing is on iOS the release workflow is structured into multiple jobs so actually the dokka docs for example is a separate job. So there I added it as a separate job and wanted to keep it consistent on all repos so added it separately here as well. The approval is just a one-off right?

marianeum: ah cool - i thought every environment workflow will need to be approved manually, but looks like it just needs to be done once for a run, so should be good!

marianeum (approved):

PR #1530, opened 2023-03-24
Koa-5408 Added a search bar in the demo app

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
+++ b/app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
@@ -91,7 +91,7 @@ fun ComponentListScreen(
         var searchQuery by rememberSaveable { mutableStateOf("") }
         BpkTextField(
             placeholder = stringResource(R.string.navigation_search),
-            value = searchQuery.replaceFirstChar { it.uppercase() },
+            value = searchQuery,
             onValueChange = { value ->
                 searchQuery = value
             },
@@ -101,11 +101,11 @@ fun ComponentListScreen(
             icon = BpkIcon.Search,
         )
         val filteredTokens = repository.tokenComponents.filter {
-            it.name.contains(searchQuery)
+            it.name.contains(searchQuery, ignoreCase = true)
         }
 
         val filteredComponents = repository.uiComponents.filter {
-            it.name.contains(searchQuery)
+            it.name.contains(searchQuery, ignoreCase = true)
         }
 
         LazyColumn {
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
```diff
@@ -78,21 +89,45 @@ fun ComponentListScreen(
                 ),
             ),
         )
+        var textState by rememberSaveable { mutableStateOf("") }
```
bvitaliyg, line 92: searchQuery, perhaps?

app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
```diff
@@ -78,21 +89,45 @@ fun ComponentListScreen(
...
+            value = textState.replaceFirstChar { it.uppercase() },
+            onValueChange = { value ->
+                textState = value
+            },
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(BpkSpacing.Base),
+            icon = BpkIcon.Search,
+        )
+        val filteredTokens = repository.tokenComponents.filter { it: Component ->
```
bvitaliyg, line 104: No need to specify it in these lambdas

bvitaliyg (commented):
Couple of small comments

app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
```diff
@@ -78,21 +89,45 @@ fun ComponentListScreen(
                 ),
             ),
         )
+        var textState by rememberSaveable { mutableStateOf("") }
+        BpkTextField(
+            placeholder = stringResource(R.string.navigation_search),
+            value = textState.replaceFirstChar { it.uppercase() },
```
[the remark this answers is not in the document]
jaanainkinen: @marianeum advised that the search text should be case insensitive, and this was my attempt to do that. Without it, it wouldn't recognise for example "badge", so with this it automatically turns the first character into uppercase. Is there a better way to do it?

app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
jaanainkinen: Thanks, renamed it!

app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
jaanainkinen: Removed them.

app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentListScreen.kt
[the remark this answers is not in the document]
jaanainkinen: This is fixed with ignoreCase.

PR #1531, opened 2023-03-24
KOA 5988 Migrate to gradle central dependency declaration

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/android-configuration.gradle
+++ b/android-configuration.gradle
@@ -48,8 +48,7 @@ android {
 }
 
 dependencies {
-    api rootProject.ext.androidxAnnotations
-    api "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    api rootProject.ext.coroutines
-    api rootProject.ext.androidxAppCompat
+    api libs.androidx.annotations
+    api libs.kotlin.coroutines
+    api libs.androidx.appCompat
 }
--- a/android-configuration.gradle
+++ b/android-configuration.gradle
@@ -49,6 +49,7 @@ android {
 
 dependencies {
     api libs.androidx.annotations
+    api libs.kotlin.stdlib
     api libs.kotlin.coroutines
     api libs.androidx.appCompat
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -97,10 +97,7 @@ android {
 }
 
 dependencies {
-    def composeBom = platform(libs.compose.bom)
     implementation fileTree(include: ['*.jar'], dir: 'libs')
-
-    implementation composeBom
     implementation libs.square.picasso
     implementation libs.accompanist.flowLayout
     implementation libs.androidx.lifecycleKtx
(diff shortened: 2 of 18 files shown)


app/src/main/java/net/skyscanner/backpack/demo/meta/Component.kt
```diff
@@ -18,7 +18,7 @@
 
 package net.skyscanner.backpack.demo.meta
 
-import javax.annotation.concurrent.Immutable
+import androidx.compose.runtime.Immutable
```
bvitaliyg, line 21: Imported wrong annotations previously

android-configuration.gradle
```diff
@@ -48,8 +48,7 @@ android {
 }
 
 dependencies {
-    api rootProject.ext.androidxAnnotations
-    api "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
marianeum, line 52: why was this removed?

app/build.gradle
```diff
@@ -97,29 +97,35 @@ android {
 }
 
 dependencies {
+    def composeBom = platform(libs.compose.bom)
```
marianeum, line 100: this would've been better as a separate PR as it makes it hard/impossible to compare

app/build.gradle
```diff
@@ -97,29 +97,35 @@ android {
...
+    implementation libs.square.picasso
+    implementation libs.accompanist.flowLayout
+    implementation libs.androidx.lifecycleKtx
+    implementation libs.compose.activity
+    implementation libs.kotlin.reflection
+    implementation libs.destinations.core
+    implementation libs.accompanist.pager
+    testImplementation libs.test.junit
+
+    debugImplementation composeBom
```
marianeum, line 113: do you actually have to re-declare it for everything? in the app we actually add the bom to all dependencies and declare it as an array instead, so you don't have to manually add it

backpack-lint/build.gradle
```diff
@@ -20,12 +20,11 @@ apply plugin: 'kotlin'
 apply from: "tokens.gradle.kts"
 
 dependencies {
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
marianeum, line 23: again why was this removed?

gradle/libs.versions.toml
marianeum, line 1: the dependabot config will need to be updated to make it work with this file

app/build.gradle
bvitaliyg: It seems so:
https://discuss.gradle.org/t/applying-a-bom-platform-to-all-configurations/30621

You were able to do this because it was from the code. Toml just tells Gradle how to generate dependency providers and it cannot do this. Here's the example of the same from Compose website:
https://developer.android.com/jetpack/compose/bom/bom#does_the_bom_work_with_version_catalogs

app/build.gradle
bvitaliyg: I can group the bomUsages in one block, so it's easier to review – WDYT?

gradle/libs.versions.toml
bvitaliyg: No need to – it works with existing configs automatically – https://github.com/dependabot/dependabot-core/issues/3121#issuecomment-1467901381

However, I removed the block for buildSrc as it's not longer needed

marianeum (approved):

PR #1544, opened 2023-04-12
KOA-6018 Migrate non broken components to MD3 – Compose

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/internal/BpkButtonImpl.kt
```diff
@@ -74,16 +74,19 @@ internal fun BpkButtonImpl(
...
                 disabledContentColor = if (loading) type.loadingContentColor() else type.disabledContentColor(),
             ),
             shape = ButtonShape,
             contentPadding = type.contentPadding,
             elevation = null,
             content = {
-                CompositionLocalProvider(LocalTextStyle provides size.textStyle()) {
+                CompositionLocalProvider(
+                    LocalTextStyle provides size.textStyle(),
+                    LocalContentColor provides androidx.compose.material3.LocalContentColor.current,
```
bvitaliyg, line 88: For some time, we're going to have 2 TextStyle/ContentColor composition locals in the project – from MD2 and MD3.

To simplify the migration, I replaced all MD2 CompositionLocals with Bpk CompositionLocals. This should have no effect and Bpk Composition Local actually refers to MD2 Composition locals.

Since our Text and Icon are used in a lot of places and actively use CompositionLocals in there, the default values for these components have been set to BpkCompositionLocal (aka MD2 composition locals). This way, the old code remains unbroken.

The new code, however, might be needed some tweaks. For instance, Button from MD3 sets MD3 CompositionLocals. Our BpkText/Icon wont be able to pick these up unless we set MD2 CompostionLocal = MD3 CompositionLocal like we're doing here.

These changes will only be temporary as we perform the migration, and it will only be affecting the MD3 migrated components (those that provide the containers with CompositionLocals). Once we finish the migration, we're going to point BPK CompositionLocals to MD3 and remove all the usages of non-BPK composition locals.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/fab/BpkFab.kt
```diff
@@ -44,14 +44,15 @@ fun BpkFab(
         interactionSource = interactionSource,
         modifier = modifier,
         shape = CircleShape,
-        backgroundColor = BpkTheme.colors.coreAccent,
+        containerColor = BpkTheme.colors.coreAccent,
```
bvitaliyg, line 47: It's been renamed to container for most of the cases

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/fab/BpkFab.kt
```diff
@@ -44,14 +44,15 @@ fun BpkFab(
...
         contentColor = BpkTheme.colors.textPrimaryInverse,
         elevation = FloatingActionButtonDefaults.elevation(BpkElevation.Xl, BpkElevation.Xl, BpkElevation.Xl, BpkElevation.Xl),
     ) {
         BpkIcon(
             icon = icon,
             contentDescription = contentDescription,
             size = BpkIconSize.Large,
+            tint = androidx.compose.material3.LocalContentColor.current,
```
bvitaliyg, line 55: Another case of containered widget which sets MD3 CompositionLocal – but BpkIcon takes MD2 local for now.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/internal/BpkButtonImpl.kt
```diff
@@ -74,16 +74,19 @@ internal fun BpkButtonImpl(
                 .requiredHeight(size.minHeight),
             interactionSource = interactionSource,
             colors = ButtonDefaults.buttonColors(
-                backgroundColor = type.backgroundColor(interactionSource),
+                containerColor = type.backgroundColor(interactionSource),
                 contentColor = type.contentColor(interactionSource),
-                disabledBackgroundColor = if (loading) type.loadingBackgroundColor() else type.disabledBackgroundColor(),
+                disabledContainerColor = if (loading) type.loadingBackgroundColor() else type.disabledBackgroundColor(),
```
bvitaliyg, line 79: Button will have some screenshots changed, and this is because in MD2 button the alpha wasn't applied.
This affects only the loading state in LinkOnDark as it uses linkOnDarkPressedForeground which has an alpha channel.

The reason for that is MD2 button was forcing alpha to be 1f and using ContentAlpha instead.
The new version is correct.

bvitaliyg (commented):
Some clarifications from me

marianeum (approved):

PR #1571, opened 2023-04-26
SONIC-1568: View Dialog Corrections

Dan7-7-7:

+ [ ] Component `README.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/dialog/internal/BpkDialogImpl.kt
```diff
@@ -19,103 +19,114 @@
...
+import android.view.WindowManager
 import android.widget.ImageView
 import android.widget.TextView
 import androidx.annotation.LayoutRes
 import net.skyscanner.backpack.R
 import net.skyscanner.backpack.button.BpkButton
 import net.skyscanner.backpack.dialog.BpkDialog
+import kotlin.math.min
+
+class BpkDialogImpl(
```
marianeum, line 42: is this mainly merging the two dialog classes? any breaking changes here from an api perspective? it doesn't look like given that it's internal, but wanted to double check

Dan7-7-7: I don't think so - @xo0ps will be able to confirm as he made this change.

xo0ps: @marianeum yes this PR merges two implementations of BpkDialog because they were using different xml layouts. By simplifiying the layouts I was able to remove specific implementations and merge them. Also nothing from api perspective is affected.

xo0ps (approved):
LGTM

marianeum (approved):

PR #1579, opened 2023-05-03
Fixed the BpkFloatingNotification content alignment

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/floatingnotification/internal/BpkFloatingNotification.kt
```diff
@@ -53,17 +55,20 @@ internal fun BpkFloatingNotificationImpl(
     modifier: Modifier = Modifier,
 ) {
 
-    Snackbar(
+    Surface(
```
bvitaliyg, line 58: MD3 implementation uses a custom layout inside and has weird content alignment issues.
Since we don't use the snackbar for anything else then colors and spacings after some investigation I decided just to use Surface instead.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/floatingnotification/internal/BpkFloatingNotification.kt
```diff
@@ -53,17 +55,20 @@ internal fun BpkFloatingNotificationImpl(
...
         contentColor = BpkTheme.colors.textOnDark,
     ) {
         CompositionLocalProvider(
             LocalContentColor provides androidx.compose.material3.LocalContentColor.current,
         ) {
             Row(
-                modifier = Modifier.fillMaxHeight(),
+                modifier = Modifier
+                    .fillMaxHeight()
+                    .padding(horizontal = BpkSpacing.Base),
```
bvitaliyg, line 71: The padding was previously provided by Snackbar

bvitaliyg (commented):
Added clarifying comments

marianeum (commented):
if there was an alignment issue how come this wasn't covered by tests? can we add further tests to ensure this is caught?

bvitaliyg:
> if there was an alignment issue how come this wasn't covered by tests? can we add further tests to ensure this is caught?

The issue was flaky and it was only appearing in certain conditions. My assumption it was attached to OS version or screen characteristics somehow.

marianeum:
@bvitaliyg were you able to reproduce it and verify that it is fixed? Is this something we need to watch out for in other migrations?

bvitaliyg:
@marianeum it uses a custom layout inside I think that was what caused it. I was able to reproduce and fix it. I think we should be fine in general

PR #1581, opened 2023-05-04
Implemented the fix for Calendar2 weekdays style for some Chinese languages

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
+++ b/backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
@@ -47,7 +47,7 @@ data class CalendarParams(
     val range: ClosedRange<LocalDate> = LocalDate.now()..LocalDate.now().plusYears(1),
     val cellsInfo: Map<LocalDate, CellInfo> = emptyMap(),
     val locale: Locale = Locale.getDefault(),
-    val dayOfWeekText: TextStyle = TextStyle.NARROW,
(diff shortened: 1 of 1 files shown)


backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -152,3 +152,13 @@ enum class CellStatusStyle {
      */
     Label,
 }
+
+private fun findBestWeekdayStyleForLocale(locale: Locale): TextStyle =
+    when (locale.language.lowercase()) {
+        "zh" -> when (locale.country.lowercase()) {
+            "mo", "hk", "tw", "sg" -> TextStyle.SHORT // Traditional
```
marianeum, line 159: can you also add a comment why we're doing this here?

bvitaliyg: In addition to what we done in Kotlin Docs?

marianeum: yeah - kotlin docs is for users modifying the format externally, but here for whoever reads the code and wonders why this is there

bvitaliyg: Done

marianeum (approved):

PR #1587, opened 2023-05-09
Switched Card to MD3

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/card/internal/BpkCardImpl.kt
```diff
@@ -41,39 +42,55 @@ import net.skyscanner.backpack.compose.tokens.BpkSpacing
...
             BpkCardElevation.Focus -> BpkTheme.colors.surfaceElevated
             BpkCardElevation.None, BpkCardElevation.Default -> BpkTheme.colors.surfaceDefault
         },
-    ).value
+        label = "BpkCard background color",
+    )
+    return CardDefaults.cardColors(
+        containerColor = value,
+        contentColor = BpkTheme.colors.textPrimary,
+        disabledContainerColor = BpkTheme.colors.surfaceDefault,
```
marianeum, line 70: wouldn't `disabledContainerColor` just match `containerColor`?

bvitaliyg: It does match – see the when statement above. The reason why we don't use the value for disabled is that we don't want to use elevated color if the card is disabled and elevated.

marianeum: It feels like quite unexpected behaviour as we've got cards that are on top of `surfaceDefault` and if they're disabled they'll blend into the background. we don't have a disabled style for cards defined and as the text colour matches the enabled one the container colour should too to avoid confusion

bvitaliyg: Perhaps we can remove this parameter in the first place?

bvitaliyg: Done

PR #1588, opened 2023-05-11
Switched CardWrapper to MD3

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardwrapper/BpkCardWrapper.kt
```diff
@@ -42,10 +44,13 @@ fun BpkCardWrapper(
     corner: BpkCardCorner = BpkCardCorner.Small,
     elevation: BpkCardElevation = BpkCardElevation.Default,
 ) {
-    Card(
-        modifier = modifier.fillMaxWidth().border(width = BpkBorderSize.Lg, color = backgroundColor, shape = cardShape(corner)),
-        backgroundColor = backgroundColor,
+    Surface(
```
bvitaliyg, line 47: This component is more suited for this case that all these custom modifiers + card (with different colours and elevations

marianeum: this also had the default elevation before (which was wrong), but based on figma the card wrapper should have the same card elevation levels set

marianeum: correct. but in this case we're changing how it looks like already, so we might as well fix the behaviour at the same time

marianeum (commented):
can you also regenerate the docs screenshots please?

bvitaliyg:
@marianeum done

marianeum (approved):

PR #1629, opened 2023-06-19
LMN-148 Add singleSelectChipGroup component

LokmaneKrizou:

+ [x] Component `README.md`
+ [x] Tests


docs/compose/SingleSelectChipGroup/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Usage
+
+Example of a SingleSelectChipGroup:
+
+```Kotlin
+import net.skyscanner.backpack.compose.singleselectchipgroup.BpkSingleSelectChipGroup
+
+BpkSingleSelectChipGroup(
+    // TODO
+)
```
LokmaneKrizou, line 38: to add info here

app/src/main/java/net/skyscanner/backpack/demo/components/SingleSelectChipGroupComponent.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.skyscanner.backpack.demo.components
+
+import net.skyscanner.backpack.meta.ComponentMarker
+
+@ComponentMarker("SingleSelectChipGroup")
+annotation class SingleSelectChipGroupComponent
```
marianeum, line 24: While we have a distinction between single & multi select the main component is still `ChipGroup` - we should have both single & multi select components in the same package & story. You can rename this marker annotation to `ChipGroupComponent` (and the name in the annotation to `Chip Group` as that's the readable name), as well as the story and then rename the package of the component to `chipgroup`

docs/compose/SingleSelectChipGroup/README.md
```diff
@@ -0,0 +1,39 @@
+# SingleSelectChipGroup
```
marianeum, line 1: also here let's have a single readme for both and name it `Chip Group` here (plus update the folder name to `ChipGroup`). make sure to update the package references in the file as well

marianeum (commented):
nice one! mainly some naming things i should've caught earlier. Also please generate the screenshots as described [here](https://github.com/Skyscanner/backpack-android/blob/main/CONTRIBUTING.md#docs-screenshots) to ensure the font matches. there is also a confluence page which describes how to install the correct font and where to find it

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/singleselectchipgroup/BpkSingleSelectChipGroup.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+}
+
+@Composable
+private fun ChipItem(chip: ChipItem, selected: Boolean, style: BpkChipStyle, onSelectedChange: () -> Unit) {
+    BpkChip(
+        modifier = Modifier
+            .padding(PaddingValues(BpkSpacing.Sm))
+            .semantics { role = Role.RadioButton },
```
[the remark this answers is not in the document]
LokmaneKrizou: when I tested the accessibility, it doesn't mention that its a radio button, it only says that its unselected item in a list and double press to toggle it, however if I set it as a checkbox/switch it mentions the type of the chip as a checkbox/switch. do you think that's enough? otherwise we may need to add a content description to explain more.

app/src/main/java/net/skyscanner/backpack/demo/compose/SingleSelectChipGroupStory.kt
```diff
@@ -0,0 +1,142 @@
...
+
+@Composable
+@SingleSelectChipGroupComponent
+@ComposeStory("On Image")
+fun SingleSelectChipGroupStoryOnImage(modifier: Modifier = Modifier) =
+    SingleSelectChipGroupDemo(BpkChipStyle.OnImage, modifier)
+
+@Composable
+@SingleSelectChipGroupComponent
+internal fun SingleSelectChipGroupDemo(style: BpkChipStyle, modifier: Modifier = Modifier) {
```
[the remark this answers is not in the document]
marianeum: yeah i think the default, on dark, on image variants we can have on a single page (just like we do now for single vs rail), so no need to navigate deeper. single vs multi select you can also just set as the name (ie `Single Select Wrap` & `Single Select Rail`) rather than specific navigation items

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/singleselectchipgroup/BpkSingleSelectChipGroup.kt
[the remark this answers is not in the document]
marianeum: got it - if that's how the built in component are announcing it let's leave it at that rather than making things custom. interesting that they don't announce it, though!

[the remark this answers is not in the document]
LokmaneKrizou: I remember adding that on Friday , but it kept saying the same sentence, no difference, I will add it in case it is treated differently when using different language than English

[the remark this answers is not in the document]
marianeum: I think it might be an OS specific thing - I'm testing on an API 33 pixel 4 emulator and it announces radiobutton + item x of y

app/src/main/java/net/skyscanner/backpack/demo/compose/SingleSelectChipGroupStory.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+@Composable
+@SingleSelectChipGroupComponent
+
+internal fun SingleSelectChipGroupSample(
+    type: ChipGroupType,
+    modifier: Modifier = Modifier,
+    selectedIndex: Int = -1,
```
marianeum, line 108: something i didn't spot before - we shouldn't allow a non-selected item aka -1/out of bounds - you always need a selected item for the single select group as otherwise it's not a valid state.

app/src/main/java/net/skyscanner/backpack/demo/compose/ChipGroupStory.kt
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+internal fun SingleSelectChipGroupSample(
+    type: BpkSingleChipGroupType,
+    modifier: Modifier = Modifier,
+    defaultIndex: Int = -1,
```
marianeum, line 115: I commented on this in the previous version, but I think it got lost. Let's remove the `-1` case (and non-selected tests) as it isn't really a supported case to _not_ have a selected item.

docs/compose/ChipGroup/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Usage
+
+Example of a Rail SingleSelectChipGroup:
+
+```Kotlin
+import net.skyscanner.backpack.compose.chipgroup.BpkSingleSelectChipGroup
+import net.skyscanner.backpack.compose.chipgroup.single.BpkSingleChipGroupType
+import net.skyscanner.backpack.compose.chipgroup.single.BpkSingleChipItem
+
+var selectedIndex by remember { mutableStateOf(-1) }
```
marianeum, line 33: also here let's use `0` as the default

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/BpkChipGroup.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.chip.BpkChipStyle
+import net.skyscanner.backpack.compose.chipgroup.single.BpkSingleChipGroupType
+import net.skyscanner.backpack.compose.chipgroup.single.BpkSingleChipItem
+import net.skyscanner.backpack.compose.chipgroup.single.BpkSingleSelectChipGroupImpl
+
+@Composable
+fun BpkSingleSelectChipGroup(
```
marianeum, line 29: if we're moving things into a separate `single` package (which is a good idea!)) let's also move the component itself otherwise it's a little confusing

marianeum (commented):
looking mostly good - just a couple of minor things

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/single/BpkSingleSelectChipGroupImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+                        onItemClicked.invoke(chips[index])
+                    }
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun ChipItem(chip: BpkSingleChipItem, selected: Boolean, style: BpkChipStyle, onSelectedChange: () -> Unit) {
```
marianeum, line 85: can you add a default modifier please and use it below? there should be a lint check for this, not sure why it's not triggering 🤔

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/single/BpkSingleSelectChipGroupImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.role
+import androidx.compose.ui.semantics.semantics
+import net.skyscanner.backpack.compose.chip.BpkChip
+import net.skyscanner.backpack.compose.chip.BpkChipStyle
+import net.skyscanner.backpack.compose.icon.BpkIcon
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+
+enum class BpkSingleChipGroupType {
```
marianeum, line 39: sorry hopefully final comment haha we use the `Rail` and `Wrap` format for the naming of enums rather than all caps

marianeum (approved):
nice one :shipit: you'll just have to push an empty commit to trigger CI again and then we can merge it

app/src/androidTest/java/net/skyscanner/backpack/compose/chipgroup/BpkSingleSelectChipGroupTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import net.skyscanner.backpack.demo.compose.SingleSelectChipGroupSample
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@RunWith(Parameterized::class)
+class BpkSingleSelectChipGroupTest(private val type: BpkSingleChipGroupType) : BpkSnapshotTest(listOf(type)) {
+
+    @Test
+    @Variants(BpkTestVariant.Default, BpkTestVariant.DarkMode)
```
marianeum, line 36: sorry one final thing that i didn't spot before - can you run this test on all variants please? (basically just remove this annotation). that way we're also testing rtl. it's fine for the others as those are just colour changes, but we should always have at least one test that runs on all variants for components

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/single/internal/BpkSingleSelectChipGroupImpl.kt
```diff
@@ -0,0 +1,102 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.role
+import androidx.compose.ui.semantics.semantics
+import net.skyscanner.backpack.compose.chip.BpkChip
+import net.skyscanner.backpack.compose.chip.BpkChipStyle
+import net.skyscanner.backpack.compose.icon.BpkIcon
+import net.skyscanner.backpack.compose.tokens.BpkSpacing
+
+enum class BpkSingleChipGroupType {
```
marianeum, line 39: actually sorry can you move the enums/data classes out of the internal folder and into the main component file as they're not internal?

PR #1636, opened 2023-06-22
LMN-23 multi select chip

LokmaneKrizou:

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/multiple/internal/BpkMultiSelectChipGroupImpl.kt
```diff
@@ -0,0 +1,190 @@
...
+
+@Composable
+private fun StickyChipItem(
+    chip: BpkMultiChipItem,
+    style: BpkChipStyle,
+    modifier: Modifier = Modifier,
+) {
+    val interactionSource = remember { MutableInteractionSource() }
+    when (chip.type) {
+        BpkChipType.Selectable, BpkChipType.Dropdown -> BpkChipImpl(
```
marianeum, line 139: i would actually say let's set the type to `Selectable` for sticky chip items. if we ever support tablets properly that means we can maintain the same API as right now and add support to it, but on mobile dropdown + dismissible makes no sense for the filter one (and dismissible in general tbf)

docs/compose/ChipGroup/README.md
```diff
@@ -1,27 +1,29 @@
 # ChipGroup
 
 [![Maven Central](https://img.shields.io/maven-central/v/net.skyscanner.backpack/backpack-compose)](https://search.maven.org/artifact/net.skyscanner.backpack/backpack-compose)
-[![Class reference](https://img.shields.io/badge/Class%20reference-Android-blue)](https://backpack.github.io/android/backpack-compose/net.skyscanner.backpack.compose.chipgroup)
-[![Source code](https://img.shields.io/badge/Source%20code-GitHub-lightgrey)](https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup)
 
-## Single Select Rail
+## Single Select
+[![Class reference](https://img.shields.io/badge/Class%20reference-Android-blue)](https://backpack.github.io/android/backpack-compose/net.skyscanner.backpack.compose.chipgroup.single)
```
marianeum, line 6: you can also just link to the parent package here and that'll cover both. at some point in the future we wanna look at how we could automate parts of the readme, so keeping the general structure makes sense. we always have the screenshots at the top for all examples, then code examples

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/multiple/internal/BpkMultiSelectChipGroupImpl.kt
LokmaneKrizou: > i would actually say let's set the type to Selectable for sticky chip items. if we ever support tablets properly that means we can maintain the same API as right now and add support to it

Do you mean we should remove the when case statement and  just set the type to be always `Selectable` and ignore Sticky chip item given type?  

> but on mobile dropdown + dismissible makes no sense for the filter one (and dismissible in general tbf)

Yep I agree, there is no use case I can think of for dismissible chips in chip group.  maybe we should use a different enum class that has only selectable and drop down types and map them to the internal types instead?

marianeum: Yeah thinking about it it probably makes sense to have a separate data class for the sticky chip which doesn't have a type (we can always add it later on). Internally we can map it to the other type

docs/compose/ChipGroup/README.md
```diff
@@ -56,3 +74,59 @@ BpkSingleSelectChipGroup(
...
+```Kotlin
+import net.skyscanner.backpack.compose.chipgroup.multiple.BpkMultiChipGroupType
+import net.skyscanner.backpack.compose.chipgroup.multiple.BpkMultiChipItem
+import net.skyscanner.backpack.compose.chipgroup.multiple.BpkMultiSelectChipGroup
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+
+val state = viewModel.uiState.collectAsStateWithLifecycle()
+
+BpkMultiSelectChipGroup(
+    chips = state.value,
```
marianeum, line 89: rather than `state.value` can you define a list of chip items with a static selected variable? 
you can add a note at the top that you'd generally drive these from a viewmodel. we generally keep our components agnostic from architecture etc

marianeum (commented):
looks good! just a tiny nit around the readme

marianeum (approved):

PR #1641, opened 2023-06-29
LMN-28: Add select component.

karljiang1:
[LMN-28](https://skyscanner.atlassian.net/browse/LMN-28)
Add select component.
Some screenshots for dropdown list as it hard to be captured automatically.
|light|dark|
|----|-----|
|image|image|

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/BpkSelect.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    object Default : BpkSelectState
+
+    object Disabled : BpkSelectState
+
+    data class Error(val message: String) : BpkSelectState
+}
+
+@Composable
+fun BpkSelect(
```
marianeum, line 35: we also need the other version which is the "self serve" one where we don't need to handle the option selection

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,108 @@
...
+    var expanded by remember { mutableStateOf(false) }
+    var selectedOptionIndex by remember { mutableStateOf(selectedIndex) }
+    val initialText = if (selectedOptionIndex > -1 && options.size >= selectedOptionIndex) options[selectedOptionIndex] else ""
+    var selectedText by remember { mutableStateOf(initialText) }
+
+    ExposedDropdownMenuBox(
+        expanded = expanded,
+        onExpandedChange = { expanded = !expanded },
+    ) {
+        BpkFieldSet(status = mapSelectState(state)) {
```
marianeum, line 72: rather than including `FieldSet` in the component it is a wrapper around select/field set - just adding an example in the `FieldSet` story with select should be good

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/BpkSelect.kt
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package net.skyscanner.backpack.compose.select
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.select.internal.BpkSelectImpl
+
+sealed interface BpkSelectState {
```
marianeum, line 25: we can just reuse the state from `FieldSet` - I know we technically don't have the validated state here, but otherwise the API between fieldset/select will be misaligned and we won't be able to pass through the state between fieldset & select

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,108 @@
...
+internal fun BpkSelectImpl(
+    options: List<String>,
+    selectedIndex: Int,
+    placeholder: String,
+    modifier: Modifier = Modifier,
+    state: BpkSelectState = BpkSelectState.Default,
+    onSelectionChange: ((selectedIndex: Int) -> Unit)? = null,
+) {
+    var expanded by remember { mutableStateOf(false) }
+    var selectedOptionIndex by remember { mutableStateOf(selectedIndex) }
```
marianeum, line 64: this should be handled outside the component as we get the `selectedIndex` from there to avoid inconsistencies of state

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/BpkTextField.kt
```diff
@@ -104,123 +60,3 @@ fun BpkTextField(
         interactionSource = interactionSource,
     )
 }
-
-@OptIn(ExperimentalAnimationApi::class)
-@Composable
-fun BpkTextField(
```
marianeum, line 110: this is a public API, so we can't just remove it unfortunately - what's the reason for removing it?

docs/compose/Select/README.md
```diff
@@ -0,0 +1,58 @@
...
+|------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/Select/screenshots/default.png" alt="Select component" width="375" />  | <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/Select/screenshots/default_dm.png" alt="Select component - dark mode" width="375" />              |
+| <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/Select/screenshots/dropdownlist.png" alt="Select component dropdownlist" width="375" /> | <img src="https://raw.githubusercontent.com/Skyscanner/backpack-android/main/docs/compose/Select/screenshots/dropdownlist_dm.png" alt="Select component dropdownlist - dark mode" width="375" /> |
```
marianeum, line 12: All our screenshots should be automatically generated to ensure they are up to date, so we shouldn't have manually created ones

app/src/androidTest/java/net/skyscanner/backpack/compose/select/BpkSelectTest.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.skyscanner.backpack.compose.BpkSnapshotTest
+import net.skyscanner.backpack.demo.compose.SelectStory
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class BpkSelectTest : BpkSnapshotTest() {
+
+    @Test
+    fun default() = snap {
```
marianeum, line 31: we generally try to have separate tests for each state, rather than the full story to ensure screenshot failures will highlight what exactly failed.
the error state with text should be part of the `field set` tests as described further down
we should also have a test covering the expanded state. you can check the `BpkBarChartTest` selected test on how to use espresso assertions - i believe you should be able to utilise that with providing a click action

app/src/main/java/net/skyscanner/backpack/demo/compose/SelectStory.kt
```diff
@@ -0,0 +1,91 @@
...
+            ErrorSelectSample()
+        }
+    }
+}
+
+@Composable
+internal fun DefaultSelectSample(modifier: Modifier = Modifier) {
+    BpkSelect(
+        modifier = modifier.widthIn(min = BpkSpacing.Xxl.times(5)),
+        options = arrayListOf("Karachi", "Lahore", "Faisalabad", "Islamabad", "Quetta", "Peshawar", "Menu item", "Menu item", "Menu item", "Menu item", "Menu item", "Menu item", "Menu item", "Menu item", "Menu item", "Menu item"),
```
marianeum, line 64: we generally try to have string resources, rather than hardcoded strings. you should be able to use the ones from chip group (city_london` etc)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/BpkSelect.kt
karljiang1: Sorry, I have not got it. Could you explain it more?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
karljiang1: thanks. I don't know if that works but I could have a try.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
karljiang1: Yes, I have made it in the initial version. But if user don't refresh the selectedIndex, we will get the weird selected status.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
karljiang1: It seems that the ExposedDropdownMenu need a parent component, and TextField component need anchor itself to the ExposedDropdownMenuBox component.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
marianeum: yep that's fine and how all of our components work - it ensures that consumers are using it correctly and update the state

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/BpkSelect.kt
marianeum: there's two API tables in Figma - we need to implement both. One is this one and the other one is for just having a click listener and the actual selection etc is done by the consumer

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
marianeum: Yep the dropdown menu can still use the text field as the anchor - we can take a look at this in pairing

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
karljiang1: Reasonable, will update it.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/BpkTextField.kt
karljiang1: Sorry, will get it back.

docs/compose/Select/README.md
karljiang1: I want to show the screen with dropdownlist but don't have a good way.

karljiang1: Will remove them.

app/src/androidTest/java/net/skyscanner/backpack/compose/select/BpkSelectTest.kt
karljiang1: ok

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/BpkSelect.kt
karljiang1: will add another version without options.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+    placeHolder: String,
+    text: String,
+    modifier: Modifier = Modifier,
+    state: BpkFieldStatus = LocalFieldStatus.current,
+    onClick: (() -> Unit)? = null,
+) {
+    var selectText by remember { mutableStateOf(text) }
+
+    BpkTextFieldImpl(
+        modifier = if (state == BpkFieldStatus.Disabled) modifier else modifier.clickable(bounded = true, role = null) {
```
karljiang1, line 109: @marianeum I don't know what role should I set here. Does it work as a **button**?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+    options: List<String>,
+    selectedIndex: Int,
+    placeholder: String,
+    modifier: Modifier = Modifier,
+    state: BpkFieldStatus = LocalFieldStatus.current,
+    onSelectionChange: ((selectedIndex: Int) -> Unit)? = null,
+) {
+    var expanded by remember { mutableStateOf(false) }
+    val initialText = if (selectedIndex > -1 && options.size >= selectedIndex) options[selectedIndex] else ""
+    var selectedText by remember { mutableStateOf(initialText) }
```
marianeum, line 57: similar to the index we shouldn't need to remember this as it is covered by the `selectedIndex` and could get out of sync otherwise

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
marianeum: yeah button makes sense in this case as we don't know what happens afterwards

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+                    contentPadding = ExposedDropdownMenuDefaults.ItemContentPadding,
+                )
+            }
+        }
+    }
+}
+
+@Composable
+internal fun BpkSelectImpl(
+    placeHolder: String,
```
marianeum, line 100: nitpick: the `h` should be lowercase

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+import net.skyscanner.backpack.compose.utils.clickable
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+internal fun BpkSelectImpl(
+    options: List<String>,
+    selectedIndex: Int,
+    placeholder: String,
+    modifier: Modifier = Modifier,
+    state: BpkFieldStatus = LocalFieldStatus.current,
```
marianeum, line 52: looking at this now i think it makes sense to deviate slightly from the API and rename this to `status` to be consistent with the field + fieldset (same goes for the other version)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+
+@Composable
+internal fun BpkSelectImpl(
+    placeHolder: String,
+    text: String,
+    modifier: Modifier = Modifier,
+    state: BpkFieldStatus = LocalFieldStatus.current,
+    onClick: (() -> Unit)? = null,
+) {
+    var selectText by remember { mutableStateOf(text) }
```
marianeum, line 106: same here - we shouldn't need this

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
```diff
@@ -0,0 +1,242 @@
...
+        var lastIcon by remember { mutableStateOf<Pair<BpkIcon, Color>?>(null) }
+        if (trailingIcon != null) {
+            BpkIcon(
+                icon = trailingIcon,
+                contentDescription = null,
+                size = BpkIconSize.Small,
+                tint = animateColorAsState(
+                    when (status) {
+                        is BpkFieldStatus.Disabled -> BpkTheme.colors.textDisabled
+                        else -> BpkTheme.colors.textSecondary
```
marianeum, line 212: Looking at figma this should be `textPrimary`

app/src/main/java/net/skyscanner/backpack/demo/compose/SelectStory.kt
```diff
@@ -0,0 +1,174 @@
...
+            horizontalArrangement = Arrangement.Start,
+        ) {
+            ErrorSelectSample()
+        }
+    }
+}
+
+@Composable
+@SelectComponent
+@ComposeStory("TextBox")
```
marianeum, line 64: nitpick on naming - `TextBox` makes it sounds like you can enter text. maybe `Select box only`

marianeum (commented):
very nice - just a couple of minor comments

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/select/internal/BpkSelectImpl.kt
karljiang1: make sense, will update it.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
karljiang1: good catch, thanks.

app/src/main/java/net/skyscanner/backpack/demo/compose/SelectStory.kt
karljiang1: nice.

marianeum (approved):

PR #1650, opened 2023-07-07
KOA-6043 Compose map markers

marianeum:
Early version of the map markers. Haven't done anything around testing etc yet
Current issues:
- Elevation is not working. That's probably due to software rendering, so we probably just have to live with it.
- Caching of the bitmap markers. Not sure what the best approach is for that - I added a very basic/dumb version that is almost certainly wrong. I found that they are[ looking at](https://github.com/googlemaps/android-maps-compose/issues/41) supporting composables for markers in the future, which is good. It is implemented for clustering - don't think we can utilise that: https://github.com/googlemaps/android-maps-compose/pull/258/files

+ [ ] Component `README.md`
+ [ ] Tests


mosmb:
Very nice!
I did exactly the same but I encountered some performance issue, so I wonder if it's the case for you. 
Also, what's your current logic in your app for that? Is your map based on Compose? If yes, how do you handle custom markers?

app/screenshots/oss/debug/dm/net.skyscanner.backpack.compose.map.BpkIconMapMarkerTest_default.png
bvitaliyg, line 1: This one looks wrong, no?

app/screenshots/oss/debug/rtl/net.skyscanner.backpack.compose.map.BpkIconMapMarkerTest_default.png
bvitaliyg, line 1: This one as well

app/src/main/java/net/skyscanner/backpack/demo/compose/MapStory.kt
```diff
@@ -0,0 +1,172 @@
...
+            state = rememberMarkerState(position = MARKER_POSITION_4),
+        )
+    }
+}
+
+@Composable
+@MapMarkersComponent
+@ComposeStory(kind = StoryKind.DemoOnly, name = "Pointer")
+fun PointerMapMarkerStory(modifier: Modifier = Modifier) {
+    val cameraPositionState = rememberCameraPositionState {
```
bvitaliyg, line 142: Why do we want to remember camera position state?

backpack-compose/build.gradle
```diff
@@ -54,8 +54,10 @@ dependencies {
     api libs.compose.foundation
     api libs.compose.uiToolingPreview
     api libs.compose.uiUtil
-    implementation libs.compose.material3
+    api 'com.google.maps.android:maps-compose:2.11.4'
```
bvitaliyg, line 57: Can we use toml file for these dependencies?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPointerMapMarker.kt
```diff
@@ -0,0 +1,117 @@
...
+        icon = iconBitmap,
+        onClick = onClick,
+        onInfoWindowClick = onInfoWindowClick,
+    ) {
+        PriceMarkerLayout(title = title, status = BpkPriceMarkerStatus.Focused)
+    }
+}
+
+@Composable
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
bvitaliyg, line 70: Are we doing this for tests? Is there a way to capture the entire Map instead?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPriceMapMarker.kt
```diff
@@ -0,0 +1,145 @@
...
+                    flareHeight = 6.dp,
+                    pointerDirection = BpkFlarePointerDirection.Down,
+                ),
+            )
+            .padding(bottom = 6.dp),
+    ) {
+        Box(
+            contentAlignment = Alignment.Center,
+            modifier = Modifier
+                .height(if (status == BpkPriceMarkerStatus.Focused) 32.dp else 24.dp)
```
bvitaliyg, line 124: Can we use tokens instead?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/internal/IconMarkerShape.kt
```diff
@@ -0,0 +1,58 @@
...
+    object : Shape {
+        override fun createOutline(size: Size, layoutDirection: LayoutDirection, density: Density): Outline =
+            Outline.Generic(
+                path = Path().apply {
+                    addPath(size)
+                },
+            )
+    }
+
+private fun Path.addPath(
```
bvitaliyg, line 38: Can we name it in a less generic way? E.g. addFlarePath()

docs/compose/ChipGroup/screenshots/single-select-wrap.png
bvitaliyg, line 1: What happened with these?

docs/compose/Maps/screenshots/pointer.png
bvitaliyg, line 1: Are these screenshots automated? I wonder why it has rounded borders

docs/compose/Maps/screenshots/pointer_dm.png
bvitaliyg, line 1: Is there a DM flag for GoogleMap we can turn on?

docs/view/Text/screenshots/hero_dm.png
bvitaliyg, line 1: If it's not intentional, lets revert it

bvitaliyg (commented):
Have some questions and suggestions

app/screenshots/oss/debug/default/net.skyscanner.backpack.compose.map.BpkIconMapMarkerTest_default.png
marianeum, line 1: i'm not sure where the line is coming from - it's not showing up on the device and the shape is closed correctly, so not sure where this is coming from

app/src/main/java/net/skyscanner/backpack/demo/compose/MapStory.kt
marianeum: we don't - removed

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPointerMapMarker.kt
bvitaliyg: But we can add these keys, right? Especially now, when your snapshots are could-based. Maybe in another ticket?

app/src/main/java/net/skyscanner/backpack/demo/compose/MapStory.kt
```diff
@@ -0,0 +1,161 @@
...
+import net.skyscanner.backpack.meta.StoryKind
+
+@Composable
+@MapMarkersComponent
+@ComposeStory(kind = StoryKind.DemoOnly, name = "Price")
+fun PriceMapMarkerStory(modifier: Modifier = Modifier) {
+    var focusedMarker by remember { mutableStateOf(0) }
+    var viewedMarkers by remember { mutableStateOf(setOf(1)) }
+
+    fun markerStatus(index: Int): BpkPriceMarkerStatus {
```
bvitaliyg, line 53: Expression return?

app/src/main/java/net/skyscanner/backpack/demo/compose/MapStory.kt
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+@Composable
+@MapMarkersComponent
+@ComposeStory(kind = StoryKind.DemoOnly, name = "Icon")
+fun IconMapMarkerStory(modifier: Modifier = Modifier) {
+    var focusedMarker by remember { mutableStateOf(0) }
+    GoogleMap(cameraPositionState = CameraPositionState(position = MapPosition)) {
+        BpkIconMapMarker(
```
bvitaliyg, line 97: I think this can be done as a for loop especially if you combine the MarkerPositions intro to a list. You might need to wrap the BpkIconMapBarker with `key(index)` though.

app/src/main/java/net/skyscanner/backpack/demo/compose/MapStory.kt
```diff
@@ -0,0 +1,161 @@
...
+        )
+    }
+}
+
+@Composable
+@MapMarkersComponent
+@ComposeStory(kind = StoryKind.DemoOnly, name = "Pointer")
+fun PointerMapMarkerStory(modifier: Modifier = Modifier) {
+    GoogleMap(cameraPositionState = CameraPositionState(position = MapPosition)) {
+        BpkPointerMapMarker(
```
bvitaliyg, line 135: Same here

app/src/main/res/values/strings.xml
```diff
@@ -290,6 +290,13 @@
     <string name="maps_markers_badges">Badges</string>
     <string name="maps_markers_badges_with_icons">With icons</string>
     <string name="map_markers">Map markers</string>
+    <string name="map_marker_price_1">£195</string>
```
bvitaliyg, line 293: We can create an array resource, or not to create a resource at all and just use build-in currency formatters

docs/compose/Maps/screenshots/pointer.png
marianeum: same as above - the current config is set up to work for release builds only

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPointerMapMarker.kt
```diff
@@ -0,0 +1,117 @@
...
+fun rememberPointer(): BitmapDescriptor {
+    val bm = rememberPointerBitmap()
+    return remember(bm) {
```
bvitaliyg, line 73: We create the bitmap in the same way as we do with other markers?

app/src/main/java/net/skyscanner/backpack/demo/compose/MapStory.kt
marianeum: it's not just the positions, though - the icon/state is also different

app/src/main/res/values/strings.xml
marianeum: I'm not sure how this is different to any other component we have where we just use standard string resources for different examples, so i'd rather not overcomplicate things

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPointerMapMarker.kt
marianeum: this just uses the bitmap descriptor vs the composable layout as it's simple. but for the test we cannot use the bitmapdescriptor as that's map specific

bvitaliyg (approved):

PR #1656, opened 2023-07-11
Bellagio/non jira remove confirmation

jonnycaley:
If you don't enter "yes" the process is cancelled which seems counter intuitive. If all new components are expected to be designed using compose then this shouldn't need to be asked during setup, hence i'm removing it.

### DIFF
--- a/scripts/generate-component.py
+++ b/scripts/generate-component.py
@@ -31,12 +31,6 @@ else:
     package = sys.argv[2]
 
 print('')
-print(f'Create compose component structure for {package}.Bpk{name}?')
-confirmation = input("[Y/n]: ").lower()
-if confirmation != "" and confirmation != "y" and confirmation != "yes":
-    sys.exit("Component creation cancelled")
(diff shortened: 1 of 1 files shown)


scripts/generate-component.py
```diff
@@ -31,13 +31,7 @@ def generate_file(template, path, name, mapping):
     package = sys.argv[2]
 
 print('')
-print(f'Create compose component structure for {package}.Bpk{name}?')
-confirmation = input("[Y/n]: ").lower()
-if confirmation != "" and confirmation != "y" and confirmation != "yes":
-    sys.exit("Component creation cancelled")
-
-print('')
-print('Creating compose component structure...')
```
olliecurtis, line 40: I guess this is useful as a final sanity check before comitting it to make sure the name you entered is correct and no issues with spelling.

So would fall on the side of keeping it to make sure its a final confirmation to make sure all is ok with the naming :)

PR #1658, opened 2023-07-20
LMN-26 add progress bar component

jonnycaley:
Adds progress bar as per designs [here](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?type=design&node-id=8192-8826&mode=design&t=NpsTDXDDpCd3E0Gv-0)

https://github.com/Skyscanner/backpack-android/assets/26006896/e7959783-48e8-4c80-a87c-2eef1a6fc5f7


.idea/inspectionProfiles/ktlint.xml
```diff
@@ -8,13 +8,15 @@
     </inspection_tool>
     <inspection_tool class="PreviewApiLevelMustBeValid" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
```
bvitaliyg, line 11: Are these changes intentional?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,141 @@
...
+    max: Int = 1,
+    stepped: Boolean = false,
+) {
+    val height = if (size == BpkProgressBarSize.Small) 8.dp else 16.dp
+
+    val progress = value
+        .coerceIn(min.toFloat(), max.toFloat())
+        .div(max.toFloat())
+
+    val animatedProgress by animateFloatAsState(
```
bvitaliyg, line 64: Noice!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,141 @@
...
+                    height = height,
+                    dividers = dividers,
+                )
+            }
+        }
+    }
+}
+
+@Composable
+private fun BuildDividers(
```
bvitaliyg, line 95: I think it's the best to do with onDraw modifier

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,141 @@
...
+                        .background(BpkTheme.colors.surfaceDefault),
+                )
+            }
+        },
+    )
+}
+
+@BpkPreviews
+@Composable
+internal fun PreviewBpkProgressBar() {
```
bvitaliyg, line 119: We usually create the previews inside the app module, and then re-use it for tests and stories

bvitaliyg (commented):
Some small comments

.idea/inspectionProfiles/ktlint.xml
jonnycaley: Nah, it was added automatically by IDE

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
jonnycaley: Ahh, I understand now. Removed 🧹

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,114 @@
...
+
+enum class BpkProgressBarSize {
+    Small,
+    Large,
+}
+
+@Composable
+fun BpkProgressBar(
+    value: Float,
+    // accessibilityLabel: String, <- not sure if we need this, in its current state talkback reads ("zero percent progress bar")
```
jonnycaley, line 50: Not sure if we need this

app/src/main/java/net/skyscanner/backpack/demo/compose/ProgressBarStory.kt
```diff
@@ -0,0 +1,122 @@
...
+
+@Composable
+@ProgressBarComponent
+@ComposeStory
+fun ProgressBarStory(modifier: Modifier = Modifier) {
+
+    Column(modifier = modifier.padding(BpkSpacing.Lg)) {
+
+        ProgressBarSample(
+            title = "Small",
```
marianeum, line 49: nit: can you use string resources please?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
marianeum: One thing you could try is adding the `progressSemantics` modifier on the progress indicator and setting the range and value as we get them from here, rather than the transformed version. that way we should get the correct range announced and we're all good

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,114 @@
...
+    modifier: Modifier = Modifier,
+    size: BpkProgressBarSize = BpkProgressBarSize.Small,
+    min: Int = 0,
+    max: Int = 1,
+    stepped: Boolean = false,
+) {
+    val height = if (size == BpkProgressBarSize.Small) 8.dp else 16.dp
```
marianeum, line 57: can you use the `BpkSpacing` values please?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,114 @@
...
+        modifier = Modifier
+            .height(height)
+            .fillMaxWidth(),
+        horizontalArrangement = Arrangement.SpaceEvenly,
+        content = {
+            repeat(dividers) {
+                Box(
+                    modifier = Modifier
+                        .fillMaxHeight()
+                        .width(2.dp)
```
marianeum, line 108: also here `BpkBorderSize.Lg`

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
```diff
@@ -83,6 +83,14 @@ class BpkComposeComponentUsageDetector : Detector(), SourceCodeScanner {
                 "androidx.compose.foundation.text.BasicTextField",
             ),
         ),
+        PROGRESS_BAR(
```
marianeum, line 86: as we no longer expose material we only define this if there's non-material alternatives

marianeum (commented):
great first contribution! just a few minor comments

app/src/main/java/net/skyscanner/backpack/demo/compose/ProgressBarStory.kt
jonnycaley: Even for story? Why?

marianeum: we're still trying to follow general standards/best practices here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
jonnycaley: Good idea, done

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
jonnycaley: Done

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
jonnycaley: Done

backpack-lint/src/main/java/net/skyscanner/backpack/lint/check/BpkComposeComponentUsageDetector.kt
jonnycaley: Ahh, i did wonder. Done

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
```diff
@@ -0,0 +1,122 @@
...
+
+        val numberOfSteps = max
+
+        LinearProgressIndicator(
+            progress = animatedProgress,
+            modifier = Modifier
+                .height(height)
+                .fillMaxWidth()
+                .clip(CircleShape)
+                .progressSemantics(
```
marianeum, line 79: can you also include the values range? that should match the min/max values (that's mainly why they exist in the api)

jonnycaley: I left it out because we convert `progress` to a float between 0..1 and coerce it [here](https://github.com/Skyscanner/backpack-android/pull/1658/files/61eceb48942db0db87039c2313242bfb98ce640d#diff-4f9bd9b6cdd30c54e45a5dabb963a30753c5d7bb8a1547788135f320933c02ffR60-R62). As the default `valueRange` parameter is 0..1 we should be able to use that.

marianeum: yeah that's what i meant in my original comment with setting the raw value and the range - that way we'll announce the actual value within the min/max range

jonnycaley: Would that be as useful for someone using the screen reader? For example, in the following it will say "0% progress" when it's loaded 50% of the progress which gives off the wrong impression of how much of the data is loaded.

Screenshot 2023-07-21 at 09 21 46

marianeum: I asked @olliecurtis for help, so he will probably come back to you as I'm also not 100% sure what the right approach is here

marianeum: Ok I had a chat with @olliecurtis. We think we can remove min from the API/component as it doesn't make a lot of sense. We should still set the value & valueRange based on max here, but min doesn't make a lot of sense and it should always go from 0 (as it does visually).

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
bvitaliyg: It just is what it was designed for. It's also faster in terms of rendering

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/progressbar/BpkProgressBar.kt
jonnycaley: I agree. If we're going to change the API, I think it's worth making the following changes:

### Remove `max` and state the value will always be clamped between 0..100

This removes the overhead of "If a parse max in I need to parse a value in between 0..max and not 0..1.", which will be required if users want to add steps to the component.

### Change `value: Float` to `percentage: Int`

When the progress is an integer (this is true for [flights](StaggeredProgressBarView) and hotels - we receive `"completionPercentage": Int` in the DTO), users have to convert to a `Float`, which looks like it will be most of the time.

### Change `stepped (Boolean)` to `numberOfSteps (Int)` and set it default to 1

Asking "How many steps are there to the progress?" with a single variable is easier to understand than "The number of steps will be max - 1, and by default max is 1 so you will not see any steps even if you set `stepped == true` if you do not set max != 1".

So will end up like this:

```
@Composable
fun BpkProgressBar(
    percentage: Int, // value between 0..100
    size: BpkProgressBarSize = BpkProgressBarSize.Small,
    numberOfSteps: Int = 1, // 1..100
)
```

Wdyt?

marianeum: Had another chat with @olliecurtis and it still makes sense to support the max value for both the stepped and non-stepped version as you may have some extra information about the maximum value. for example of web we get back x providers out of y and the progress bar indicates that. 
i think having `numberOfSteps` combined with `percentage` can be equally confusing as you need to do mapping between the percentage and step count. also you may have 55.5% for example and an int may not be enough.

so to keep things in sync across platforms we will need to maintain this (min will be removed on web too).

jonnycaley: Fair enough, done

PR #1660, opened 2023-07-20
KOA-6208 Migrate the screenshot server to KTor

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


app/screenshots.gradle
```diff
@@ -23,7 +23,7 @@ android {
             managedDevices.devices {
                 Docs(ManagedVirtualDevice) {
                     device = "Pixel"
-                    apiLevel = 31
+                    apiLevel = 30
```
bvitaliyg, line 26: INSTRUMENTATION_ABORTED: System has crashed.

Had to migrate to API 30. At least it doesn't produce transparent screenshots anymore

marianeum: rather than downgrading we should try to upgrade our screenshot emulator to be on the latest version. if we're stuck on an older version because of ktor we might not want to migrate to ktor

bvitaliyg: Yep, 31 is the latest available version. And we have to use 30 otherwise the tests might crash

marianeum: then maybe ktor isn't the right way to go if it forces us to downgrade. the further we fall behind the worse

bvitaliyg: Reverted. The issue was related to my local gradle-managed device setup.

marianeum (approved):
i guess the select test is still flaky :(

bvitaliyg:
They are way less flaky but yeah, we need to have another round

PR #1665, opened 2023-07-25
Add Modal bottom sheet and update styling for existing sheet

marianeum:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
```diff
@@ -0,0 +1,98 @@
...
+@OptIn(ExperimentalMaterial3Api::class)
+fun rememberBpkModalBottomSheetState(
+    skipPartiallyExpanded: Boolean = false,
+    confirmStateChange: (BpkModalBottomSheetValue) -> Boolean = { true },
+): BpkModalBottomSheetState {
+    val delegate = rememberModalBottomSheetState(
+        skipPartiallyExpanded = skipPartiallyExpanded,
+        confirmValueChange = { confirmStateChange(it.toBpkBottomSheetValue()) },
+    )
+    return remember { BpkModalBottomSheetState(delegate) }
```
bvitaliyg, line 39: remember(delegate)

app/src/main/java/net/skyscanner/backpack/demo/compose/BottomSheetStory.kt
```diff
@@ -65,3 +69,39 @@ fun BottomSheetStory(
...
+    var openBottomSheet by rememberSaveable { mutableStateOf(true) }
+    val state = rememberBpkModalBottomSheetState()
+    Box(
+        modifier = modifier
+            .fillMaxSize()
+            .padding(top = BpkSpacing.Xxl),
+        contentAlignment = Alignment.Center,
+    ) {
+        BpkButton(text = stringResource(R.string.generic_show)) {
+            openBottomSheet = true
```
bvitaliyg, line 88: Don't we suppose to use .show() method of the state here?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
```diff
@@ -0,0 +1,98 @@
...
+
+enum class BpkModalBottomSheetValue {
+    Collapsed,
+    Expanded,
+    Hidden,
+}
+
+@Stable
+@OptIn(ExperimentalMaterial3Api::class)
+class BpkModalBottomSheetState internal constructor(
```
bvitaliyg, line 50: We also need to provide a constructor to hoist the state inside the logic layer. It should mimic the SheetState constructor.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    suspend fun show() = delegate.show()
+
+    suspend fun hide() = delegate.hide()
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+private fun SheetValue.toBpkBottomSheetValue(): BpkModalBottomSheetValue =
+    when (this) {
+        SheetValue.PartiallyExpanded -> BpkModalBottomSheetValue.Collapsed
```
bvitaliyg, line 87: Do we actually use the PartiallyExpended value?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
bvitaliyg: Also, can we reuse the state/the value from non-modal version?

bvitaliyg (commented):
Have some questions/requests

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
marianeum: we're not returning the delegate - we're returning the `BpkModalBottomSheetState` - it's the same in the non-modal version

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
marianeum: > We also need to provide a constructor to hoist the state inside the logic layer. It should mimic the SheetState constructor.

Not sure what exactly you mean - I followed the same approach as on the non-modal version

> Also, can we reuse the state/the value from non-modal version?
They have different functionality/values (ie this one has hidden), so no

app/src/main/java/net/skyscanner/backpack/demo/compose/BottomSheetStory.kt
marianeum: show/hide behaves differently - both from an animation case (the bottom sheet is not coming in from the bottom) , but also it just hides it, but still retains the screen focus. even when using `state.isVisible` to decide to create the bottom sheet the behaviour is still wrong. i'm not sure when the show/hide functions _should_ actually be used tbh

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheetState.kt
marianeum: again i followed the same pattern as in the non-modal version - can you clarify what you mean here?

app/src/androidTest/java/net/skyscanner/backpack/compose/BpkSnapshotTest.kt
```diff
@@ -91,11 +89,7 @@ open class BpkSnapshotTest(private val tags: List<Any> = emptyList()) : Screensh
         scenario.onActivity { activity ->
             activity.setContent {
                 @Suppress("DEPRECATION")
-                activity.window.clearFlags(
-                    FLAG_TRANSLUCENT_STATUS or
-                        SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN or
-                        SYSTEM_UI_FLAG_LAYOUT_STABLE,
-                )
+                activity.window.addFlags(SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN)
```
marianeum, line 92: with the bottom sheet the modal takes up the whole screen, so we need to make sure we don't have a nav bar as otherwise the snapshot will change everytime (as the time will be different). this will probably affect a few other tests that are at full height

bvitaliyg (approved):

PR #1666, opened 2023-07-25
Kotlin 1.9, Compose-Compiler 1.5, Gradle 8, AGP 8

bvitaliyg:

+ [ ] Component `README.md`
+ [ ] Tests


Backpack/src/main/AndroidManifest.xml
```diff
@@ -1 +0,0 @@
-<manifest />
```
bvitaliyg, line 1: No need to have these empty manifests anymore

android-configuration.gradle
```diff
@@ -25,11 +25,11 @@ android {
...
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
     kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_1_8.toString()
+        jvmTarget = JavaVersion.VERSION_17.majorVersion
```
bvitaliyg, line 32: "17"

app/build.gradle
```diff
@@ -31,6 +31,7 @@ if (rootProject.file("local.properties").exists()) {
 }
 
 android {
+    namespace 'net.skyscanner.backpack'
```
bvitaliyg, line 34: New requirement of AGP8

buildSrc/src/main/kotlin/net/skyscanner/backpack/tokens/BpkIcons.kt
```diff
@@ -72,7 +72,7 @@ data class BpkIcon(
                     }
                     folder.listFiles()!!.map { file ->
                         val stream = ByteArrayOutputStream()
-                        Svg2Vector.parseSvgToXml(file, stream)
+                        Svg2Vector.parseSvgToXml(file.toPath(), stream)
```
bvitaliyg, line 75: Svg2Vector breaking change – it now requires Path instead of File

gradle.properties
```diff
@@ -16,3 +16,4 @@ org.gradle.jvmargs=-Xmx4096m
 # org.gradle.parallel=true
 android.useAndroidX=true
 android.enableJetifier=true
+android.nonTransitiveRClass=false
```
bvitaliyg, line 19: Another AGP breaking behavioral change – it doesn't generate transitive R classes by default anymore to squeeze the performance. We need to have to in order to share icon ids from the common module

gradle/libs.versions.toml
```diff
@@ -1,13 +1,13 @@
 [versions]
-kotlin = "1.8.21"
-ksp = "1.8.21-1.0.11" # depends on Kotlin version
-composeCompiler = "1.4.7" # depends on Kotlin version
+kotlin = "1.9.0"
+ksp = "1.9.0-1.0.12" # depends on Kotlin version
+composeCompiler = "1.5.0" # depends on Kotlin version
 compilerTesting = "1.5.0"
 destinations = "1.8.42-beta" # update this alongside with compiler version
```
bvitaliyg, line 6: We can use 1.9 (which is stable, btw), but it brings the latest compose beta libraries which we don't want for now. They also require compileSdk=34, which we don't want as well for now

.github/workflows/_build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v3
         with:
-          java-version: '11'
+          java-version: '17'
```
bvitaliyg, line 45: Java 17 is a requirement now

.github/workflows/_build.yml
```diff
@@ -77,10 +77,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Set up JDK 11
+      - name: Set up JDK 17
```
marianeum, line 80: we should probably just remove the version from these step names

Backpack/src/main/java/net/skyscanner/backpack/skeleton/BpkShimmerOverlay.kt
```diff
@@ -56,11 +56,11 @@ class BpkShimmerOverlay @JvmOverloads constructor(
 
     private fun startShimmer() {
         // Use ObjectAnimator to draw the animation for translationX, translate the position from left to right.
-        ObjectAnimator.ofFloat(findViewById(R.id.bpk_skeleton_shimmer), "translationX", -500f, 500f).apply {
-            duration = 1000 // Per specification.
-            repeatCount = ObjectAnimator.INFINITE
-            startDelay = 200 // Per specification.
-            start()
```
bvitaliyg, line 63: New version of lint has a rule to validate these animations have actually started, and it doesn't recognise the start from apply method

marianeum (commented):
can you verify that all screenshots are correct and merge main into this to ensure the map markers also get generated correctly

app/screenshots/oss/debug/default/net.skyscanner.backpack.badge.BpkBadgeTest_default.png
bvitaliyg, line 1: Those bitmaps are actually the same. The PNGs are different, but not the image. An artifact from Shot update I suppose

marianeum (approved):

PR #1718, opened 2023-09-08
Koa-6266 Bumped to Compose 1.5 & target SDK 34

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/carousel/BpkCarouselState.kt
```diff
@@ -58,21 +58,13 @@ fun rememberBpkCarouselState(
...
     val initialPage = (Int.MAX_VALUE / 2) + initialImage
-    val pagerState = rememberPagerState(initialPage = initialPage)
+    val pageCount = if (totalImages > 1) Int.MAX_VALUE else 1 // if count > 1, set to Int.MAX_VALUE for infinite looping
+    val pagerState = rememberPagerState(initialPage = initialPage) { pageCount }
     return remember(pagerState, totalImages) {
         BpkCarouselInternalState(delegate = pagerState, totalImages = totalImages)
     }
 }
 
-@OptIn(ExperimentalFoundationApi::class)
```
bvitaliyg, line 67: Why did we remove it?

jaanainkinen: Now the `PagerState` is an abstract class and cannot be instantiated, and the only way of creating the `PagerStateImpl` is by using the remember or creating a new class which would be a completely new implementation… and if we do it as it’s suggested on the documentation. Plus the `BpkCarouselState` was not being used on the app or internally.

bvitaliyg: We created this constructor so it can be instantiated and managed by business logic classes – as Compose guidelines suggest. Let's keep this for now.

bvitaliyg (approved):

PR #1730, opened 2023-09-20
Fix chipgroup spacing

marianeum:
The spacing was originally around items, rather than between items. Now that FlowRow added support for spacing between items we can correct the component and only apply spacing between items.

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chipgroup/multiple/internal/BpkMultiSelectChipGroupImpl.kt
```diff
@@ -85,13 +85,14 @@ internal fun BpkMultiSelectChipGroupImpl(
         BpkMultiChipGroupType.Wrap -> {
             FlowRow(
                 modifier = modifier.selectableGroup(),
+                horizontalArrangement = Arrangement.spacedBy(BpkSpacing.Md),
```
marianeum, line 88: now that we're on 1.5 we can set both the horizontal & vertical arrangement, so we can actually only apply the spacing internally as is required 🎉

bvitaliyg (approved):

PR #1731, opened 2023-09-20
Allow consuming of Marker onClick behaviour and set default focused zIndex

marianeum:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkIconMapMarker.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkIconMapMarker.kt
@@ -47,6 +47,9 @@ enum class BpkIconMarkerStatus {
     Disabled,
 }
 
+/**
+ * @param onClick Callback invoked when the marker is clicked. Return true to consume the event, for example to disable the default centre map on marker behaviour.
+ */
 @Composable
 fun BpkIconMapMarker(
     contentDescription: String,
@@ -55,8 +58,8 @@ fun BpkIconMapMarker(
     state: MarkerState = rememberMarkerState(),
     tag: Any? = null,
     visible: Boolean = true,
-    zIndex: Float = 0.0f,
-    onClick: (Marker) -> Unit = {},
+    zIndex: Float? = null,
+    onClick: (Marker) -> Boolean = { false },
 ) {
     val iconBitmap = rememberCapturedComposeBitmapDescriptor(icon, status) {
         IconMarkerLayout(status = status, icon = icon)
@@ -67,9 +70,9 @@ fun BpkIconMapMarker(
         tag = tag,
         title = contentDescription,
         visible = visible,
-        zIndex = zIndex,
+        zIndex = if (status == BpkIconMarkerStatus.Focused && zIndex == null) 1.0f else zIndex ?: 0.0f,
         icon = iconBitmap,
-        onClick = { onClick(it); false },
+        onClick = onClick,
     ) {}
 }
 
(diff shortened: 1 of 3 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkIconMapMarker.kt
```diff
@@ -55,8 +58,8 @@ fun BpkIconMapMarker(
     state: MarkerState = rememberMarkerState(),
     tag: Any? = null,
     visible: Boolean = true,
-    zIndex: Float = 0.0f,
-    onClick: (Marker) -> Unit = {},
+    zIndex: Float? = null,
```
marianeum, line 61: made this nullable, so it can either be configured if needed or takes default behaviour for focused vs unfocused markers. we still want consumers to be able to change this as there might be many overlapping markers & different types of markers on the screen and we don't control the importance/priority between them, however a sensible default behaviour is needed

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkIconMapMarker.kt
```diff
@@ -55,8 +58,8 @@ fun BpkIconMapMarker(
...
+    onClick: (Marker) -> Boolean = { false },
```
marianeum, line 62: we need to allow consumers to consume the event - for example on hotels we do not want to centre the map when tapping on a marker

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkIconMapMarker.kt
```diff
@@ -67,9 +70,9 @@ fun BpkIconMapMarker(
         tag = tag,
         title = contentDescription,
         visible = visible,
-        zIndex = zIndex,
+        zIndex = if (status == BpkIconMarkerStatus.Focused && zIndex == null) 1.0f else zIndex ?: 0.0f,
```
marianeum, line 73: this sets the zIndex to 1 for a focused marker to ensure it shows on top of the others

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPriceMapMarker.kt
```diff
@@ -58,8 +58,8 @@ fun BpkPriceMapMarker(
     state: MarkerState = rememberMarkerState(),
     tag: Any? = null,
     visible: Boolean = true,
-    zIndex: Float = 0.0f,
-    onClick: (Marker) -> Unit = {},
+    zIndex: Float? = null,
```
marianeum, line 61: same changes here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/map/BpkPriceMapMarker.kt
```diff
@@ -70,9 +70,9 @@ fun BpkPriceMapMarker(
         tag = tag,
         title = title,
         visible = visible,
-        zIndex = if (status == BpkPriceMarkerStatus.Focused && zIndex == 0.0f) 1.0f else zIndex,
+        zIndex = if (status == BpkPriceMarkerStatus.Focused && zIndex == null) 1.0f else zIndex ?: 0.0f,
```
marianeum, line 73: there already was a change for the price marker to do this, but we missed it for the icon one. also aligned the nullability to be clearer about what the behaviour should be

bvitaliyg (approved):

PR #1741, opened 2023-09-28
OGM-1973 Add selectableMonthRange parameter to Calendar2

henrik-sky:
Background / discussion [here](https://skyscanner.slack.com/archives/C0JHPDSSU/p1693299807034489?thread_ts=1693298468.619219&cid=C0JHPDSSU) 

`selectableMonthRange` defaults to the months in the `range` parameter. 
Only these months show the 'Select whole month' button.

+ [ ] Component `README.md`
+ [x] Tests


marianeum (approved):

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -48,6 +50,7 @@ import java.util.Locale
 data class CalendarParams(
     val selectionMode: SelectionMode,
     val range: ClosedRange<LocalDate> = LocalDate.now()..LocalDate.now().plusYears(1),
+    val selectableMonthRange: ClosedRange<YearMonth> = range.start.yearMonth()..range.endInclusive.yearMonth(),
```
bvitaliyg, line 53: I think this should be a parameter of CalendarSelectionMode.SelectWholeMonth instead

bvitaliyg (changes requested):
One small request

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
henrik-sky: Good idea, I've updated

bvitaliyg (approved):

PR #1745, opened 2023-09-29
lumen/LMN-420 update Range slider to support custom thumb with label

LokmaneKrizou:

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/internal/BpkRangeSliderImpl.kt
```diff
@@ -0,0 +1,217 @@
...
+    modifier: Modifier = Modifier,
+    onValueChangeFinished: (() -> Unit)? = null,
+    steps: Int = 0,
+) {
+    Column(
+        modifier = modifier
+            .sliderSemantics(
+                value = value,
+                enabled = enabled,
+                description = "$sliderRangeType, $label",
```
marianeum, line 123: I think rather than combining the two it'd be better the `sliderRangeType` to be used as the `contentDescription` rather than the `stateDescription` - that when updating the range you'll only hear the state update, rather than the content description again

marianeum (approved):

PR #1756, opened 2023-10-09
LMN-525 Implement App search modal

LokmaneKrizou:
LMN-525

Implemented App search modal with its internal components, for more details checkout the screenshots attached with this PR.

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -0,0 +1,94 @@
...
+    data class Content(
+        val sections: List<Section>,
+        val shortcuts: List<Shortcut>? = null,
+    ) : AppSearchModalResult()
+
+    data class Loading(val accessibilityLabel: String) : AppSearchModalResult()
+    data class Error(
+        val title: String,
+        val description: String,
+        @DrawableRes val image: Int,
```
marianeum, line 40: I think this should be a composable node instead - this will allow consumers to also supply remote images

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.annotation.DrawableRes
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.appsearchmodal.internal.BpkAppSearchModalImpl
+import net.skyscanner.backpack.compose.icon.BpkIcon
+import net.skyscanner.backpack.compose.modal.BpkModal
+import net.skyscanner.backpack.compose.modal.rememberBpkModalState
+import net.skyscanner.backpack.compose.navigationbar.NavIcon
+
+sealed class AppSearchModalResult {
```
marianeum, line 30: we generally prefix these with the component name (ie `BpkAppSearchModalResult` in this case)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -0,0 +1,94 @@
...
+    data class Loading(val accessibilityLabel: String) : AppSearchModalResult()
+    data class Error(
+        val title: String,
+        val description: String,
+        @DrawableRes val image: Int,
+        val action: Action,
+    ) : AppSearchModalResult()
+}
+
+data class Shortcut(
```
marianeum, line 45: same for all the others

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -0,0 +1,94 @@
...
+    val text: String,
+    val icon: BpkIcon,
+    val onShortcutSelected: () -> Unit,
+)
+
+data class Section(val headings: SectionHeading? = null, val items: List<Item>)
+
+data class Item(
+    val title: String,
```
marianeum, line 54: this should be an `AnnotatedString`, so we can do the highlighting on the client side

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
```diff
@@ -0,0 +1,204 @@
...
+    Box(modifier = Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
+        BpkButton(
+            text = stringResource(R.string.generic_show),
+            onClick = { showModal.value = true },
+        )
+    }
+    if (showModal.value) {
+        BpkAppSearchModal(
+            modifier = modifier,
+            title = "Destination",
```
marianeum, line 93: can you use string resourced please? we can probably simplify by just having one copy for each item for example as the content doesn't matter much

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+@Composable
+internal fun BpkAppSearchModalImpl(
+    inputText: String,
+    inputHint: String,
+    results: AppSearchModalResult,
+    onInputChanged: (String) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    when (results) {
+        is AppSearchModalResult.Error -> {
```
marianeum, line 49: i just noticed this... we probably want to double check this with mathilda as i believe the text field should still be shown in the error case otherwise you can't start typing anything else (or if something fails while you're typing you'll get thrown into this)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+                    onValueChange = onInputChanged,
+                )
+                LazyColumn {
+                    if (results is AppSearchModalResult.Content) {
+                        results.shortcuts?.let {
+                            item {
+                                BpkShortcuts(it)
+                            }
+                        }
+                        items(results.sections) {
```
marianeum, line 71: from a lazy perspective it might be better to unbox the sections here and have the section header as a separate item and then each item as an item otherwise we don't really gain a lot from this being lazy depending on the section sizes/count

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkShortcuts.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+    BpkSingleSelectChipGroup(
+        modifier = modifier.padding(
+            top = BpkSpacing.Lg,
+            start = BpkSpacing.Base,
+        ),
+        chips = singleSelectChips,
+        selectedIndex = selectedChip.intValue,
+        onItemClicked = {
+            selectedChip.intValue = singleSelectChips.indexOf(it)
```
marianeum, line 48: i don't think it's expected behaviour that the shortcut will actually be marked as selected when clicking (we don't have any designs for selection states) - it certainly would be more something that gets passed in, rather than an internal state, so i think we just have no selection here and just the callback

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -0,0 +1,94 @@
...
+import net.skyscanner.backpack.compose.modal.rememberBpkModalState
+import net.skyscanner.backpack.compose.navigationbar.NavIcon
+
+sealed class AppSearchModalResult {
+    data class Content(
+        val sections: List<Section>,
+        val shortcuts: List<Shortcut>? = null,
+    ) : AppSearchModalResult()
+
+    data class Loading(val accessibilityLabel: String) : AppSearchModalResult()
```
marianeum, line 36: i believe this is unused - this needs to be announced as a state description when we set the loading state

marianeum (commented):
looks mostly good - a few small comments

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
LokmaneKrizou: nice one ! I just posted this question in bpk bellagio channel

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkShortcuts.kt
LokmaneKrizou: so probably we don't need to use `BpkSingleSelectChipGroup`, instead just display a lazy row of chips

marianeum: probably still good to use the chip group to get the spacing etc right as well, but with just a static selected index -1

app/src/androidTest/java/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModalTest.kt
```diff
@@ -0,0 +1,72 @@
...
+@RunWith(Parameterized::class)
+class BpkAppSearchModalTest(private val permutation: Permutation) : BpkSnapshotTest(listOf(permutation)) {
+
+    @Test
+    fun default() = record {
+        permutation.composable()
+    }
+
+    @Test
+    fun on_dark() = record(background = { BpkTheme.colors.surfaceContrast }) {
```
marianeum, line 42: not sure what this is for? there's no separate on dark mode of the modal, so this shouldn't be needed. i think just keeping it simple without a permutation and having a `content` `error` and `loading` test is enough

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
LokmaneKrizou: after discussion in [slack ](https://skyscanner.slack.com/archives/C05D9KNENRX/p1696925303853649) we will keep it as it is.

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
LokmaneKrizou: done

app/src/androidTest/java/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModalTest.kt
LokmaneKrizou: I added on dark just to make sure we check screenshot tests for the case when modal is dark, and be 100% it looks as intended 😅

marianeum: i don't think it's needed tbh - we already have the light/dark screenshots that ensure the background is updated/correct

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
```diff
@@ -0,0 +1,232 @@
...
+        BpkButton(
+            text = stringResource(R.string.generic_show),
+            onClick = { showModal.value = true },
+        )
+    }
+    if (showModal.value) {
+        BpkAppSearchModal(
+            modifier = modifier,
+            title = stringResource(id = R.string.destination),
+            inputText = "",
```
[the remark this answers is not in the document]
LokmaneKrizou: I didn't want to because we already using `BpkTextField` which should have its own tests already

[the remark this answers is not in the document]
LokmaneKrizou: you don't trust me 😞, just kidding, its a good point, that could happen, I will add test to make sure it will be highlighted in case someone changes it accidentally

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
```diff
@@ -0,0 +1,232 @@
...
+
+internal fun contentResult(resources: Resources) = BpkAppSearchModalResult.Content(
+    sections = listOf(
+        BpkSection(
+            items = listOf(
+                BpkItem(
+                    title = buildAnnotatedString {
+                        append(resources.getString(R.string.current_location_title))
+                    },
+                    subTitle = resources.getString(R.string.current_location_subtitle),
```
marianeum, line 123: massive nit: `subtitle`

marianeum (approved):
:shipit:

PR #1780, opened 2023-11-01
LMN-705 Fix range slider jumping

marianeum:
The track width is being dynamically calculated by the thumb width. Therefore we cannot have a dynamic thumb width. Change the label to not take any spacing into account when measuring to avoid this issue.
Also fixed a few spacing issues and made everything work correctly with dynamic font sizes as well.

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/internal/BpkRangeSliderImpl.kt
```diff
@@ -96,6 +96,15 @@ internal fun BpkRangeSliderImpl(
                 enabled = enabled,
             )
         },
+        track = {
+            SliderDefaults.Track(
+                colors = sliderColors(),
+                sliderPositions = it,
+                modifier = Modifier.applyIf(upperThumbLabel != null || lowerThumbLabel != null) {
+                    padding(top = LabelOffset)
```
marianeum, line 104: offset the track by the label if available. before there was extra spacing below the track as well in the label which is incorrect. the track is always centred otherwise.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/internal/BpkRangeSliderImpl.kt
```diff
@@ -108,44 +117,72 @@ private fun SlideRangeLabel(
...
-            textAlign = TextAlign.Center,
+    LabelLayout(modifier) {
+        Box(
             modifier = Modifier
                 .background(
                     color = BpkTheme.colors.coreAccent,
                     shape = FlareShape(
-                        borderRadius = BpkBorderRadius.Sm,
-                        flareHeight = BpkSpacing.Sm,
+                        borderRadius = BorderRadius,
```
marianeum, line 126: these were slightly wrong

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/internal/BpkRangeSliderImpl.kt
```diff
@@ -108,44 +117,72 @@ private fun SlideRangeLabel(
...
-        )
-        Spacer(
-            modifier = Modifier.height(BpkSpacing.Sm),
-        )
+                .padding(bottom = FlareHeight)
+                .height(BpkSpacing.Lg)
+                .padding(horizontal = BpkSpacing.Md),
+            contentAlignment = Alignment.Center,
+        ) {
+            BpkText(
```
marianeum, line 136: put this in a box to ensure we can centre align the text vertically and ensure we always centre it no matter the font scale

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/slider/internal/BpkRangeSliderImpl.kt
```diff
@@ -108,44 +117,69 @@ private fun SlideRangeLabel(
...
+                .semantics { stateDescription = label },
             interactionSource = interactionSource,
             colors = sliderColors(),
             enabled = enabled,
         )
     }
 }
 
+@Composable
+private fun LabelLayout(
```
marianeum, line 157: had to create a custom layout here to ensure that the label itself wasn't taking up any space, otherwise the range slider track will jump around constantly as it tries to recalculate the width as the label width changes

jonnycaley (approved):
Nice change! 👏

PR #1790, opened 2023-11-13
[NOJIRA] Remove clickable instead of not enable

jamipouchi:
# Description
When we first introduced the `Inset Banner`, we wanted to be able to disable the click in case there was no body to be shown. To do so, we introduced a change on the `clickable` modifier, with a boolean `enabled` that could be set to false.

Upon use of the `Inset Banner`, we realized that the click on the banner, if it was disabled would not call the onClick, but it would capture the click event, not letting it bubble up to the parent. That was not our desired behaviour, as we wanted the click to be handled by the parent card.

This pr reverts the modification to the `clickable` modifier, removing the enabled modifier, and changing the way we apply the onlick, instead of adding the boolean, using the `applyIf` modifier extension.


gabriellferreira (approved):
LGTNeutron

bvitaliyg:
Can you please bring some code to reproduce the original issue?

jamipouchi:
> @jamipouchi can you please bring some code to reproduce the original issue?

Basically, using the banner inside any composable which the `clickable` modifier, makes it so that when you click the banner it consumes the click event, not bubbling it to the parent composable. (even when the clickable is disabled).
What we expected was that when disabled, it acted as if it didn't have the click modifer, which is why I am introducing this fix.
Ex.
```
BpkCard(
        padding = BpkCardPadding.None,
        modifier = modifier.background(color = BpkTheme.colors.surfaceDefault)
            .semantics(mergeDescendants = true) {
                contentDescription = unpricedPackage.mainAccessibilityText
            },
        onClick = { onClick() }
    ) {
BpkInsetBanner(
                variant = unpricedPackage.headerTextVariant.toBpkVariant(),
                backgroundColor = Color(unpricedPackage.brandColor.toColorInt()),
                modifier = Modifier.padding(BpkSpacing.Md),
                title = unpricedPackage.headerTitle,
                subHeadline = unpricedPackage.headerSubheadline,
            ) {
                <image composable...>
            }
    <more composables....>
}
```

Video for reference:

https://github.com/Skyscanner/backpack-android/assets/84200829/75586b08-23a2-489f-a929-8bc2a5aed543

bvitaliyg (approved):

PR #1801, opened 2023-11-22
LMN-773: Add clear action for app search modal

karljiang1:
https://skyscanner.atlassian.net/browse/LMN-773
We need to support a clear action for textfield in app search modal so that user can clear it easily.

+ [x] Component `README.md`
+ [x] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
```diff
@@ -36,6 +37,7 @@ internal fun BpkAppSearchModalImpl(
     results: BpkAppSearchModalResult,
     onInputChanged: (String) -> Unit,
     modifier: Modifier = Modifier,
+    clearAction: BpkClearAction? = null,
```
LokmaneKrizou, line 40: clearAction should be required always for `BpkAppSearchModal`

karljiang1: Have updated it.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/BpkTextField.kt
```diff
@@ -45,6 +50,7 @@ fun BpkTextField(
     keyboardActions: KeyboardActions = KeyboardActions(),
     maxLines: Int = 1,
     interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    clearAction: BpkClearAction? = null,
```
bvitaliyg, line 53: Do we have this parameter in Figma? Does it uses the same name and type parameter?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
```diff
@@ -243,6 +252,14 @@ private fun TextFieldBox(
                     },
                 ).value,
             )
+        } else if (clearAction != null) {
+            BpkIcon(
```
bvitaliyg, line 256: I don't see the similar to the start icon padding here – is that intentional?

bvitaliyg (commented):
Have few questions, mostly about the design

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
karljiang1: ooh, I will add an end padding here for symmetry.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
```diff
@@ -35,6 +36,7 @@ internal fun BpkAppSearchModalImpl(
     inputHint: String,
     results: BpkAppSearchModalResult,
     onInputChanged: (String) -> Unit,
+    clearAction: BpkClearAction,
```
marianeum, line 39: rather than adding a new action here can we add a `Clear` type in the `BpkFieldStatus` with the same properties? that way we're not running into that situation where you can have clear AND validation/error, which is not possible and matches figma states as well

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -95,6 +97,7 @@ fun BpkAppSearchModal(
             inputHint = inputHint,
             results = results,
             onInputChanged = onInputChanged,
+            clearAction = clearAction,
```
marianeum, line 100: we should also make sure that we're only showing the clear action if the text is not empty here

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
karljiang1: Should we still add a callback for the clear event when we remove the clearAction?

marianeum: yes sorry that's what i meant with the same params - we still need the content description + callback

marianeum: sorry this wasn't clear - what i meant is that the `Clear` status should contain the properties here, rather than introducing a new `clearAction` property - otherwise we'll have conflicting sources for the clear action and can run into an invalid state (where status is clear, but the action is not provided)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
```diff
@@ -242,6 +252,21 @@ private fun TextFieldBox(
                         else -> BpkTheme.colors.textPrimary
                     },
                 ).value,
+                modifier = Modifier
+                    .padding(end = BpkSpacing.Sm),
```
marianeum, line 256: i think this adds unwanted spacing - as you can see the select snapshots were updated and have extra padding now

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
marianeum: i don't think the padding is needed here as the text field already has padding as a whole which matches figma

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
karljiang1: will remove it.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkAppSearchModalImpl.kt
marianeum: really sorry for the back and forth, BUT thinking this through again and looking at this in place it does make sense to have a separate clear action, rather than having a `BpkFieldStatus` of clear. few reasons for this:
the clear action is _only_ valid for the text field, not any other field set types
the logic to show/hide the clear action depending on content should be within text field to simplify the API (as you wrote it)
this also matches iOS API

one thing to change from the previous implementation, though: if we DO have a valid/error status this should take precedence over the clear action being available or not (ie if the status is valid/error we show that, rather than the clear action)

karljiang1: Yes, it would make more sense.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
```diff
@@ -244,23 +253,45 @@ private fun TextFieldBox(
...
+                        }
+                        .testTag("textFieldClearButton"),
+                )
+
+                else -> lastIcon = null // reset it to null
             }
+
             AnimatedVisibility(
-                visible = status is BpkFieldStatus.Validated || status is BpkFieldStatus.Error,
+                visible = status is BpkFieldStatus.Validated || status is BpkFieldStatus.Error || status is BpkFieldStatus.Default,
```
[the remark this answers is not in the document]
karljiang1: Will update it.

marianeum (approved):
🚀

PR #1809, opened 2023-11-28
LMN-790 Fix modal show/hide duration

marianeum:
The show/hide function was using a fixed delay. This wasn't working correctly when the animation speed is set to a different value/disabled as the animation would finish at a different point as the delay. This resulted in unexpected behaviour when animation speed was different to the default

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModalState.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModalState.kt
@@ -21,7 +21,8 @@ package net.skyscanner.backpack.compose.modal
 import androidx.compose.animation.core.MutableTransitionState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
-import kotlinx.coroutines.delay
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
 @Composable
 fun rememberBpkModalState() = remember {
@@ -32,12 +33,22 @@ class BpkModalState internal constructor(
     internal val isVisible: MutableTransitionState<Boolean>,
 ) {
     suspend fun show() {
-        isVisible.targetState = true
-        delay(ModalAnimationDurationMs.toLong())
+        animateState(true)
     }
 
     suspend fun hide() {
-        isVisible.targetState = false
-        delay(ModalAnimationDurationMs.toLong())
+        animateState(false)
+    }
+
+    private suspend fun animateState(newVisibility: Boolean) {
+        isVisible.targetState = newVisibility
+        if (isVisible.currentState == newVisibility) {
+            return
+        }
+        withContext(Dispatchers.IO) {
+            while (!isVisible.isIdle) {
+                // continue until the animation is finished
+            }
+        }
     }
 }


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModalState.kt
```diff
@@ -32,12 +33,22 @@ class BpkModalState internal constructor(
...
+        animateState(false)
+    }
+
+    private suspend fun animateState(newVisibility: Boolean) {
+        isVisible.targetState = newVisibility
+        if (isVisible.currentState == newVisibility) {
+            return
+        }
+        withContext(Dispatchers.IO) {
+            while (!isVisible.isIdle) {
```
marianeum, line 49: we now just wait until the animation is no longer idle - this means that no matter the animation speed we'll run this function as long as the animation runs

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModalState.kt
```diff
@@ -32,12 +33,22 @@ class BpkModalState internal constructor(
     internal val isVisible: MutableTransitionState<Boolean>,
 ) {
     suspend fun show() {
-        isVisible.targetState = true
-        delay(ModalAnimationDurationMs.toLong())
```
marianeum, line 36: this was causing issues when a different animation speed was set on the device

alejandrorosas (approved):

bvitaliyg (approved):

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModalState.kt
chraac: IIRC, it's kind of spin lock here, and will busy looping till the animation end (in most device will took a cpu core  ~300ms), this look a bit expensive to just for the animation?

How about put a `delay` inside the `while`:
```
while (!isVisible.isIdle) {
    delay(33L)  // 33ms is a common frame interval for most devices
}
```

chraac: just looping: 
image

looping with delay 8ms:
image

from the profiler, we can see that the peak cpu usage is dropped after add a `delay(8L)`

chraac: yeah, gonna create a PR about this

marianeum: hmm yes you might be right - i think i'd like to spend some more time on this later on to find a good solution for this tbh - with the delay we risk there being issues as well as currently the idle check will always happen at the same time (or rather there won't be an ui update in between) - this means we can rely on this callback to be called when the modal/viewmodel scope is still alive. as soon as we add a delay we risk a race condition where sometimes depending on the delay the modal is already dismissed/the viewmodel gone and our bug is back

chraac: > hmm yes you might be right - i think i'd like to spend some more time on this later on to find a good solution for this tbh - with the delay we risk there being issues as well as currently the idle check will always happen at the same time (or rather there won't be an ui update in between) - this means we can rely on this callback to be called when the modal/viewmodel scope is still alive. as soon as we add a delay we risk a race condition where sometimes depending on the delay the modal is already dismissed/the viewmodel gone and our bug is back

emmmm, i think you're right, if we add a `delay` here, then this couroutine will be switch out of thread, at that time, if a recompose was trigged, the couroutine job will be close.
maybe we can keep the current source code (without `delay`), so that we can fix the bug, then create another PR to optimize this spin lock.

PR #1820, opened 2023-12-08
Switch githooks to standard git hook path

marianeum:
Changing the githooks path sometimes broke the path for the app, meaning that the githooks for the main app weren't run anymore. This changes the githooks to just copy the file and use the standard folder, rather than changing the hookspath.

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,14 +86,10 @@ tasks.dokkaHtmlMultiModule.configure {
     }
 }
 
-task installGitHooks() {
-    def command = Runtime.getRuntime().exec("git config core.hooksPath")
-    def currentGitHookDir = command.inputStream.text.trim()
-    def hooksPath = "hooks"
-    if (currentGitHookDir != hooksPath) {
-        Runtime.getRuntime().exec("git config --local core.hooksPath ${hooksPath}")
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -86,14 +86,10 @@ tasks.dokkaHtmlMultiModule.configure {
...
 
-task installGitHooks() {
-    def command = Runtime.getRuntime().exec("git config core.hooksPath")
-    def currentGitHookDir = command.inputStream.text.trim()
-    def hooksPath = "hooks"
-    if (currentGitHookDir != hooksPath) {
-        Runtime.getRuntime().exec("git config --local core.hooksPath ${hooksPath}")
-        Runtime.getRuntime().exec("chmod -R +x ${hooksPath}")
-    }
+task installGitHooks(type: Copy) {
```
marianeum, line 89: the old version caused issues with the main app as it seems to have changed the hookspath globally, rather than just for backpack (despite using `--local`). copying the file seems to be the standard approach used for installing githooks on apps, so switching to that. 
while this doesn't "uninstall" the custom hookpath this should still work if a custom path is set as the directory doesn't change, but it reduces the risk of a custom hookspath being changed accidentally.

alejandrorosas (approved):

PR #1840, opened 2024-01-04
Switch screenshot tests to Roborazzi

marianeum:
I put together a test to switch our snapshot tests to roborazzi. This will allow us to stop relying on emulators for snapshot tests and should massively speed up and stabilise our snapshot tests. 

There's a couple remaining issues:
- Shadows are not working right now (this is in progress in robolectric and should be available soon after which roborazzi can be updated)
- View Map markers are not tinted. This used to be an issue in the past already and resurfaced now - we can work around that by applying the tinting programmatically if needed
- The view bar chart select line is not rendered in the correct position. The component is fully unused and we don't expect any new components for views anymore
- The toast test isn't working (it can't access the view) - the component is deprecated and is just a wrapper around the old one anyways

Aside from shadows these are all things we can live with IMO and shadows are coming, so we need to make a decision if it's worth while adopting this now or wait.

This change brings down the time for running ALL snapshots to ~7 minutes on CI, rather than up to 40 minutes for one variant.

+ [ ] Component `README.md`
+ [ ] Tests


.github/workflows/_build.yml
```diff
@@ -127,11 +127,11 @@ jobs:
 
   Screenshots:
     name: Screenshots tests
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
```
marianeum, line 130: was hoping we can keep the same OS and keep snapshot test changes to a minimum, but unfortunately the gradle task wasn't running on macos - it didn't use the right jvm (i don't think that has anything to do with roborazzi, though). this matches the other non-emulator tasks

.github/workflows/_build.yml
```diff
@@ -127,11 +127,11 @@ jobs:
...
     permissions:
       pull-requests: write
       contents: write
-    timeout-minutes: 40
+    timeout-minutes: 20
```
marianeum, line 134: 20 minutes is still too high :D

.github/workflows/_build.yml
```diff
@@ -154,77 +154,21 @@ jobs:
           distribution: 'adopt'
           cache: gradle
 
-      - name: AVD cache
```
marianeum, line 157: byeeee

.github/workflows/_build.yml
```diff
@@ -132,16 +132,6 @@ jobs:
       pull-requests: write
       contents: write
     timeout-minutes: 20
-
-    strategy:
```
marianeum, line 136: now that the tests run so much faster it's much more likely that there's a clash between two different variants trying to commit the changes, resulting in a CI failure. so run all variants in one go. the whole thing still takes under 10 minutes anyways

app/src/test/java/net/skyscanner/backpack/meta/StoriesTests.kt
```diff
@@ -28,21 +28,21 @@ class StoriesTests {
 
     @Test
     fun assertNoDemoOnlyStoriesInScreenshots() {
-        assertTrue(repository.screenshotStories.none { it.kind == StoryKind.DemoOnly })
+        Assert.assertTrue(repository.screenshotStories.none { it.kind == StoryKind.DemoOnly })
```
bvitaliyg, line 31: Can we add the import here so less diff is produced?

app/src/test/java/net/skyscanner/backpack/meta/StoriesTests.kt
```diff
@@ -28,21 +28,21 @@ class StoriesTests {
...
-        assertTrue(repository.uiComponents.none {
+        Assert.assertTrue(repository.uiComponents.none {
             repository.storiesOf(it.name, compose = true).any { it.kind == StoryKind.ScreenshotOnly }
         })
-        assertTrue(repository.uiComponents.none {
+        Assert.assertTrue(repository.uiComponents.none {
             repository.storiesOf(it.name, compose = false).any { it.kind == StoryKind.ScreenshotOnly }
         })
-        assertTrue(repository.tokenComponents.none {
+        Assert.assertTrue(repository.tokenComponents.none {
```
bvitaliyg, line 42: And here

app/src/test/java/net/skyscanner/backpack/toast/BpkToastTests.kt
```diff
@@ -18,18 +18,16 @@
...
+import org.junit.Ignore
 import org.junit.Test
-import org.junit.runner.RunWith
 
-@RunWith(AndroidJUnit4::class)
 @Suppress("DEPRECATION")
 class BpkToastTests : BpkSnapshotTest() {
 
-    @Test
+    @Test @Ignore("This currently doesn't seem to work")
```
bvitaliyg, line 28: So we commented all the tests that doesn't work?

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ coroutines = "1.7.3"
 espresso = "3.5.1"
 kotlinPoet = "1.15.3"
 ktor = "2.3.7"
+roborazzi = "1.9.0-alpha-4"
```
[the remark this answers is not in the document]
marianeum: we need it for any test that captures multiple windows, like dialogs, bottom sheets etc otherwise the snapshot is empty

app/src/test/java/net/skyscanner/backpack/toast/BpkToastTests.kt
marianeum: this is the only one as it crashes

bvitaliyg (approved):

PR #1848, opened 2024-01-16
Ralp-4728 Update visual appearance of Badges in Compose

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
```diff
@@ -102,11 +117,11 @@ private val BpkBadgeType.contentColor: Color
 private val BpkBadgeType.backgroundColor: Color
     @Composable
     get() = when (this) {
-        BpkBadgeType.Normal -> BpkTheme.colors.surfaceHighlight
```
alejandrorosas, line 105: Looking at the new screenshots seems that we also need to update the `contentColor` to `textOnDark` for the Success, Warning and Destructive. See: 
image

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
```diff
@@ -60,26 +63,38 @@ fun BpkBadge(
...
             .semantics(mergeDescendants = true) { }
-            .requiredHeight(BpkSpacing.Lg)
             .border(1.dp, type.borderColor, BadgeShape)
             .background(type.backgroundColor, BadgeShape)
-            .padding(horizontal = BpkSpacing.Md),
+            .padding(horizontal = BpkSpacing.Md, vertical = BpkSpacing.Sm),
         horizontalArrangement = Arrangement.spacedBy(BpkSpacing.Sm),
         verticalAlignment = Alignment.CenterVertically,
     ) {
         val contentColor = type.contentColor
```
marianeum, line 72: I think we need to update the text colour as well looking at the dark mode screenshots the text is quite hard to read for some

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
```diff
@@ -60,26 +63,38 @@ fun BpkBadge(
...
+            else -> icon
+        }
+        if (badgeIcon != null) {
             BpkIcon(
-                icon = icon,
+                icon = badgeIcon,
                 contentDescription = null,
                 size = BpkIconSize.Small,
-                tint = contentColor,
+                tint = when (type) {
```
marianeum, line 84: nit: to make it a little more readable it might be better to create a variable for this (or even create an extension function, like `iconColor` in the same place the `contentColor` is defined to use this

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
```diff
@@ -60,26 +63,38 @@ fun BpkBadge(
...
             .border(1.dp, type.borderColor, BadgeShape)
             .background(type.backgroundColor, BadgeShape)
-            .padding(horizontal = BpkSpacing.Md),
+            .padding(horizontal = BpkSpacing.Md, vertical = BpkSpacing.Sm),
         horizontalArrangement = Arrangement.spacedBy(BpkSpacing.Sm),
         verticalAlignment = Alignment.CenterVertically,
     ) {
         val contentColor = type.contentColor
-        if (icon != null) {
+        val badgeIcon = when (type) {
```
marianeum, line 73: this will always set the icon to the pre-defined one for the warning, destructive, success types even if the consumer provided a different icon. we should check if the icon is null and only then use the default one instead

marianeum (commented):
great change! mainly some minor comments + what @alejandrorosas also reported

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
jaanainkinen: Oh I think I misunderstood then, I thought the badges were fixed for these types, i.e. always the same icon for Success, Warning and Destructive. I asked Adam if the consumer should be able to change the icon colours and he said no, but maybe I asked the wrong question and the icon should still be optional?

marianeum: they should not be able to change the icon colours, but i believe the icon itself. so we _always_ show an icon for these types, but we only use the default if no other icon is provided

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
jaanainkinen: Thanks, I've fixed this now!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
jaanainkinen: Ahh I see! Thanks for clarifying this, I'll fix it now.

jaanainkinen: I think this should be fixed now!

app/src/main/java/net/skyscanner/backpack/demo/compose/BadgeStory.kt
```diff
@@ -69,21 +68,25 @@ private fun BadgeRow(
...
+            modifier = Modifier
+                .weight(1f)
+                .wrapContentWidth(align = Alignment.CenterHorizontally),
             type = type,
-            icon = when (type) {
-                BpkBadgeType.Warning -> BpkIcon.HelpCircle
-                BpkBadgeType.Destructive -> BpkIcon.CloseCircle
-                else -> BpkIcon.TickCircle
-            },
+            icon = BpkIcon.TickCircle,
```
marianeum, line 89: To match what Figma does here I would use a null icon for warning, destructive and success, otherwise the tick - otherwise all icons will be ticks now on the sample/docs
Also can you re-generate the docs screenshots please? btw as badge is used in some other components these will have changed too (as you can see by the snapshot tests) - can you keep the changes for these docs screenshots as well?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
jaanainkinen: This should now be ok!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/badge/BpkBadge.kt
```diff
@@ -60,40 +63,56 @@ fun BpkBadge(
     Row(
```
marianeum, line 63: Commenting here because I can't comment on the `BpkBadge` composable function itself.
This doesn't have anything to do with your PR, but while we're at it can you set a default value for `type` (should be `Normal`). then we can also update the `default` test to not set a type - it confused me a lot that the default snapshot had a destructive case.

jaanainkinen: Thanks for pointing this out, it should now be fixed!

marianeum (approved):
great job 👏

PR #1871, opened 2024-02-07
[NOJIRA] Fix Default badge type to be Normal in View

jaanainkinen:

+ [ ] Component `README.md`
+ [ ] Tests


Backpack/src/main/java/net/skyscanner/backpack/badge/BpkBadge.kt
```diff
@@ -121,7 +121,7 @@ open class BpkBadge @JvmOverloads constructor(
      * @property type
      * Type of badge. Default Type.Success
      */
-    var type: Type = Type.Success
+    var type: Type = Type.Normal
```
marianeum, line 124: i think you also need to update this line:
`type = Type.fromId(a.getInt(R.styleable.BpkBadge_badgeType, 1))`
the 1 defines the default type as read in xml and matches the id defined in the enum

jaanainkinen: Thanks Maria! I think this should be fixed now.

marianeum (approved):
we fixed so many bugs here 🚀

PR #1880, opened 2024-02-16
RALP-4761 use index from HorizontalPager to select current image 

LewisI224:
This PR introduces a fix for the BpkImageGallerySlideshow component. Previously when swiping to the next image the current image was shown twice while swiping rather than showing the current image and the next image stitched together.

This fix involves using the index provided by the horizontal pager to select the currently used image rather than tracking the index independently.

| Before  | After |
| ------------- | ------------- |
| [before.webm](https://github.com/Skyscanner/backpack-android/assets/19375170/36236361-c298-4301-9c82-f5437577bdab) | [after.webm](https://github.com/Skyscanner/backpack-android/assets/19375170/27a065e3-b6da-4053-9c34-cf8f1e278322) |


marianeum (approved):
very clean fix - congrats on your first backpack contribution 🎉

PR #1897, opened 2024-03-11
Add navigation bar in bottom sheet modal

LokmaneKrizou:

+ [x] Component `README.md`
+ [x] Tests


app/src/main/java/net/skyscanner/backpack/demo/compose/BottomSheetStory.kt
```diff
@@ -106,8 +107,30 @@ fun ImageBottomSheetStory(
...
+        title = stringResource(id = R.string.modal_bottom_sheet_title),
+        isClosable = BpkModalBottomSheetCloseAction.Default(stringResource(id = R.string.navigation_close)),
+    )
+}
+
+@Composable
+internal fun ModalBottomSheetStory(
+    modifier: Modifier = Modifier,
+    title: String? = null,
+    isClosable: BpkModalBottomSheetCloseAction = BpkModalBottomSheetCloseAction.None,
```
marianeum, line 133: i think the api table means that `isClosable` should be a boolean. but yes we can enforce the "needs to be set if `isClosable=true` by making this an enum. i would name this `closeButton`, though to align with the API table and be less confusing than a boolean name which really is an enum

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
```diff
@@ -0,0 +1,198 @@
...
+            is BpkDragHandleStyle.OnImage -> {
+                Box(modifier = modifier) {
+                    Column { content() }
+                    BpkModalBottomSheetHeader(
+                        title = title,
+                        state = state,
+                        dragHandleStyle = dragHandleStyle,
+                        isClosable = isClosable,
+                        onDismissRequest = onDismissRequest,
+                    )
```
LokmaneKrizou, line 128: I had to wrap this with box so we make sure the header will be on top of the content

LokmaneKrizou: I just noticed we have similar thing `BottomSheetContent`i will use that instead 😅

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
```diff
@@ -0,0 +1,198 @@
...
+        containerColor = BpkTheme.colors.surfaceElevated,
+        contentColor = BpkTheme.colors.textPrimary,
+        scrimColor = BpkTheme.colors.scrim,
+        tonalElevation = BpkElevation.Lg,
+        onDismissRequest = onDismissRequest,
+    )
+}
+
+@Composable
+fun ColumnScope.BpkModalBottomSheetContent(
```
marianeum: should this be `internal`?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/BpkModalBottomSheet.kt
```diff
@@ -19,37 +19,33 @@
...
-import net.skyscanner.backpack.compose.bottomsheet.internal.BpkBottomSheetHandle
 import net.skyscanner.backpack.compose.bottomsheet.internal.BpkDragHandleStyle
-import net.skyscanner.backpack.compose.theme.BpkTheme
-import net.skyscanner.backpack.compose.tokens.BpkBorderRadius
-import net.skyscanner.backpack.compose.tokens.BpkElevation
+import net.skyscanner.backpack.compose.bottomsheet.internal.BpkModalBottomSheetImpl
+
+sealed class BpkModalBottomSheetCloseAction {
+    data object None : BpkModalBottomSheetCloseAction()
+    data class Default(val contentDescription: String) : BpkModalBottomSheetCloseAction()
```
marianeum, line 29: given that it's not actually default - `Close`?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
```diff
@@ -0,0 +1,198 @@
...
+    state: BpkModalBottomSheetState,
+    onDismissRequest: () -> Unit,
+    content: @Composable ColumnScope.() -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    if (isClosable is BpkModalBottomSheetCloseAction.Default || !title.isNullOrEmpty()) {
+        when (dragHandleStyle) {
+            BpkDragHandleStyle.Default -> {
+                Column(modifier = modifier,
+                    verticalArrangement = Arrangement.spacedBy(BpkSpacing.Lg),
```
marianeum, line 105: as we're exposing the column scope to the content this will mean every other item within the content will have a larger spacing - this is probably a case where we want a spacer. also not sure the spacing is right? i don't think there's any spacing between the header + content enforced

marianeum (commented):
mainly some nits on the API & layout - also can you please generate the docs screenshots?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
LokmaneKrizou: you checked this few minutes before I pushed the latest changes

marianeum (commented):
code looks good - just needs the docs snapshots generated (this needs to happen locally still)

PR #1903, opened 2024-03-18
Fix Chip background not clipped when pressed

andra-georgescu:
| Background | Before | After |
|:------------:|:-------:|:-----:|
| default |   |   |
| contrast |       |    |
| on image |  |     |

**Dismissible chip**
| Background | Before | After |
|:------------:|:-------:|:-----:|
| default |   |   |
| contrast |       |    |
| on image |  |     |


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chip/internal/BpkChipImpl.kt
```diff
@@ -80,12 +81,13 @@ internal fun BpkChipImpl(
         type = type,
         interactionSource = interactionSource,
         modifier = modifier.applyIf(onSelectedChange != null) {
-            selectable(
-                selected = selected,
-                enabled = enabled,
-                interactionSource = interactionSource,
-                indication = LocalIndication.current,
-            ) { onSelectedChange!!.invoke(!selected) }
+            clip(ChipShape)
```
marianeum, line 84: i think we might need to apply the same for the dismissable chip

marianeum (approved):
thanks for the fix!

PR #1913, opened 2024-03-25
Add support for Behavioural callbacks

michaeltweed:
See internal decision doc: https://skyscanner.atlassian.net/l/cp/xZy2ipBP.

As the callback is nullable and set to `null` as default, there will be no changes required for consumers.

+ [X] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-android/blob/main/CONTRIBUTING.md)

+ [ ] Component `README.md`
+ [ ] Tests


app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
```diff
@@ -117,6 +119,15 @@ internal fun DefaultAppSearchModalSample(
             onClose = { showModal.value = false },
             onInputChanged = { destination.value = it },
             clearAction = BpkClearAction(stringResource(id = R.string.text_field_clear_action_description)) { destination.value = "" },
+            behaviouralCallback = object : BehaviouralCallback {
+                override fun onDrawn(element: Any) {
+                    Log.i("AppSearchModalStory", "onDrawn $element")
```
michaeltweed, line 124: We can delete these log lines if needed- I wasn't sure whether to keep them in for the example or not

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/BehaviouralCallback.kt
```diff
@@ -0,0 +1,6 @@
+package net.skyscanner.backpack.compose.utils
+
+interface BehaviouralCallback {
```
michaeltweed, line 3: Should this be combined into one file with the `BehaviouralModifiers` (which is just one extension function)? That way we could also add full inline docs to the class

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
marianeum: yeah no harm doing so. i think having some tests for this would be important, though - for the most part we just use snapshot tests in backpack, but these call for some ui tests (there's already a few for some components)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/BehaviouralCallback.kt
marianeum: yeah same file makes sense 👍

marianeum (commented):
implementation looks good. we should also update the readme for the component to mention the callback

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
michaeltweed: Good point- I'll just wait for a couple more reviews/iOS alignment before writing tests and docs, just in case the approach changes

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
```diff
@@ -117,6 +119,15 @@ internal fun DefaultAppSearchModalSample(
             onClose = { showModal.value = false },
             onInputChanged = { destination.value = it },
             clearAction = BpkClearAction(stringResource(id = R.string.text_field_clear_action_description)) { destination.value = "" },
+            behaviouralCallback = object : BehaviouralCallback {
```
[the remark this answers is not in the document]
marianeum: `drawAndClickCallback`? 😬  i guess that'll restrict us to only the two events.
i guess one other question here: with this implementation the click callbacks are duplicated - once here and once as part of the data - is the intention that these callbacks will _only_ be used for behavioural logging? or should we actually update the other callbacks to have both draw + click?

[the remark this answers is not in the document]
marianeum: Makes sense. I can't think of a better name as behavioural callback fwiw, so maybe let's stick to that.

Yeah I agree it's nice not having to set them and it also means it isn't a breaking API change. We just need to be clear in the documentation that these should ONLY be used for behavioural events and nothing else. Tbf this probably even makes it a little easier to integrate in the app side as we can register all relevant events at once.

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
michaeltweed: @marianeum I'm getting back to this now 😀 which class(es) would you recommend adding the tests for? For the high-level `BpkAppSearchModal`, which just passes the callback through, or the `BpkSectionItem` which actually contains the `registerForBehaviouralEvents` call?

marianeum: basically testing that the callback gets called at the right place with the right item

michaeltweed: @marianeum I've added one test now, does this match what you'd expect? If so I can cover the other test cases and add documentation

backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModalTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Test
+
+class BpkAppSearchModalTest {
+
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    @Test
+    fun withBehaviouralCallbackAndClickActionAdded() = runTest {
+        var didReceiveBehaviouralOnDrawnCallback = false
```
michaeltweed, line 44: I could update these to mocks if needed, wasn't sure if there was a preferred Backpack testing approach but thought I'd keep it simple for the first draft of the test

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
marianeum: yep that makes sense, thank you!

backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModalTest.kt
marianeum: yeah always happy not to use mocks if we don't need to :)

marianeum (approved):

PR #1923, opened 2024-04-09
Try out new behavioural event API

marianeum:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
```diff
@@ -79,7 +79,7 @@ fun BpkAppSearchModal(
     clearAction: BpkClearAction,
     modifier: Modifier = Modifier,
     state: BpkModalState = rememberBpkModalState(),
-    behaviouralCallback: BehaviouralCallback? = null,
+    behaviouralEventWrapper: (@Composable (BpkItem, content: @Composable BpkClickHandleScope.() -> Unit) -> Unit)? = null,
```
marianeum, line 82: i don't know if there's a way to have this as a reusable declaration, rather than having to define this api again and again. but like this it has the benefit that we can actually provide the right type as well, rather than just any

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/internal/BpkSection.kt
```diff
@@ -65,11 +65,14 @@ internal fun BpkSectionHeading(
 }
 
 @Composable
-internal fun BpkSectionItem(item: BpkItem, modifier: Modifier = Modifier, behaviouralCallback: BehaviouralCallback? = null) {
+internal fun BpkSectionItem(item: BpkItem, modifier: Modifier = Modifier, clickHandleScope: BpkClickHandleScope? = null) {
     Row(
         modifier = modifier
             .fillMaxWidth()
-            .registerForBehaviouralEvents(item = item, behaviouralCallback = behaviouralCallback, onClick = item.onItemSelected),
+            .clickable {
```
marianeum, line 72: this also means we can reuse this for other types of events (ie making something selectable, rather than clickable to ensure the right accessibility announcements are made)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModal.kt
alejandrorosas: a `typealias`?

marianeum: yup :D i somehow doubted this was gonna work with composable functions, but it does

app/src/main/java/net/skyscanner/backpack/demo/compose/AppSearchModalStory.kt
```diff
@@ -119,15 +117,6 @@ internal fun DefaultAppSearchModalSample(
             onClose = { showModal.value = false },
             onInputChanged = { destination.value = it },
             clearAction = BpkClearAction(stringResource(id = R.string.text_field_clear_action_description)) { destination.value = "" },
-            behaviouralCallback = object : BehaviouralCallback {
```
marianeum, line 122: it's hard to provide a sample here as it requires the handling of the `BpkClickHandleScope` (as that needs to be passed to the content)

backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/appsearchmodal/BpkAppSearchModalTest.kt
marianeum, line 1: Now that backpack is just doing the "dumb" stuff and the app handling all the complexity around the view events this test isn't that necessary anymore (we're not testing other clicks for example either. maybe we should but that's a different topic)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/utils/BpkBehaviouralEventWrapper.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+/**
+ * This is used for registering components for behavioural events. Right now this supports `onClick` events, while any other
+ * events must be handled by the consumer.
+ */
+
+typealias BpkBehaviouralEventWrapper = @Composable (item: Any, modifier: Modifier, content: @Composable BpkClickHandleScope.() -> Unit) -> Unit
```
marianeum, line 29: also added a `modifier` here as i'm sure there will be examples where we'll need to handle the wrapper taking the modifier (for example in the case where the scope is important). The alternative would be having to wrap the wrapper within the component to apply the modifier

alejandrorosas (approved):

PR #1931, opened 2024-04-17
Update tokens for foundations updates

marianeum:
A small CI improvement to update the tokens when there's a foundations update, rather than failing CI and having to perform the update locally.

+ [ ] Component `README.md`
+ [ ] Tests


.github/workflows/_build.yml
```diff
@@ -35,7 +42,6 @@ jobs:
 
       - name: Danger
         uses: docker://ghcr.io/danger/danger-kotlin:1.2.0
-        if: ${{ github.event_name == 'pull_request' }}
```
marianeum, line 38: i'm not sure why this was only running on pull requests - doesn't feel correct

.github/workflows/_build.yml
```diff
@@ -54,10 +60,26 @@ jobs:
           ./gradlew assemble${{ env.flavour }}${{ env.config }} -PdisablePreDex
 
       - name: Tokens check
+        if: ${{ github.event_name != 'pull_request' || github.actor != 'dependabot[bot]' }} || ${{ github.event_name == 'pull_request' && github.actor == 'dependabot[bot]' }} && !contains(steps.dependabot-metadata.outputs.dependency-names, 'bpk-')
```
marianeum, line 63: we always want to run this, except if it is a backpack dependency update check

.github/workflows/_build.yml
```diff
@@ -54,10 +60,26 @@ jobs:
...
         run: |
           ./gradlew generateTokens -PdisablePreDex
           ./scripts/check-no-changes.sh
 
+      - name: Token update
+        if: ${{ github.event_name == 'pull_request' && github.actor == 'dependabot[bot]' }} && contains(steps.dependabot-metadata.outputs.dependency-names, 'bpk-')
```
marianeum, line 69: and this is the opposite - we only want to run this for the backpack dependency update. in that case we just commit the changes. i tried testing this in the currently open dependabot PR, but then the actor is me, not dependabot. will merge and then recreate the dependabot PR to test

xo0ps (approved):

PR #1972, opened 2024-05-14
AI-308 origin drop down text overlap

LokmaneKrizou:
Making a small adjustment for buttons to display long text along side with icon
|Before|After|
|-|-|
Screenshot 2024-05-14 at 16 30 24|Screenshot 2024-05-14 at 16 28 06

Please include a description of the changes you are introducing and some screenshots if appropriate.

Please ensure your pull request title is clear as it will be used to generate the changelog.

Add `major`, `minor` or `patch` label depending on the change according to [semver](semver.org) or `skip-changelog` if the change shouldn't be added to the changelog (e.g. a change to a test or documentation)
-->

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/BpkButton.kt
```diff
@@ -131,7 +130,7 @@ fun BpkButton(
                 ButtonText(text)
```
marianeum, line 130: does it also need to be applied here or is this one correct? if so maybe the modifier should be applied internally to `ButtonText`

LokmaneKrizou: hmm in that shouldn't be a problem as for this case the icon is always at the start and it will take its allocated space first then the text will come after
Screenshot 2024-05-14 at 16 34 09

marianeum: also can you add a snapshot test example that covers these cases to make sure it is covered by a test

LokmaneKrizou: I wasn't sure if we need one, but sure let me add that 😄

marianeum (approved):
nice one, thanks! just do an empty commit to re-run CI otherwise looks good 👍

PR #1986, opened 2024-05-23
[MUON-304] - CardList - Rail

gabriellferreira:
* This PR adds the CardList BPK component, only the RAIL variation - Stack will be handled in a following PR.

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
```diff
@@ -0,0 +1,57 @@
...
+        description = description,
+        layout = layout,
+        modifier = modifier,
+        dataList = dataList,
+        elements = elements,
+    )
+}
+
+sealed class BpkCardListLayout(
+    open val button: BpkCardListButtonAccessory?,
```
marianeum, line 46: looking at the api in figma this is not correct. there should be a button prop on the top level and in the stack case there's either a button or expand option.
as you're not implementing the stack option right now i would just leave this out and if we add the stack option later on the default will just be rail to have a non-breaking change

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
```diff
@@ -0,0 +1,57 @@
...
+import net.skyscanner.backpack.compose.cardlist.internal.BpkCardListImpl
+
+@Composable
+fun <T> BpkCardList(
+    title: String,
+    description: String,
+    layout: BpkCardListLayout,
+    dataList: List<T>,
+    modifier: Modifier = Modifier,
+    elements: @Composable (LazyItemScope.(Int) -> Unit),
```
marianeum, line 33: we generally follow the compose conventions of calling this `content`.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.cardlist.internal.BpkCardListImpl
+
+@Composable
+fun <T> BpkCardList(
+    title: String,
+    description: String,
+    layout: BpkCardListLayout,
+    dataList: List<T>,
```
marianeum, line 31: rather than providing this (the component doesn't need to know anything about the data) i'd follow what we do in `BpkCarousel` and use a state here. we can probably reuse the same state (we do that in a few other components)

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/internal/BpkCardListImpl.kt
```diff
@@ -0,0 +1,97 @@
...
+@Composable
+fun <T> RailLayout(
+    dataList: List<T>,
+    modifier: Modifier = Modifier,
+    elements: @Composable (LazyItemScope.(Int) -> Unit),
+) {
+    LazyRow(
+        modifier = modifier,
+        horizontalArrangement = Arrangement.spacedBy(BpkSpacing.Base),
+        contentPadding = PaddingValues(start = 16.dp, end = 16.dp),
```
marianeum, line 87: please use bpk spacing values here

marianeum (commented):
nice changes so far - just an early review around the api

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
gabriellferreira: I got your point but I think we will still have a breaking change for stack later.

from the API definition, we can not have a header button and an accessory button at the same time. If we keep the API definitions we could have both being instantiated but which one would have priority to be rendered?

marianeum: I think you'd show both in the stack case: the accessory button is for the end of the stack - not the button for the section header

gabriellferreira: Screenshot 2024-05-24 at 17 08 18
thats the issue 😄

marianeum: Ah I see. In that case we just don't show the accessory button even if you set the accessory type to button. I don't think we can really control this from an api perspective without being too complex as there's a bunch of combinations. For example you might ONLY want an accessory button and not a section header button - having the button in the layout property would then not handle that.
I'm also not 100% sure if the button accessory option is even a case on apps as I can't see any designs for it (just the desktop version).
So to simplify the API & usage i would keep the section header button in the top level as that's clearer where it'll be shown.

gabriellferreira: cool, I'll raise a thread in the group later - because we will need to align with iOS too 😄

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
gabriellferreira: fixed ✅

gabriellferreira: fixed ✅

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
gabriellferreira: great point. Just removed and added the minimum cards to show instead - which is handy because we need it for the "adapter"

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/internal/BpkCardListImpl.kt
gabriellferreira: also fixed ✅

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/internal/BpkCardListImpl.kt
```diff
@@ -0,0 +1,87 @@
...
+
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+fun RailLayout(
+    initiallyShownCards: Int,
+    modifier: Modifier = Modifier,
+    content: @Composable (LazyItemScope.(Int) -> Unit),
+) {
+    val lazyListState = rememberLazyListState()
+    LazyRow(
```
marianeum, line 75: it might be worth checking with design if this should behave like a pager or just a row like this. i would expect it to be more a pager, like our other carousels in which case we'd have to use `HorizontalPager`

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
```diff
@@ -0,0 +1,50 @@
...
+import net.skyscanner.backpack.compose.sectionheader.BpkSectionHeaderButton
+
+@Composable
+fun BpkCardList(
+    title: String,
+    description: String,
+    layout: BpkCardListLayout,
+    modifier: Modifier = Modifier,
+    headerButton: BpkSectionHeaderButton? = null,
+    initiallyShownCards: Int = 3,
```
marianeum, line 34: this is something not very clear in the api table, but this only really makes sense in the stack version where there is an expand option.
i think we still need to expose a state, like in carousel with the size of the items, current position etc otherwise the component won't know how many items there are

gabriellferreira: any suggestion how should we proceed here?

marianeum: So I think the best approach that will be backwards compatible for when we add the stack variant as well is to actually have a `BpkRailCardList` composable (similar to how we have the single vs multi select chip group) and remove the `layout` param.
Thinking through the API for the stack version again I think this will need to look quite different to be able to support nesting scrolling etc (we won't be able to have a `LazyColumn` internally as you can't nest `LazyColumn`s with flexible height)

For the rail variant internally the card list should use a `HorizontalPager` with the `pageSize` set (it's worth looking at `HotelMapCardsCarousel` in the app for that) - that will then set the size of the items correctly and do the docking. Again - worth double checking with design if that's what we want or not, but I suspect we do :)
For the state: we can then reuse the `BpkCarouselState` and add it as a parameter in the API. However it is worth double checking with design again if we want to have the endless scrolling here as well. If not we can just expose the default pager state.
Give me a shout if you have more questions - happy to pair on this.

docs/compose/CardList/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+BpkCardList(
+    title = stringResource(R.string.card_list_title),
+    description = stringResource(R.string.card_list_description),
+    layout = BpkCardListLayout.Rail,
+    initiallyShownCards = INITIALLY_SHOWN_CARDS,
+    modifier = Modifier,
+    content = dataList,
+) { position ->
+    CardLayout(dataList[position])
```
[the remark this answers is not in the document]
gabriellferreira: great point

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/BpkCardList.kt
gabriellferreira:
```
So I think the best approach that will be backwards compatible for when we add the stack variant as well is to actually have a BpkRailCardList composable (similar to how we have the single vs multi select chip group) and remove the layout param.
```
Great point, I'll migrate the code for this style
Screenshot 2024-05-29 at 15 43 22

about the `HorizontalPager`, I'll check

docs/compose/CardList/README.md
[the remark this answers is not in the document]
gabriellferreira: done ✅

marianeum (approved):
nice one!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/internal/BpkCardListImpl.kt
gabriellferreira: As agreed this will be handled later.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/rail/BpkRailCardList.kt
```diff
@@ -16,16 +16,16 @@
...
-import net.skyscanner.backpack.compose.cardlist.internal.BpkCardListImpl
+import net.skyscanner.backpack.compose.cardlist.rail.internal.BpkRailCardListImpl
 import net.skyscanner.backpack.compose.sectionheader.BpkSectionHeaderButton
 
 @Composable
-fun BpkCardList(
+fun BpkRailCardList(
     title: String,
     description: String,
     layout: BpkCardListLayout,
```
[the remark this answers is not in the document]
gabriellferreira: Ops haha
on it 😄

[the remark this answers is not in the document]
gabriellferreira: done ✅

PR #2003, opened 2024-06-04
RALP-4861 make page indicator invisible to screenreader

LewisI224:
[RALP-4861](https://skyscanner.atlassian.net/browse/RALP-4861)

Previously the screenreader could move focus from the image to the page indicator element, this caused repeated information to be read out to the user which is potentially confusing. The page indicator gives no additional info on top of what is already provided by the narration from the image so best approach is to make the indicator invisible to the user

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/imagegallery/BpkImageGalleryCarousel.kt
```diff
@@ -65,9 +69,10 @@ fun BpkImageGalleryCarousel(
                 pageIndicator?.invoke()
                 BpkBadge(
                     modifier = Modifier
-                        .align(Alignment.BottomEnd),
+                        .align(Alignment.BottomEnd).semantics { this.invisibleToUser() },
```
marianeum, line 72: nit: you can use `invisibleSemantic` as a shortcut for this

PR #2014, opened 2024-06-14
[PAN-3132] Fix bpk nudger for subtitle long case

noah-wen:
### Screenshot
| Old| New|
| ----| ----|
| image| image|

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
```diff
@@ -160,8 +160,8 @@ fun BpkNudger(
                         text = subtitle,
                         style = BpkTheme.typography.bodyDefault,
                         color = BpkTheme.colors.textSecondary,
-                        maxLines = 1,
-                        overflow = TextOverflow.Ellipsis,
+                        maxLines = Int.MAX_VALUE,
+                        overflow = TextOverflow.Clip,
```
marianeum, line 164: nit: both lines can be removed as we're just using the default values

marianeum (commented):
can you please add a snapshot test for the long line variant?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/nudger/BpkNudger.kt
noah-wen: Oh, I just see.

noah-wen: Yeah, It upgraded.

app/src/main/java/net/skyscanner/backpack/demo/compose/NudgerStory.kt
```diff
@@ -72,6 +72,11 @@ fun NudgerStory(modifier: Modifier = Modifier) {
             modifier = Modifier.fillMaxWidth(),
         )
 
+        NudgerRowExample(
```
marianeum, line 75: this is good too, but what i meant is a snapshot test (so `BpkNudgerTest`)

app/src/main/java/net/skyscanner/backpack/demo/compose/NudgerStory.kt
```diff
@@ -72,6 +72,11 @@ fun NudgerStory(modifier: Modifier = Modifier) {
...
+            modifier = Modifier.fillMaxWidth(),
+            subtitle = "subtitle is an example of long text that can be wrapped",
```
marianeum, line 77: we generally use string resources - there's probably already a generic one that can be reused)

noah-wen: Right! Have updating with above comment too, could you help to take a look again? Thanks!

marianeum (approved):
thanks - looks good! when the snapshots are re-generated please push an empty commit and then ping me to merge

noah-wen:
> thanks - looks good! when the snapshots are re-generated please push an empty commit and then ping me to merge

Hey @marianeum , seems snapshots updated, I have push the empty commit just now.

PR #2033, opened 2024-07-03
Enable custom content description for the bottom sheet title

vivek-as-sky:
This PR introduces a new parameter to `BpkModalBottomSheet` called `titleContentDescription`. This parameter allows developers to provide a custom content description for the title of `BpkModalBottomSheet`. It enables developers to override the default content description, allowing customisation of the text read by screen readers in accessibility mode (e.g., to correct pronunciation issues).

+ [x] Component `README.md`
+ [x] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
@@ -183,7 +183,7 @@ private fun BpkModalBottomSheetHeader(
                         }
                         .applyIf(titleContentDescription != null) {
                             semantics {
-                                contentDescription = titleContentDescription ?: ""
+                                contentDescription = titleContentDescription!!
                             }
                         },
                 )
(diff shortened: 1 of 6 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
```diff
@@ -170,7 +177,15 @@ private fun BpkModalBottomSheetHeader(
...
                     style = BpkTheme.typography.heading5,
                     textAlign = TextAlign.Center,
-                    modifier = Modifier.semantics { heading() },
+                    modifier = Modifier
+                        .semantics {
+                            heading()
+                        }
+                        .applyIf(titleContentDescription != null) {
+                            semantics {
+                                contentDescription = titleContentDescription ?: ""
```
alejandrorosas, line 186: small one, but `titleContentDescription` should not be null at this point

alejandrorosas (approved):

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/bottomsheet/internal/BpkModalBottomSheetImpl.kt
vivek-as-sky: Yes true, but the editor was complaining since the null check is not in context, okay to replace with `!!` ?

alejandrorosas: let's use `!!` like we use in other similar situations like https://github.com/Skyscanner/backpack-android/blob/055475462b6d645476467fe9a151b2239638c8bd/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/button/internal/BpkButtonImpl.kt#L93
if at some point the compiler turns smarter we'll just need to remove those and we do not need to define empty strings as default

marianeum (approved):

PR #2038, opened 2024-07-05
Ogm 2609 android add icon as label support to calendar cell info

ernestkamara:
[OGM-2609](https://skyscanner.atlassian.net/browse/OGM-2609)

### Context

We’d like to re-run the _Change missing price on the calendar from "-" to "🔍" experiment_ but using a BpkIcon instead of an Emoji. However, the current implementation of the Calendar doesn’t support Icon as a label. This PR adds support for BpkIcon as a label.

Screenshot 2024-07-09 at 10 34 17

+ [x] Component `README.md`
+ [x] Tests


backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -136,7 +136,7 @@ data class CalendarParams(
 data class CellInfo(
     val disabled: Boolean = false,
     val status: CellStatus? = null,
-    val label: String? = null,
+    val label: CellLabel? = null,
```
alejandrorosas, line 139: just confirming, is this an intentional breaking change? just to set the correct labels and version backpack accordingly

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -145,6 +145,19 @@ data class CellInfo(
...
 }
 
+/**
+ * Describes the label to be added to the cell
+ *
+ * @param text the text to be displayed
+ * @param icon the icon to be displayed
+ * @param iconTint the tint of the icon
+ */
+data class CellLabel(
```
alejandrorosas, line 155: can we have a text and an icon at the same time? If not maybe a sealed class could make it easier to understand

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
andra-georgescu: I don't think we should make this a breaking change. Even if we only use it in one place in the Skyscanner app and wouldn't experience a regression ourselves, we need to keep in mind that this is an open sourced library and it should be kept backwards compatible if there isn't a good case for changing this type.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
andra-georgescu: Following up from the previous comment, if the intention is to make it possible to have both a label and an icon, then we can just leave the text out of here and rename this to CellIcon to wrap around the icon id and the tint.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -145,6 +145,19 @@ data class CellInfo(
...
+    val text: String? = null,
+    val icon: Int? = null,
```
andra-georgescu, line 157: Is there a reason for not using `BpkIcon` here? In our case we could pass `BpkIcon.Search` in. Looking at the screenshots in this PR, the icon is visibly pixelated due to using the smaller version. The size adaptability we'd get out of the box with `BpkIcon`.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
ernestkamara: Based on the current specifications, we can have both the label and the icon in the same calendar configuration. However, I agree that we should avoid making this change disruptive. I will revert the label and add the icon as a new variable, as suggested below by @andra-georgescu.

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
ernestkamara: @andra-georgescu To use BpkIcon directly, I would need to add the compose module as a dependency in the view-based module. Since this API is currently shared between both the compose and view-based versions of the calendar, this solution was the simplest one that came to mind. I'm open to exploring other options as well.

alejandrorosas (commented):
can you also add a screenshot test/s covering this new use case?

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
frugoman: i've a different view from Andra's on this one, depending on how the component should work, having the 2 properties (a label and an icon prop) could give the impression that you can have the 2 at the same time, wich doest seem to be the case. ie, i could set the label to `$50` and the icon to `.search`, but when I run the app i'd only see one of the 2, giving me unexpected behaviour.

if only one of the 2 needs to be shown at a given time, we need to make the API of the component enforce the correct behaviour, or else it just wont compile. I believe a sealed class is best in this case with associated values (as in, an enum with values) so it can either be a label or an icon, but not both 😬

frugoman: (or have 2 constructors for the data class, that'd also work)

alejandrorosas (approved):

backpack-common/src/main/java/net/skyscanner/backpack/calendar2/CalendarParams.kt
```diff
@@ -145,6 +145,17 @@ data class CellInfo(
     }
 }
 
+/**
+ * Describes the label of the cell
+ */
+sealed class CellLabel {
+    data class Text(val text: String) : CellLabel()
+    data class Icon(
+        val resId: Int? = null,
```
andra-georgescu, line 154: I think this could be non-nullable. Can't see a case where you'd create an `Icon` without a resource id?

ernestkamara: Good call, I'll update it. Thanks

Backpack/src/main/java/net/skyscanner/backpack/calendar2/list/CalendarCellDayHolder.kt
```diff
@@ -66,7 +66,7 @@ internal class CalendarCellDayHolder(
         when (val cellLabel = model.info.label) {
             is CellLabel.Text -> label.text = cellLabel.text
             is CellLabel.Icon -> {
-                cellLabel.resId?.let { resId -> icon.setImageResource(resId) }
+                cellLabel.resId.let { resId -> icon.setImageResource(resId) }
```
andra-georgescu, line 69:
```suggestion
                icon.setImageResource(resId)
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -141,7 +141,7 @@ internal fun BpkCalendarDayCell(
                     }
                 }
                 is CellLabel.Icon -> {
-                    cellLabel.resId?.let { resId ->
+                    cellLabel.resId.let { resId ->
```
andra-georgescu, line 144: no need for this let check either

andra-georgescu (approved):

PR #2078, opened 2024-08-30
DON-739 - Update Search Input Summary based on new designs

henrik-sky:
Based on the [new designs](https://www.figma.com/design/RkSuI77vgPXkTf2SvuEvKf/Apps---Search-controls-GC?node-id=1459-117949&t=c2sC1YUUrQOFgfz9-4) for the Date Selector, we need to make some slight changes to the Search Input Summary component.

Support there being no prefix and allowing for it to be read only.
I additionally had to make a modification deeper in `BpkTextFieldImpl` so that the clear button will appear even though it is read only (this is why there is the `None` prefix rather than it being optional).

I added the new case to the bottom of the screen in the app, see below.

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/textfield/internal/BpkTextFieldImpl.kt
```diff
@@ -86,6 +86,7 @@ internal fun BpkTextFieldImpl(
     minLines: Int = 1,
     maxLines: Int = 1,
     interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    isFocused: Boolean? = null,
```
bvitaliyg, line 89: Why do we need to add this field? Can we always use the mutable interaction source instead?

henrik-sky: We found it caused issues with the screen reader when changing the focus state programatically.

## Before

https://github.com/user-attachments/assets/93455770-979f-4c68-877d-b7ddbe5abb10

## After
https://github.com/user-attachments/assets/acdd051d-7778-4390-af0e-51626b8c639e

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/searchinputsummary/BpkSearchInputSummary.kt
```diff
@@ -45,15 +52,18 @@ fun BpkSearchInputSummary(
...
         value = inputText,
         onValueChange = onInputChanged,
         modifier = modifier,
+        readOnly = type is BpkSearchInputSummaryType.ReadOnly,
         placeholder = inputHint,
         prefix = prefix,
         status = BpkFieldStatus.Default,
         clearAction = clearAction,
         type = BpkTextFieldType.Search,
+        isFocused = type is BpkSearchInputSummaryType.ReadOnly && type.isFocused,
```
marianeum, line 67: small thing here: I think it'd be better to change this condition to only set a non-null value for the `ReadOnly` type as the default input one should use the system focused state

marianeum (approved):

PR #2097, opened 2024-09-26
Fix modal statusbar icon colour

marianeum:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModal.kt
```diff
@@ -72,8 +74,12 @@ fun BpkModal(
         val activityWindow = getActivityWindow()
         val dialogWindow = getDialogWindow()
         val parentView = LocalView.current.parent as View
+        val isSystemInDarkTheme = isSystemInDarkTheme()
         SideEffect {
             if (activityWindow != null && dialogWindow != null) {
+                if (!isSystemInDarkTheme) {
```
marianeum, line 80: couldn't find that solution earlier but turns out i was just looking at the wrong window and have to set it for the dialog window (dark mode is correct, so only need it for light mode)

docs/view/FloatingActionButton/screenshots/default.png
marianeum, line 1: this one was taken while animating last time, so to avoid it being regenerated every time update

xo0ps (approved):

PR #2103, opened 2024-10-03
Fix modal navigation bar

marianeum:
Turns out it was wrong after all. Also update screenshots to be edge to edge to be correct.

+ [ ] Component `README.md`
+ [ ] Tests


.idea/compiler.xml
marianeum, line 1: post gradle update change

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/modal/BpkModal.kt
```diff
@@ -79,6 +79,7 @@ fun BpkModal(
             if (activityWindow != null && dialogWindow != null) {
                 if (!isSystemInDarkTheme) {
                     WindowInsetsControllerCompat(dialogWindow, dialogWindow.decorView).isAppearanceLightStatusBars = true
+                    WindowInsetsControllerCompat(dialogWindow, dialogWindow.decorView).isAppearanceLightNavigationBars = true
```
marianeum, line 82: this is the fix for the modal

.idea/compiler.xml
alejandrorosas: weird that it's 21 for intellij but we have 17 setup everywhere 😅

alejandrorosas (approved):

.idea/compiler.xml
marianeum: I know - i undid this and it came back again, so I think it really wants it 🤷‍♀️

PR #2130, opened 2024-11-01
[MUON-624] - Expose interactionSource to support custom animations

gabriellferreira:
https://skyscanner.atlassian.net/browse/MUON-624

Exposing `interactionSource` so we could customize ripple animation to match the old cards in the Homescreen

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/graphicpromotion/internal/BpkGraphicPromoImpl.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/graphicpromotion/internal/BpkGraphicPromoImpl.kt
@@ -78,6 +78,7 @@ import net.skyscanner.backpack.compose.utils.isTablet
 internal fun BpkGraphicPromoImpl(
     headline: String,
     image: @Composable BoxScope.() -> Unit,
+    interactionSource: MutableInteractionSource,
     modifier: Modifier = Modifier,
     kicker: String? = null,
     subHeadline: String? = null,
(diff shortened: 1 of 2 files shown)


gabriellferreira:
I could also add this parameter to the example in the readme, but might not be necessary as this parameter is optional.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/graphicpromotion/internal/BpkGraphicPromoImpl.kt
```diff
@@ -86,7 +87,6 @@ internal fun BpkGraphicPromoImpl(
     verticalAlignment: BpkGraphicPromoVerticalAlignment = BpkGraphicPromoVerticalAlignment.Top,
     sponsor: BpkGraphicsPromoSponsor? = null,
     sponsorLogo: (@Composable () -> Unit)? = null,
-    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
```
bvitaliyg, line 89: NIT: can you keep these changes on the same lines?

bvitaliyg (commented):
Just one nit

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/graphicpromotion/internal/BpkGraphicPromoImpl.kt
gabriellferreira: unfortunately no, because lint force to move up when default value is not set

marianeum (approved):

PR #2249, opened 2025-03-25
Expose onClick on BkpCardWrapper

fpiardi:
At the moment **BpkCardWrapper** is lacking the ability to set click listener, since internal Card is consuming all clicks. In order for card clicks to work, we need to forward onClick callback the same way as it's done in BpkCard.

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardwrapper/BpkCardWrapper.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardwrapper/BpkCardWrapper.kt
@@ -68,3 +68,39 @@ fun BpkCardWrapper(
         }
     }
 }
+
+@Composable
+fun BpkCardWrapper(
+    onClick: () -> Unit,
+    backgroundColor: Color,
+    headerContent: @Composable () -> Unit,
+    cardContent: @Composable () -> Unit,
+    modifier: Modifier = Modifier,
+    cardPadding: BpkCardPadding = BpkCardPadding.Small,
+    corner: BpkCardCorner = BpkCardCorner.Small,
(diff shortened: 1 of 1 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardwrapper/BpkCardWrapper.kt
```diff
@@ -68,3 +68,39 @@ fun BpkCardWrapper(
         }
     }
 }
+
+@Composable
+fun BpkCardWrapper(
+    onClick: () -> Unit,
```
LokmaneKrizou, line 74: QQ: Before updating Backpack component API, did you check using Modifier.clickable
```
BpkCardWrapper(
        modifier = Modifier .clickable {onClick(state.id)}
```
)

LokmaneKrizou: Also its worth to mention, when changing an API, we always request our contributors to:
* Provide new API table confirmed by both platforms
* Update Documentation

fpiardi: It works, thank you. I'll close the PR

vganin: I remember regular Compose `Card` (which `BpkCard` is just based upon) having problems with ripple effect not being set if using `Modifier.clickable` instead of explicit `onClick`. Does it work now as well?

PR #2254, opened 2025-04-01
DON-1319 Add scroll callback to BpkCalendar

henrik-sky:
Add an optional callback to observe scrolling in the list of dates. (Which is a list of months).

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
@@ -36,6 +36,7 @@ import net.skyscanner.backpack.calendar2.data.dispatchClick
 import net.skyscanner.backpack.calendar2.data.dispatchSetSelection
 import java.io.Serializable
 import java.time.LocalDate
+import java.time.YearMonth
 
 class BpkCalendarController(
     initialParams: CalendarParams,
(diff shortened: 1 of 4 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
```diff
@@ -89,11 +90,14 @@ class BpkCalendarController(
         }
     }
 
+    val firstVisibleItemMonth: YearMonth
```
LokmaneKrizou, line 93: since both iOS and Android agreed to only exposing `onScrollToMonth()` callback, and since i don't see u using `firstVisibleItemMonth` in your PoC, i am assuming this is used internally, right?

docs/compose/Calendar2/README.md
```diff
@@ -89,6 +89,12 @@ This property is a lambda that will be called when a date is selected by the use
 You can use it to handle the selection manually.
 You can then set the selection in the controller using the `setSelection` method.
 
+### (Optional) Handling Calendar Scroll Updates
+
+To receive notifications when the calendar is scrolled, you can utilize the `onScrollToMonth` parameter.
+This callback is triggered whenever the `lazyGridState.firstVisibleItemIndex` property of the `BpkCalendarController` is
+updated.
+
```
LokmaneKrizou, line 97: can we have an example of how consumers can use `onScrollToMonth `

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
henrik-sky: Good point, I'll update it to `internal`

LokmaneKrizou (approved):
LGTM, only few questions to address before merging the PR, well done!

peterInTown (approved):

PR #2257, opened 2025-04-07
OGM-3426 CardList wth Stack layout

xo0ps:
https://skyscanner.atlassian.net/browse/OGM-3426

This PR adds the Stack variant of the CardList component.
It supports 2 type of accessory, Expand and Button.
It supports section header button.

Figma: https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=33783-33859&p=f&t=q6rQYG9r6FloFwkh-0
iOS impl: https://www.skyscanner.design/latest/components/card-list/swift-ui-2uNePvch

+ [ ] Component `README.md`
+ [ ] Tests


.idea/inspectionProfiles/ktlint.xml
```diff
@@ -58,10 +66,14 @@
...
+      <option name="previewFile" value="true" />
+    </inspection_tool>
     <inspection_tool class="PreviewPickerAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
     <inspection_tool class="RedundantSemicolon" enabled="true" level="ERROR" enabled_by_default="true" />
   </profile>
-</component>
+</component>
```
LokmaneKrizou, line 79: are these changes necessary they are inside .idea directory which is a local one if i recall correctly 😅

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    var isExpanded by rememberSaveable { mutableStateOf(false) }
+    val targetCount = when (accessoryStyle) {
+        is BpkStackCardAccessoryStyle.Expand ->
+            if (isExpanded) accessoryStyle.expandedCount else accessoryStyle.collapsedCount
+        else -> totalCount
+    }
+    val animatedVisibleCount by animateIntAsState(targetValue = targetCount)
+
+    LazyColumn(
```
henrik-sky, line 67: I think you should leverage the `LazyColumn` more, right now there's one item. I think you an instead have:
```
 LazyColumn(
        modifier = modifier,
        contentPadding = PaddingValues(start = BpkSpacing.Base, end = BpkSpacing.Base),
        verticalArrangement = Arrangement.spacedBy(BpkSpacing.Base),
    ) {
        item {
            BpkSectionHeader(
                title = title,
                description = description,
                modifier = Modifier,
                button = headerButton,
                accessibilityHeaderTagEnabled = accessibilityHeaderTagEnabled,
            )
        }
        item {
            Spacer(modifier = Modifier.height(BpkSpacing.Base))
        }

        items(animatedVisibleCount) {
            content(it)
        }

        accessoryStyle?.let {
            when (it) {
                is BpkStackCardAccessoryStyle.Expand -> {
                    item {
                        BpkButton(
                            modifier = Modifier
                                .fillMaxWidth()
                                .padding(top = BpkSpacing.Base, bottom = BpkSpacing.Base),
                            text = if (isExpanded) it.collapsedText else it.expandText,
                            icon = if (isExpanded) BpkIcon.ChevronUp else BpkIcon.ChevronDown,
                            position = BpkButtonIconPosition.End,
                            type = BpkButtonType.Link,
                            onClick = { isExpanded = !isExpanded },
                        )
                    }
                }

                is BpkStackCardAccessoryStyle.Button -> {
                    if (headerButton == null) {
                        item {
                            BpkButton(
                                modifier = Modifier
                                    .fillMaxWidth()
                                    .padding(top = BpkSpacing.Base, bottom = BpkSpacing.Base),
                                text = it.title,
                                type = BpkButtonType.Primary,
                                position = BpkButtonIconPosition.End,
                                icon = it.icon,
                                onClick = it.onClick,
                            )
                        }
                    }
                }
            }
        }
    }
}
```

henrik-sky: This also lets you use
```
items(targetCount) {
    Box(Modifier.animateItem()) {
        content(it)
    }
}
```
to smoothly animate the items appearing/disappearing

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
```diff
@@ -0,0 +1,58 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
```
DanJayF, line 4: nit: Should this read 2025?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+    content: @Composable ((Int) -> Unit),
+) {
+
+    var isExpanded by rememberSaveable { mutableStateOf(false) }
+    val targetCount = when (accessoryStyle) {
+        is BpkStackCardAccessoryStyle.Expand ->
+            if (isExpanded) accessoryStyle.expandedCount else accessoryStyle.collapsedCount
+        else -> totalCount
+    }
+    val animatedVisibleCount by animateIntAsState(targetValue = targetCount)
```
DanJayF, line 65: I'm not too keen on using an animated Integer to control the 'expand' animation here. It just makes it look a bit janky compared to iOS. Firstly it feels way too fast by default, but you can also see each individual card appear one after the other to form the animation, which doesn't feel as smooth as it could be.

Is there a way we could instead use the default LazyList diff animation by updating the list to contain either `totalCount` of items or `collapsedCount` based on the state of `isExpanded`? I'd imagine when collapsed, we'd need to construct a sub-list like `content.take(collapsedCount)`, then replace that with the full list when `isExpanded == true`. If we use the default list animation it should result in a much smoother transition between expanded and collapsed states. Here's the iOS component animation:

https://github.com/user-attachments/assets/bae70e92-d651-488e-8e75-1ead49f30bfb

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
```diff
@@ -0,0 +1,58 @@
...
+        val expandText: String,
+        val collapsedText: String,
+    ) : BpkStackCardAccessoryStyle()
+    data class Button(val title: String, val icon: BpkIcon, val onClick: () -> Unit) : BpkStackCardAccessoryStyle()
+}
+
+@Composable
+fun BpkStackCardList(
+    title: String,
+    description: String,
```
LokmaneKrizou, line 40: description is optional according to `BpkCardList` API Table

DanJayF (changes requested):

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
```diff
@@ -0,0 +1,58 @@
...
+    totalCount: Int,
+    modifier: Modifier = Modifier,
+    accessoryStyle: BpkStackCardAccessoryStyle? = null,
+    headerButton: BpkSectionHeaderButton? = null,
+    accessibilityHeaderTagEnabled: Boolean? = true,
+    content: @Composable ((Int) -> Unit),
```
LokmaneKrizou, line 46: The API is different than what Figma contains, we supposed to have one API `BpkCardList` and  both Stack and Rail passed as a `layout`
any reasons behind this differences?

LokmaneKrizou:
On tablet and large screen devices, specifically `BpkStackCardList`, should we support displaying the items in grid? 
check [breakpoints section](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?m=auto&node-id=33783-40996)

image

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+                    is BpkStackCardAccessoryStyle.Expand -> {
+                        BpkButton(
+                            modifier = Modifier.fillMaxWidth()
+                                .padding(top = BpkSpacing.Base, bottom = BpkSpacing.Base),
+                            text = if (isExpanded) it.collapsedText else it.expandText,
+                            icon = if (isExpanded) BpkIcon.ChevronUp else BpkIcon.ChevronDown,
+                            position = BpkButtonIconPosition.End,
+                            type = BpkButtonType.Link,
+                            onClick = { isExpanded = !isExpanded },
+                        )
```
DanJayF, line 99: I think we need to fix the touch feedback on this "Show more" button because it's way too big right now and is quite distracting. I think we should either restrict the dimensions of the button to wrap just the text (so not full width), or disable the touch feedback completely. Donburi and Adam Wilson may need to make a call on this as I'm not sure what's best. Here's how it looks currently:

https://github.com/user-attachments/assets/0a69ddb8-dfc3-4c87-bfc1-157ef0a7d626

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
xo0ps: It's not generated by a script, I just copied from another class. I'll fix this one but Donburi should think about a solution for it on all files! 😅

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
xo0ps: good point. I fixed it for both Stack and Rail variants.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
xo0ps: The reason is the way the Rail variant was implemented. They decided to change the implementation to be 2 different implementations.
https://github.com/Skyscanner/backpack-android/pull/1986
Also I cannot change them back to align with the API table now as it will cause a breaking change.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
xo0ps: Thank you @henrik-sky for the great suggestion. I applied and it works better than the animatedInt way.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
xo0ps: Thanks @DanJayF for the suggestion. I applied Henrik's solution

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
xo0ps: Well I think the Button accessory is designed in full width, so the Expand accessory should follow.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
```diff
@@ -0,0 +1,59 @@
...
+import net.skyscanner.backpack.compose.cardlist.stack.internal.BpkStackCardListImpl
+import net.skyscanner.backpack.compose.icon.BpkIcon
+import net.skyscanner.backpack.compose.sectionheader.BpkSectionHeaderButton
+
+sealed class BpkStackCardAccessoryStyle {
+    data class Expand(
+        val collapsedCount: Int,
+        val expandedCount: Int,
+        val expandText: String,
+        val collapsedText: String,
```
DanJayF, line 32: These two strings should also be nullable, as the API table [says they're not required](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=33783-40612&t=IfCRfDwzm5Ba0kYf-4). I suppose it make sense for the case of `"expand accessory with less than minimum expanded count"` since there's no point in passing Strings in if you're not planning to display the accessory.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+            Box(Modifier.animateItem()) {
+                content(it)
+            }
+        }
+
+        item {
+            accessoryStyle?.let {
+                when (it) {
+                    is BpkStackCardAccessoryStyle.Expand -> {
+                        if (it.expandedCount > MINIMUM_EXPANDED_COUNT) {
```
DanJayF, line 99: We shouldn't be using `MINIMUM_EXPANDED_COUNT` for this check IMO. This would result in always showing the accessory when `expandedCount  > 2`, but if `collapsedCount` AND `expandedCount` are both equal to 4, for example, we wouldn't want to show the accessory in this case. So I think we could be checking against `collapsedCount` instead:
```suggestion
                        if (it.expandedCount > it.collapsedCount) {
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import net.skyscanner.backpack.compose.cardlist.stack.internal.BpkStackCardListImpl
+import net.skyscanner.backpack.compose.icon.BpkIcon
+import net.skyscanner.backpack.compose.sectionheader.BpkSectionHeaderButton
+
+sealed class BpkStackCardAccessoryStyle {
+    data class Expand(
+        val collapsedCount: Int,
+        val expandedCount: Int,
```
DanJayF, line 30: I'm wondering if we should make `expandedCount` nullable and somehow default to `totalCount`. I'm just concerned that if devs need to manually input **both** `totalCount` and `expandedCount` when creating an instance of `BpkStackCardAccessoryStyle.Expand`, then there's margin for human error whereby they could accidentally truncate the full list like this without realising:
```
BpkStackCardList(
    ...
    totalCount = 10,
    accessoryStyle = BpkStackCardAccessoryStyle.Expand(
        collapsedCount = 2,
        expandedCount = 9, // This should default to 10 unless explicitly set
        ...
    ),
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+                            BpkButton(
+                                modifier = Modifier
```
DanJayF, line 101: We could add `animateItem()` to the accessory modifier as well, just to make the movement of the expand button smoother. I'd prefer it to slide down out of the way rather than just fade in and out, but maybe @henrik-sky you have a better solution for this?
```suggestion
                                modifier = Modifier
                                    .animateItem()
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+                                    it.onExpansionChange?.invoke(isExpanded)
+                                },
+                            )
+                        }
+                    }
+
+                    is BpkStackCardAccessoryStyle.Button -> {
+                        if (headerButton == null) {
+                            BpkButton(
+                                modifier = Modifier
```
DanJayF, line 119: We should add `animateItem()` here as well, as the button will need to move down after adding a new item.
```suggestion
                                modifier = Modifier
                                    .animateItem()
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
henrik-sky: I think I have the solution:
```
accessoryStyle?.let {
    when (it) {
        is BpkStackCardAccessoryStyle.Expand -> {
            item(it.expandText) {
                BpkButton(
                    modifier = Modifier
                        .animateItem()
                        .fillMaxWidth()
                        .padding(bottom = BpkSpacing.Base),
                    text = if (isExpanded) it.collapsedText else it.expandText,
                    icon = if (isExpanded) BpkIcon.ChevronUp else BpkIcon.ChevronDown,
                    position = BpkButtonIconPosition.End,
                    type = BpkButtonType.Link,
                    onClick = {
                        isExpanded = !isExpanded
                        it.onExpansionChange?.invoke(isExpanded)
                    },
                )
            }
        }
        is BpkStackCardAccessoryStyle.Button -> {
            if (headerButton == null) {
                item(it.title) {
                    BpkButton(
                        modifier = Modifier
                            .animateItem()
                            .fillMaxWidth()
                            .padding(bottom = BpkSpacing.Base),
                        text = it.title,
                        type = BpkButtonType.Primary,
                        position = BpkButtonIconPosition.End,
                        icon = it.icon,
                        onClick = it.onClick,
                    )
                }
            }
        }
    }
}
```

https://github.com/user-attachments/assets/271a7169-bcc7-4252-82f0-faeb039b9e1f


The individual componsables should be wrapped with `item` and a key provided so the list knows how to animate the position of the item as it knows it is the same item and so should change position not appear/disappear.

DanJayF: Ah of course, makes perfect sense now that you explain it 😂  Thanks again for the snippet!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
DanJayF: See [Henrik's comment above](https://github.com/Skyscanner/backpack-android/pull/2257#discussion_r2039537345) which includes this fix **_but better_**

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
```diff
@@ -0,0 +1,59 @@
...
+        val expandText: String,
+        val collapsedText: String,
+        val onExpansionChange: ((Boolean) -> Unit)? = null,
+    ) : BpkStackCardAccessoryStyle()
+    data class Button(val title: String, val icon: BpkIcon, val onClick: () -> Unit) : BpkStackCardAccessoryStyle()
+}
+
+@Composable
+fun BpkStackCardList(
+    title: String,
```
LokmaneKrizou, line 40: @xo0ps looking at this [figma design](https://www.figma.com/design/LaYKZu5Utu80iwnB5xKATE/Homepage---MVP?node-id=1604-41314&p=f&t=q6S7likgR0eqr3pj-0)  
you may need to change the title to something else than string as you could display `bpkIcon` as a prefix later on and not just a plain title. it is worth checking with design about this as [Card List backpack figma](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=33783-33859&p=f&t=vDcPUCB0AZNyuD8F-0) doesn't show that

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
xo0ps: If both `expandedCount ` and `collapsedCount ` are equal, it means the user intentionally wants to hide the expand accessory. For this use-case the user should set the expand accessory to null as it is optional.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/BpkStackCardList.kt
xo0ps: Makes sense. I applied your suggestion.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
xo0ps: Great suggestion @henrik-sky. I applied your suggestion and the accessory buttons animate now!

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
xo0ps: done

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
DanJayF: Same as [my reply above](https://github.com/Skyscanner/backpack-android/pull/2257#discussion_r2042268030), this makes sense now 🙏

DanJayF (approved):
Nice work on this @xo0ps and thanks for putting up with all our comments. The animation looks beautiful 😍 

https://github.com/user-attachments/assets/c906a703-02a3-476b-ac5b-344949fdabfd

LokmaneKrizou (approved):
Great work @xo0ps 🚀 LGTM

PR #2266, opened 2025-04-15
[DON-1368] add boolean for multicity highlighted

peterInTown:

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
@@ -40,6 +40,7 @@ import androidx.compose.ui.graphics.RectangleShape
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.isEditable
 import androidx.compose.ui.semantics.selected
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.semantics.stateDescription
@@ -93,6 +94,7 @@ internal fun BpkCalendarDayCell(
                 } else {
                     selected = selection != null && selection != Selection.Middle
                 }
+                isEditable = model.info.highlighted
             },
     ) {
         Box(


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/internal/BpkCalendarDayCell.kt
```diff
@@ -93,6 +94,7 @@ internal fun BpkCalendarDayCell(
                 } else {
                     selected = selection != null && selection != Selection.Middle
                 }
+                isEditable = model.info.highlighted
```
peterInTown, line 97: We have two way to support showing highlighted info in the semantic/ ui test:
1. create a costume property to hold the information
`val CircledKey = SemanticsPropertyKey<Boolean>("Circled")
var SemanticsPropertyReceiver.isCircled by CircledKey`
2. or we use `isEditable ` although it is not relevant to the property, but it serves the same purpose

LokmaneKrizou: Here is another option we may consider: 

3 - Add `highlighted` state directly to the `testTag`, similar to how we currently indicate `active` vs `inactive` states. Instead of an if/else for only `inactive` and `active`, we would switch to a when-statement that produces one of the four tags as enums:
```
enum class CalendarDayCellTestTag(val tag:String){
    INACTIVE("inactive"),
    INACTIVE_HIGHLIGHTED("inactive_highlighted"),
    ACTIVE("active"),
    ACTIVE_HIGHlIGHTED("active_highlighted")
}
```
I would say option 3 isolates test-specific state information within our `testTag` string, ensuring that nothing we add will affect the accessibility announcements (unlike using system semantics like `isEditable`). Although `TalkBack` may handle the misuse of `isEditable` properly in our case and not announce it,  we want to avoid any system semantics that could be announced by other screen readers or lead to unexpected behaviour.

While Option 1 using a custom semantics property as [Google recommends](https://developer.android.com/develop/ui/compose/testing/common-patterns#custom-semantics-properties) when trying to expose information to tests, Option 3 is the safest in that it entirely avoids dealing with semantics properties. The only downside is that the testTags may become more complex as additional states pops up.

I would choose between Option 1 and Option 3 which option you think we should go with?

peterInTown: I think option 1 and option 3 both are good approaches. However, I was thinking if option3 becomes too complex we might end up switching to option 1, but at the same time, we might need to fix lots of failing tests in the main app for consumers who have adopted option 3. In this case, we might just go for option 1.

peterInTown: Turns out to use constume semantics property we need to in someway share the key to where we test it, at the moment we are not sharing costume semantics key between backpack and skysanner app. Therefore, I went with option 1 to not overcomplicate this work

henrik-sky (approved):

PR #2274, opened 2025-04-25
DON-1396 Add Loading Label to BpkCalendarDayCell, update BpkSkeleton

henrik-sky:
This adds a new form of Label in the BpkCalendar, Loading.

As this is the smallest BpkSkeleton in use the BpkShimmerOverlay looked odd so the designers cooked up something new for small shimmers.
Therefore I added an enum to allow the consumer to set the size of shimmering animation.

I have also wrapped the Calendar label with `AnimatedContent` allowing for a smooth transition between the loading and the text label/icon.
The transition is delayed by 2x the length of the shimmer animation (design requirement).

Further documentation around our decision making for this can be found [here](https://skyscanner.atlassian.net/wiki/x/8UgkU):

Design:

https://github.com/user-attachments/assets/157e04b4-c55a-48ff-91a6-4a0c1ca81617

Built:

https://github.com/user-attachments/assets/c99bd2fe-0548-4d44-8d5e-5a649b70afad

+ [x] Component `README.md`
+ [x] Tests


app/screenshots/oss/debug/default/net.skyscanner.backpack.compose.calendar.BpkCalendarTest_default.png
henrik-sky, line 1: Spacing is updated in figma
https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=20677-22316&m=dev

peterInTown (approved):
LGTM

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
```diff
@@ -102,6 +102,13 @@ enum class BpkSkeletonCornerType {
     Rounded,
 }
 
+enum class BpkShimmerSize(
```
bvitaliyg, line 105: Should it be really called size? It looks more like timings for me. Is that the naming we use on Figma?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
```diff
@@ -153,10 +168,37 @@ private fun ShimmerBox(
...
+ * @param shimmerSize To get the right animation for the size of the component.
+ * @param content Child component.
+ */
+@Composable
+fun BpkShimmerOverlay(
+    modifier: Modifier = Modifier,
+    shimmerSize: BpkShimmerSize = BpkShimmerSize.Large,
+    content: @Composable BoxScope.() -> Unit,
+) {
+    BoxWithConstraints(modifier = modifier) {
```
bvitaliyg, line 183: `BoxWithConstraints` will have performance implications, especially when used in lists (which we do in many places). How about this: we'll animate the float "fraction/percent" value, then simply multiply it to the width that's available in graphicsLayer {} modifier?

bvitaliyg (commented):
A couple of small comments

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
henrik-sky: Ah yes, I think I had something similar previously but got confused with the `typeConverter` when making the animation but then today I realised it's just `Float.VectorConverter` so now it goes from

  `-1f * GraphicsLayerScope.size.width` to `1f * GraphicsLayerScope.size.width`

🚀

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
henrik-sky: It's the timings for the size of skeleton

bvitaliyg (approved):
Approved, but please make the timings internal for the size

PR #2283, opened 2025-05-06
DON-1469 Tweak animations

henrik-sky:
I realised that I wanted to delay the animation _between_ each cycle but instead I delayed the animation _before_ each cycle so you wouldn't see it animate most times. To fix this I added the shimmer delay to the BpkCalendarDayCell delay so we now see it and it's a bit smoother.

| Before | After |
| ----- | ----- |
|  <video src=https://github.com/user-attachments/assets/27af3963-147d-46a1-9a44-e996999edb8c /> | <video src=https://github.com/user-attachments/assets/029d908c-3074-461d-89e3-2ae1c620c4d7 /> |

Note 2 shimmers visible in the after video.

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
@@ -109,7 +109,7 @@ enum class BpkShimmerSize(
     internal val delayMillis: Int,
 ) {
     Large(durationMillis = 1000, delayMillis = 200),
-    Small(durationMillis = 300, delayMillis = 800),
+    Small(durationMillis = 300, delayMillis = 400),
 }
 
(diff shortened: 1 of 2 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/skeleton/BpkSkeleton.kt
```diff
@@ -109,7 +109,7 @@ enum class BpkShimmerSize(
     internal val delayMillis: Int,
 ) {
     Large(durationMillis = 1000, delayMillis = 200),
-    Small(durationMillis = 300, delayMillis = 800),
+    Small(durationMillis = 300, delayMillis = 400),
```
LokmaneKrizou, line 112: can u attach a video of before, after to highlight the difference, thank u

henrik-sky: Done

LokmaneKrizou (approved):

PR #2284, opened 2025-05-07
[SONIC-3515] Minor fixes in BpkSwitch

smilinghedgehog:
[SONIC-3515](https://skyscanner.atlassian.net/browse/SONIC-3515)

- Fixed text overlaping with the toggle + added support for truncation
- Added support for AnnotatedString
- Added support for controlling the toggle alignment inside the component

Screenshot_20250507_101222

[SONIC-3515]: https://skyscanner.atlassian.net/browse/SONIC-3515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/switch/BpkSwitch.kt
```diff
@@ -54,24 +58,64 @@ fun BpkSwitch(
     onCheckedChange: ((Boolean) -> Unit)?,
     modifier: Modifier = Modifier,
     enabled: Boolean = true,
+    shouldTruncate: Boolean = false,
```
LokmaneKrizou, line 61: QQ: 
I can see in the slack conversation, `shouldTruncate` should be set to `true` per default.
any reason why the default here is false, same for the ones bellow

LokmaneKrizou:
@smilinghedgehog Great work!
one note for future contributions, its better to set the label as minor, as we usually set patch for dependencies  and small updates.

LokmaneKrizou (approved):
LGTM @smilinghedgehog well done 🚀

PR #2288, opened 2025-05-08
[OGM-3529]Added column to backpack stack card list

Caoilinn:

+ [ ] Component `README.md`
+ [ ] Tests


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -66,12 +72,130 @@ internal fun BpkStackCardListImpl(
...
+    accessibilityHeaderTagEnabled: Boolean?,
+    totalCount: Int,
+    targetCount: Int,
+    accessoryStyle: BpkStackCardAccessoryStyle?,
+    modifier: Modifier = Modifier,
+    content: @Composable ((Int) -> Unit),
+) {
     LazyColumn(
         modifier = modifier,
         contentPadding = PaddingValues(start = BpkSpacing.Base, end = BpkSpacing.Base),
```
DanJayF, line 201: Same here, we should remove this padding even if we don't plan to use the LazyColumn.

app/src/main/java/net/skyscanner/backpack/demo/compose/CardListStory.kt
```diff
@@ -117,6 +117,7 @@ fun CardListStackWithExpandAccessoryStory(modifier: Modifier = Modifier) {
             title = stringResource(R.string.card_list_title),
             description = stringResource(R.string.card_list_description),
             totalCount = 5,
+            insideScrollableView = true,
```
DanJayF, line 120: We should have separate stories to showcase both versions of the column, otherwise the only way to see the `LazyColumn` version is to build it locally.

DanJayF (commented):
We've talked about this already in Origami, but just for Donburi's visibility, the list currently only animates when it expands, not when it's collapsing. Suggestions welcome 💡🙏  

https://github.com/user-attachments/assets/8580258b-3f2e-40fb-86db-5f42dfbf60cb

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -66,12 +72,130 @@ internal fun BpkStackCardListImpl(
...
+    accessibilityHeaderTagEnabled: Boolean?,
+    accessoryStyle: BpkStackCardAccessoryStyle?,
+    totalCount: Int,
+    targetCount: Int,
+    modifier: Modifier = Modifier,
+    content: @Composable (Int) -> Unit,
+) {
+    Column(
+        modifier = modifier
+            .padding(start = BpkSpacing.Base, end = BpkSpacing.Base)
```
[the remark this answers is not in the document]
Caoilinn: Yes thank you, completely forgot about the padding

app/src/main/java/net/skyscanner/backpack/demo/compose/CardListStory.kt
Caoilinn: Yup agreed, but I didn't want to add it until we all agreed this was the way forward

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
```diff
@@ -66,17 +71,129 @@ internal fun BpkStackCardListImpl(
...
+
         else -> totalCount
     }
 
-    LazyColumn(
-        modifier = modifier,
-        contentPadding = PaddingValues(start = BpkSpacing.Base, end = BpkSpacing.Base),
-        verticalArrangement = Arrangement.spacedBy(BpkSpacing.Base),
-    ) {
+    if (insideScrollableView) {
```
DanJayF, line 81: Correct me if I'm wrong, but should this boolean be the other way around? As in have the `ExpandableLazyColumn` be the scrollable one?

docs/compose/CardList/README.md
```diff
@@ -204,3 +204,126 @@ BpkStackCardList(
     // content
 }
 ```
+
+# Stack Column Variants 
+> **NOTICE**: The column varint should only be used when the compoent is to be lace inside of a scrollable View/Composable or one with infinite height
```
DanJayF, line 209: nit: Typo in `lace`

app/src/main/java/net/skyscanner/backpack/demo/compose/CardListStory.kt
DanJayF, line 1: Only a minor thing, but it may be easier to navigate this list with tabs for `Column`/`LazyColumn` because the variant naming has gotten quite long and confusing. But if it's a lot of work to implement, feel free to ignore. @Skyscanner/donburi tell me if you have a preference to make this easier to read.

| Without Tabs | With Tabs |
|--------|--------|
| Screenshot_20250509-152653 | Screenshot_20250509-152900 |

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
Caoilinn: I can see why it looks like that. It `insideScrollableView` as in will this be places inside a scrollable view, therefore it would need to be the Column version that would be used. But this is confusing so I'll change the name

app/src/main/java/net/skyscanner/backpack/demo/compose/CardListStory.kt
Caoilinn: Yeah I was looking into that, it looks like those tabs are handled by `[ComponentScreen](https://github.com/Skyscanner/backpack-android/blob/16b4267d4849be1496747850e7254b2e06b831a0/app/src/main/java/net/skyscanner/backpack/demo/ui/ComponentScreen.kt#L60)` so not too sure if there's a clean way to implement this. But I do agree

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/cardlist/stack/internal/BpkStackCardListImpl.kt
Caoilinn: So I've change the name so it's maybe a bit clearer, but if not we can change it to something like `useScrollable` and then flip the if

DanJayF (approved):
LGTM 🚀

app/src/main/java/net/skyscanner/backpack/demo/compose/CardListStory.kt
LokmaneKrizou: Thanks for raising that @DanJayF 
Usually when creating a story, we should focus on the component and its variants, not where the consumers will implement it. however, there is no rule or a policy on the current contribution guide enforces what contributors need to put on their component story. I'll raise this internally and come up with better guidance, and improvements to ComponentScreen to better organise stories if any updates are required.

DanJayF: Thanks @LokmaneKrizou, good suggestion. We'll amend these soon 👍

LokmaneKrizou (approved):

PR #2299, opened 2025-05-16
Updated references and tokens

EmmanuelBuraimo:
Updated color references in theme and component files, in line with updated Backpack requirements


Backpack/src/main/java/net/skyscanner/backpack/horisontalnav/BpkHorizontalNav.kt
```diff
@@ -60,7 +60,7 @@ open class BpkHorizontalNav @JvmOverloads constructor(
         Alternate(
             id = 1,
             styleAttribute = R.attr.bpkHorizontalNavStyleAlternate,
-            defaultTextColor = R.color.bpkSkyGrayTint07,
+            defaultTextColor = R.color.bpkCanvasContrast,
```
henrik-sky, line 63: As it is text should it not be `bpkTextOnDark` (though this style isn't used and can probably be entirely deleted)

henrik-sky (approved):

Backpack/src/main/java/net/skyscanner/backpack/horisontalnav/BpkHorizontalNav.kt
EmmanuelBuraimo: Screenshot 2025-05-16 at 16 31 33


very valid point but just following what was written in the doc

LokmaneKrizou: lets not update anything besides the tokens in xml files as they going to be deleted later in the next phase

PR #2308, opened 2025-05-26
DON-1639 Change scroll handling in BpkCalendar to return visible months

henrik-sky:
This supports taller screens better as if the calendar is scrolled to the end the consumer only knew what the top month was.
Now the consumer can observe a changing list of visible months, with no duplicates.

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
+++ b/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
@@ -90,8 +90,11 @@ class BpkCalendarController(
         }
     }
 
-    internal val firstVisibleItemMonth: YearMonth
+    private val firstVisibleItemMonth: YearMonth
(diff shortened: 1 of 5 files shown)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/calendar/BpkCalendarController.kt
```diff
@@ -90,9 +90,12 @@ class BpkCalendarController(
         }
     }
 
-    internal val firstVisibleItemMonth: YearMonth
+    private val firstVisibleItemMonth: YearMonth
         get() = _state.cells[lazyGridState.firstVisibleItemIndex].yearMonth
 
+    internal val visibleMonths: Set<YearMonth>
+        get() = lazyGridState.layoutInfo.visibleItemsInfo.map { _state.cells[it.index].yearMonth }.toSet()
```
LokmaneKrizou, line 97: Nice change @henrik-sky 
So now instead of passing only one month, you pass all visible months, curios to know what happens when a user scrolls by just half a month, what this callback will send?

PR #2336, opened 2025-07-03
Update missed Gradle file after failed deployment.

henrik-sky:
Update missed gradle file that refers to Sonatype still.

Robot says:
> With GitHub Packages, you do **not** need to sign artifacts (and thus do not need `signing.keyId`, `signing.password`, or `signing.secretKeyRingFile`) unless you specifically want to cryptographically sign your published artifacts for extra trust.
> GitHub Packages does not require artifact signing for publishing. Signing is only needed if you want consumers to verify the authenticity of your artifacts, or if you plan to also publish to Maven Central (which requires signing).
> If you only publish to GitHub Packages, you can safely remove the signing configuration and related secrets. If you want to keep signing for extra security, you can leave it as is.

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/gradle-maven-push.gradle
+++ b/gradle-maven-push.gradle
@@ -1,5 +1,4 @@
 apply plugin: 'maven-publish'
-apply plugin: 'signing'
 
 android {
     publishing {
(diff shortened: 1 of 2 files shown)


unlobito (approved):
lgtm 🚀 remove sonatype signing and explicitly define `GITHUB_TOKEN` as an environment variable

PR #2341, opened 2025-07-09
PAN-4378: Fix price map marker status change issue.

karljiang1:
https://skyscanner.atlassian.net/browse/PAN-4378

### Context
I found that the marker with different status but without size change can't trigger bitmap refresh so that we can't see the visual changes
Fix solution: Always refresh the bitmap if compose keys are different to avoid detail difference.

### ScreenRecords
|Before|After|
|-------|-----|
|Jul-09-2025 11-15-02|Jul-09-2025 11-15-52|

+ [x] Component `README.md`
+ [x] Tests


peterInTown (approved):
LGTM

PR #2356, opened 2025-07-23
DON-1853 Add uniform and progressive blur modifiers

henrik-sky:
Add modifiers allowing consumers to add blur to any Composable.
[Figma link](https://www.figma.com/design/KXf2gHNLDe2cXWUoHl4cTX/Backpack-Foundations?node-id=9848-459&t=UspLBKEcqWPbXZMg-4)

Progressive blur uses RenderEffect and so is only available on Android 13+.
AGSL was mostly written by ChatGPT based on how the [Skia Blur engine](https://github.com/google/skia/blob/main/src/core/SkBlurEngine.cpp) works.

| Uniform | Progressive |
| ------- | --------- |
| Screenshot_20250723_160819| Screenshot_20250723_160830 |

+ [ ] Component `README.md`
+ [ ] Tests


app/screenshots.gradle
```diff
@@ -23,7 +23,7 @@ android {
             managedDevices.devices {
                 Docs(ManagedVirtualDevice) {
                     device = "Pixel"
-                    apiLevel = 31
+                    apiLevel = 34
```
LokmaneKrizou: is this part of the scope of this work?

henrik-sky: > Note: The blur effect is only supported on Android 12 and above. Attempts to use this Modifier on older Android versions are ignored.
https://developer.android.com/develop/ui/compose/graphics/images/customize#blur_an_image_composable

So I needed to bump the version in order for the effect to be visible in the documentation screenshots.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/blur/BpkBlur.kt
```diff
@@ -0,0 +1,112 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.blur
+import androidx.compose.ui.draw.clipToBounds
+import androidx.compose.ui.graphics.asComposeRenderEffect
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.unit.dp
+import org.intellij.lang.annotations.Language
+
+fun Modifier.bpkUniformBlur() = this.blur(radius = 4.dp)
```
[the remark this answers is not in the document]
LokmaneKrizou: nice one partner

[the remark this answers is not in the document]
LokmaneKrizou: @henrik-sky can we at least have that as global variable, it is being defined in multiple places in this file.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/blur/BpkBlur.kt
```diff
@@ -0,0 +1,130 @@
...
+                val combinedEffect = RenderEffect.createChainEffect(blurEffectY, blurEffectX)
+
+                this.renderEffect = combinedEffect.asComposeRenderEffect()
+            }
+    } else {
+        bpkUniformBlur()
+    }
+
+@Language("AGSL")
+val BLUR_SHADER = """
```
LokmaneKrizou: can we have this in a separate file?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/blur/BpkBlur.kt
```diff
@@ -0,0 +1,130 @@
...
+ * - The effect is clipped to the composable bounds.
+ * - The blur strength and direction are handled via uniforms for two-pass chaining (X then Y).
+ * - Falls back to a uniform blur on unsupported platforms.
+ *
+ * This modifier is designed to be accessible and themable, following Backpack's
+ * component standards. It does not interfere with touch targets, semantics, or
+ * theming, and is compatible with both light and dark modes.
+ */
+@Composable
+fun Modifier.bpkProgressiveBlur(): Modifier =
```
LokmaneKrizou: could we move the implementation to internal functions similarly to other componenets and use this just for API definition.
think of it as an interface

docs/compose/Blur/screenshots/progressive_dm.png
LokmaneKrizou, line 1: nit: is there a reason why the images don't have the same width and height
it would be cooler :v

henrik-sky: They are the same width/height when I run the app 😕 
Might be an emulator thing I'll take a look.

henrik-sky: Also would look better if it wasn't obscured by the system UI

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/blur/BpkBlur.kt
henrik-sky: You can have it in the implementation file.

LokmaneKrizou (approved):
WELL DONE🚀

PR #2475, opened 2025-10-31
DON-2189 VDL2 Changes for Cards

henrik-sky:
Modify the default elevation for cards to 0.dp
https://www.figma.com/design/Ou6rQB1IHt5NBCoQc31PoJ/Cards-Panels---Shadows?node-id=231-41890&m=dev

### 📱 UI Comparison

| Mode | State | Before | After |
|------|--------|---------|--------|
| **Light** | Default View | Before Light | After Light |
| **Dark** | Default View | Before Dark | After Dark |
| **Light** | Compose | Compose Before Light | Compose After Light |
| **Dark** | Compose | Compose Before Dark | Compose After Dark |

+ [ ] Component `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/src/main/java/net/skyscanner/backpack/card/BpkCardView.kt
+++ b/Backpack/src/main/java/net/skyscanner/backpack/card/BpkCardView.kt
@@ -61,8 +61,7 @@ open class BpkCardView @JvmOverloads constructor(
         fun toDimension(context: Context): Float =
             when (this) {
                 NONE -> 0f
-                DEFAULT -> BpkConfiguration.cardConfig?.defaultElevation?.value?.times(context.resources.displayMetrics.density)
-                    ?: context.resources.getDimension(R.dimen.bpkElevationSm)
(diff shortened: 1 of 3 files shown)


Backpack/src/main/java/net/skyscanner/backpack/card/BpkCardView.kt
```diff
@@ -60,7 +61,9 @@ open class BpkCardView @JvmOverloads constructor(
         fun toDimension(context: Context): Float =
             when (this) {
                 NONE -> 0f
-                DEFAULT -> context.resources.getDimension(R.dimen.bpkElevationSm)
+                DEFAULT -> BpkConfiguration.cardConfig?.defaultElevation?.value?.times(context.resources.displayMetrics.density)
+                    ?: context.resources.getDimension(R.dimen.bpkElevationSm)
```
LokmaneKrizou: if  ` val defaultElevation: Dp = 0.dp`  then shouldn't that mean DEFAULT is simply 0f when configuration is enabled for card?

henrik-sky: GOOD point

PR #2480, opened 2025-11-04
DON-2267 Use caption for VDL2 badges

henrik-sky:
Didn't put the values in `BpkConfiguration` as they are in the theme so it would be annoying. Do as I say not as I do 😇 .

image

image

No change for VDL off (see no new screenshots after snapshot run)

+ [ ] Component `README.md`
+ [ ] Tests


peterInTown (approved):

RichSutherSky (approved):
LGTM

PR #2522, opened 2025-12-12
[DON-2137] Spike: Parallelize CI build workflow

peterInTown:
## Summary

This PR implements a proof of concept for parallelizing the CI build workflow to reduce PR validation time.

### Changes
- Split the monolithic Build job into 4 independent parallel jobs:
  - **Detekt** - Static code analysis **Lint** - Android lint checks
  - **UnitTests** - Unit test execution
  - **BuildAndTokens** - Build + token validation

### Expected Results
| Metric | Current | Proposed | Change |
|--------|---------|----------|--------|
| Wall-clock time | ~7 min | ~4 min | **~43% faster** |
| Runner minutes | ~21 min | ~24 min | +14% |

### Documentation
- [Confluence: Investigation of CI tasks and potential improvement](https://skyscanner.atlassian.net/wiki/spaces/Donburi/pages/1886618226/Android+Investigation+of+CI+tasks+and+potential+improvement)
- [Jira: DON-2137](https://skyscanner.atlassian.net/browse/DON-2137)

## Test plan
- [ ] Verify all parallel jobs complete successfully
- [ ] Measure actual wall-clock time improvement
- [ ] Compare runner minutes consumed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -31,12 +31,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Set up JDK
-        uses: actions/setup-java@v5
-        with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: gradle
+      - uses: ./.github/actions/setup-java
 
       - name: Detekt check
         run: ./gradlew detekt -PdisablePreDex
@@ -52,12 +47,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Set up JDK
-        uses: actions/setup-java@v5
-        with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: gradle
+      - uses: ./.github/actions/setup-java
 
       - name: Unit Tests
         run: ./gradlew test${{ env.flavour }}${{ env.config }}UnitTest -PdisablePreDex
@@ -102,12 +92,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
 
-      - name: Set up JDK
-        uses: actions/setup-java@v5
-        with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: gradle
+      - uses: ./.github/actions/setup-java
 
       - name: Validate Gradle Wrapper
         uses: gradle/actions/wrapper-validation@v5
@@ -161,12 +146,7 @@ jobs:
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
 
-      - name: Set up JDK
-        uses: actions/setup-java@v5
-        with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: gradle
+      - uses: ./.github/actions/setup-java
 
       - name: AVD cache
         uses: actions/cache@v4
@@ -234,12 +214,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
-      - name: Set up JDK
-        uses: actions/setup-java@v5
-        with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: gradle
+      - uses: ./.github/actions/setup-java
 
       - name: Screenshot Tests - ${{ matrix.variant.name }}
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/_build.yml
```diff
@@ -169,13 +209,59 @@ jobs:
...
+
+  Screenshots-Collect:
+    name: Collect & commit screenshots
     runs-on: ubuntu-24.04-16cores-public
+    needs: Screenshots
     permissions:
       pull-requests: write
       contents: write
-    timeout-minutes: 20
+    timeout-minutes: 5
```
LokmaneKrizou, line 239: Are we confident about dropping the timeout to 5 minutes? That’s a pretty aggressive reduction (20 → 5). If we’re confident that, after splitting, each job will consistently finish within 5 minutes, then I’m happy to keep it.

.github/workflows/_build.yml
```diff
@@ -169,13 +209,59 @@ jobs:
...
+      matrix:
+        variant:
+          - name: default
+            flag: default
+          - name: dark mode
+            flag: dm
+          - name: RTL
+            flag: rtl
+          - name: themed
+            flag: themed
```
LokmaneKrizou, line 212: **Not a blocker**
Is there a way to make this dynamic, instead of setting all variants manually could we read existing variants in the repo and extract the variant names

.github/workflows/_build.yml
```diff
@@ -18,10 +18,54 @@ env:
...
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up JDK
+        uses: actions/setup-java@v5
+        with:
+          java-version: '17'
+          distribution: 'adopt'
+          cache: gradle
```
LokmaneKrizou: **Not a blocker**
I noticed we set up the JDK in each job. Could we extract that into a reusable action and call it from the jobs that need it? That way we only update the Java version in one place instead of touching every job.

.github/workflows/_build.yml
```diff
@@ -188,18 +274,28 @@ jobs:
...
+        uses: actions/download-artifact@v4
         with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: gradle
+          path: screenshot-artifacts
 
-      - name: Screenshot Tests
-        id: screenshotTests
+      - name: Merge screenshot results
```
LokmaneKrizou: nit: could we rename this from merge to combine? We usually say “merged PR”, so “merge” here might make people think we’re actually merging screenshots rather than just combining them.

.github/workflows/_build.yml
peterInTown: Why 5 minutes should be sufficient:
  - This job doesn't run any tests or builds - it only downloads and merges artifacts
  - Artifact downloads are typically fast on GitHub Actions
  - File operations are simple bash commands (mkdir, cp, rm)
  - The previous 20-minute timeout was likely inherited from when this job also ran the actual screenshot tests

  Why it was safe to reduce:
  - The actual screenshot tests now run in parallel in the separate Screenshots matrix job (which has 10-minute timeout)
  - The collect job only does I/O operations, which are much faster than running tests
  - Even with network delays, 5 minutes provides a ~3-4x safety margin for this type of work

.github/workflows/_build.yml
peterInTown: It is possible to extract variant from `BpkTestVariant`. However: 
  Trade-offs:
  - Pro: Single source of truth (the Kotlin enum)
  - Pro: Automatically picks up new variants
  - Con: Adds complexity and another job dependency
  - Con: Harder to debug when something goes wrong
  - Con: The merge step (line 290) would also need to be updated to be dynamic
  - Con: Less explicit - harder to see what variants run at a glance

  My recommendation:
  Keep it hardcoded for now. The variants rarely change (4 variants in the entire project history), and the explicit list makes the workflow easier to understand and debug. The maintenance burden of keeping two lists in
  sync is minimal compared to the complexity of parsing Kotlin in bash.

LokmaneKrizou (approved):

PR #2740, opened 2026-07-08
[Android] Add BpkVideoPlayer composable

antonicg:
## Summary

- Adds `BpkVideoPlayer`, `BpkVideoPlayerController`, `BpkVideoPlayerConfig`, `BpkVideoPlaybackState`, `BpkVideoPlayerDefaultControls`, and `BpkVideoPlayerError` to `backpack-compose`
- Pure Jetpack Compose implementation built on Media3 ExoPlayer + `ContentFrame` (no `AndroidView`)
- `startsMuted = true` default so playback never interrupts background audio (iOS `.ambient` audio session equivalent)
- Reduced motion support: checks `ANIMATOR_DURATION_SCALE` and `TRANSITION_ANIMATION_SCALE` via `Settings.Global`, rechecked on `ON_RESUME` so a setting change while backgrounded is picked up when the app returns to the foreground
- App-level load timeout (default 7s) — ExoPlayer has no built-in timeout for a load that never errors, only hangs
- `scaleToFill` param controls `ContentScale.Crop` vs `ContentScale.Fit` on Media3's `ContentFrame`

### Component capabilities
- Scale-to-fill, freeform sizing via `Modifier` — no enforced aspect ratio
- `autoPlay`, `loop`, `resetToStart()` for carousel use cases
- `toggle()`, `setMuted()` for consumer-driven controls
- `BpkVideoPlayerDefaultControls` pre-built play/pause overlay for common cases (hidden while loading, matching iOS parity)
- Shareable `BpkVideoPlayerController` for continuous playback across card→fullscreen transitions
- `BpkVideoPlayerError` exposes `LoadTimeout` and `PlaybackFailed(cause: PlaybackException)` — matches iOS's two-path failure model (no `InvalidUrl`/`Interrupted`, since iOS has no equivalent)

### API
```kotlin
// Default controls
BpkVideoPlayer(controller = rememberBpkVideoPlayerController(config))

// Custom overlay
Box {
    BpkVideoPlayer(controller = controller, modifier = Modifier.fillMaxSize())
    MyPlayPause(controller = controller)
}

// Shared controller for continuous playback
val controller = rememberBpkVideoPlayerController(config)
BpkVideoPlayer(controller = controller) // card
// … pass same controller to fullscreen
BpkVideoPlayer(controller = controller) // fullscreen — playback continues uninterrupted
```

### Demo app
3 use cases mirroring [backpack-ios#2606](https://github.com/Skyscanner/backpack-ios/pull/2606):
- **Default Controls** — 16:9 card with built-in play/pause overlay
- **Continuous Playback** — card tap opens fullscreen using the same controller; position and mute state preserved
- **Fullscreen** — standalone fullscreen player

https://github.com/user-attachments/assets/78eae054-6938-4f6c-a783-a69e6e110663

See `docs/compose/VideoPlayer/README.md` for full usage docs including error handling.

## Test plan

- [ ] Run `backpack-compose:testDebugUnitTest` — `BpkVideoPlaybackStateTest` passes
- [ ] Run `BpkVideoPlayerTest` instrumentation tests — semantics, mute toggle, error guard pass
- [ ] Demo app → Video Player → verify all 3 stories render and play correctly
- [ ] Verify `BpkVideoPlayerDefaultControls` hides the button while loading, switches to pause when playing
- [ ] Verify Continuous Playback: tap card → fullscreen opens, video continues from same position
- [ ] Verify reduced motion: set Animator Duration Scale to 0 in developer options → video does not autoplay; background/foreground the app after toggling mid-session to confirm live update
- [ ] Verify no background music interruption when video plays

🤖 Generated with [Claude Code](https://claude.com/claude-code)


backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayer.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    LaunchedEffect(reducedMotion) {
+        if (reducedMotion && controller.playbackState.value.isPlaying) {
+            controller.pause()
+        }
+    }
+
+    Box(
+        modifier = modifier
+            .background(Color.Black)
```
henrik-sky, line 54: Why?

app/src/main/java/net/skyscanner/backpack/demo/compose/VideoPlayerStory.kt
```diff
@@ -0,0 +1,178 @@
...
+                controller = controller,
+                modifier = Modifier.fillMaxSize(),
+                scaleToFill = true,
+            )
+        }
+    } else {
+        Box(
+            modifier = modifier
+                .fillMaxWidth()
+                .aspectRatio(16f / 9f)
```
henrik-sky: Should it be `9f / 16f`?

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerController.kt
```diff
@@ -0,0 +1,152 @@
...
+import androidx.media3.exoplayer.ExoPlayer
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import net.skyscanner.backpack.compose.videoplayer.internal.PlayerFactory
+import net.skyscanner.backpack.compose.videoplayer.internal.isReducedMotionEnabled
+
+@Stable
+class BpkVideoPlayerController internal constructor(
```
henrik-sky, line 42: I think this should maybe be in the `internal` folder but keep `rememberBpkVideoPlayerController` in the main bit

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerDefaultControls.kt
```diff
@@ -0,0 +1,106 @@
...
+        modifier = modifier,
+    )
+}
+
+@Composable
+private fun VideoPlayerControlButton(
+    isPlaying: Boolean,
+    modifier: Modifier = Modifier,
+) {
+    Box(
```
henrik-sky: The design has a `BpkButton`. If you find yourself using `.copy()` something has generally gone wrong. I will caveat that the `onClick` for the button makes it a bit more complicated but you should be able to pass it along
```
   BpkButton(
        icon = if (isPlaying) BpkIcon.Pause else BpkIcon.Play,
        contentDescription = "",
        modifier,
        type = BpkButtonType.SecondaryOnDark,
        size = BpkButtonSize.Default,
    ) {}
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerController.kt
antonicg: Not sure, because the constructor is already internal but the class should be public since it is something the client needs 🤔

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerDefaultControls.kt
antonicg: Is this comment outdated? Not sure to understand, sorry

app/src/main/java/net/skyscanner/backpack/demo/compose/VideoPlayerStory.kt
antonicg: Since the video we have is in portrait for demo, this is the proper aspect ratio.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayer.kt
antonicg: Just to display a black background while loading. But the client has access to the events api, so probably the video won't be shown until is loaded and an image or any loading state, can be shown from the client instead.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerController.kt
```diff
@@ -0,0 +1,152 @@
...
+                    timeoutJob?.cancel()
+                    if (_playbackState.value !is BpkVideoPlaybackState.Playing) {
+                        _playbackState.value = BpkVideoPlaybackState.ReadyToPlay
+                    }
+                }
+                Player.STATE_BUFFERING -> {
+                    if (_playbackState.value is BpkVideoPlaybackState.Playing) {
+                        _playbackState.value = BpkVideoPlaybackState.Buffering
+                    }
+                }
```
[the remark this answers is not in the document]
antonicg: Fixed in [840a8c6](https://github.com/Skyscanner/backpack-android/pull/2740/commits/840a8c6622a5f7348ebe300508e93f4dadd9367d)

docs/compose/VideoPlayer/README.md
```diff
@@ -0,0 +1,177 @@
...
+The player defaults to `startsMuted = true`. Volume can be toggled at any time:
+
+```kotlin
+controller.setMuted(false) // unmute
+controller.setMuted(true)  // mute
+```
+
+### Accessibility
+
+- Reduced motion: autoplay is blocked and playback pauses when `ANIMATOR_DURATION_SCALE` is set to 0 or `AccessibilityManager.isAnimationsEnabled` returns false (API 33+).
```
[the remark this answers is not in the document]
antonicg: Fixed in [840a8c6](https://github.com/Skyscanner/backpack-android/pull/2740/commits/840a8c6622a5f7348ebe300508e93f4dadd9367d)

backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+class BpkVideoPlayerTest {
+
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    private val stubConfig = BpkVideoPlayerConfig(
+        videoUrl = "https://example.com/stub.mp4",
+        accessibilityLabel = "Test video",
+    )
```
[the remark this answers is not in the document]
antonicg: - `accessibilityLabel_isAppliedToSemantics`, `initialState_isLoading`,` setMuted_...` — these read state immediately or interact with local controller state only. The network URL is never actually used. 
- `play_whenFailed_doesNotCrash` — this one depends on reaching Failed state. It uses the timeout of 100 ms. In practice: `example.com/stub.mp4` will either timeout in 100ms, get a DNS error on CI with no internet, or return a 404/non-video response. All of which produce Failed. The test is reliable.

app/src/main/java/net/skyscanner/backpack/demo/compose/VideoPlayerStory.kt
henrik-sky: image

Looks like it should be the other way around to me

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerController.kt
```diff
@@ -0,0 +1,152 @@
...
+        override fun onPlaybackStateChanged(playbackState: Int) {
+            when (playbackState) {
+                Player.STATE_READY -> {
+                    timeoutJob?.cancel()
+                    _playbackState.value =
+                        if (player.isPlaying) BpkVideoPlaybackState.Playing
+                        else BpkVideoPlaybackState.ReadyToPlay
+                }
+                Player.STATE_BUFFERING -> {
+                    if (_playbackState.value is BpkVideoPlaybackState.Playing) {
```
henrik-sky: **Bug: `STATE_BUFFERING` guard misses `ReadyToPlay`**

If `autoPlay = false` and the user calls `play()` from `ReadyToPlay`, ExoPlayer can fire `STATE_BUFFERING` before `onIsPlayingChanged(true)`. The guard only matches `Playing`, so state stays `ReadyToPlay` through the whole buffer window — `isLoading` returns `false` and `BpkVideoPlayerDefaultControls` shows the play button while actually buffering.

```kotlin
Player.STATE_BUFFERING -> {
    if (_playbackState.value is BpkVideoPlaybackState.Playing ||
        _playbackState.value is BpkVideoPlaybackState.ReadyToPlay) {
        _playbackState.value = BpkVideoPlaybackState.Buffering
    }
}
```

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerError.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package net.skyscanner.backpack.compose.videoplayer
+
+import androidx.media3.common.PlaybackException
+
+sealed class BpkVideoPlayerError {
+    data object LoadTimeout : BpkVideoPlayerError()
+    data class PlaybackFailed(val cause: PlaybackException) : BpkVideoPlayerError()
+}
```
henrik-sky, line 24: **API stability: `PlaybackException` leaks into public surface**

`PlaybackException` is from `androidx.media3.common` (a `@UnstableApi` type). Consumers who match on `PlaybackFailed` or read `cause` take a direct Media3 dependency into their call sites. If the backend ever changes, removing this field is a breaking API change.

Consider typing `cause` as `Exception` instead — callers that need the Media3 details can cast, and the signature stays stable regardless of the underlying player library.

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerController.kt
```diff
@@ -0,0 +1,152 @@
...
+    private var timeoutJob: Job? = null
+
+    init {
+        player.repeatMode = if (config.loop) Player.REPEAT_MODE_ONE else Player.REPEAT_MODE_OFF
+        player.volume = if (config.startsMuted) 0f else 1f
+        player.playWhenReady = config.autoPlay && !(config.respectsReducedMotion && isReducedMotionEnabled(context))
+        player.addListener(playerListener())
+        player.setMediaItem(MediaItem.fromUri(config.videoUrl))
+        player.prepare()
+        _playbackState.value = BpkVideoPlaybackState.Loading
```
henrik-sky, line 64: **Testability: `PlayerFactory` is not injectable**

Being a sealed `internal object`, `PlayerFactory` cannot be replaced in tests. The instrumentation tests work around it via a 100ms load-timeout to reach `Failed`, but `Playing`/`Paused` states are unreachable without a real network call. If future tests need to exercise playback state transitions, this will be a blocker.

Not a blocker for this PR, but worth noting if unit-testable controller logic becomes a goal.

henrik-sky:
Two minor nits:

**`Idle` state is unreachable** — `_playbackState` is initialised to `Idle` but immediately overwritten to `Loading` in `init`. No observer will ever see it. Either remove the state or initialise directly to `Loading`. The README state table documents it as visible.

**`resetToStart()` does not clear `Ended` state** — `player.seekTo(0)` is issued but the state stays `Ended` until playback events fire. A caller that checks `playbackState` immediately after `resetToStart()` will observe `Ended`, which is surprising. Consider setting `_playbackState.value = BpkVideoPlaybackState.Paused` (or `ReadyToPlay`) synchronously alongside the seek.

app/src/main/java/net/skyscanner/backpack/demo/compose/VideoPlayerStory.kt
antonicg: Ah, the Box you mean. I can change that

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayer.kt
```diff
@@ -0,0 +1,63 @@
...
+    modifier: Modifier = Modifier,
+    scaleToFill: Boolean = false,
+) {
+    val reducedMotion by rememberReducedMotionEnabled()
+
+    LaunchedEffect(reducedMotion) {
+        if (reducedMotion && controller.playbackState.value.isPlaying) {
+            controller.pause()
+        }
+    }
```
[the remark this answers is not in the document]
antonicg: Implemented

app/src/main/java/net/skyscanner/backpack/demo/compose/VideoPlayerStory.kt
antonicg: Done in a34545b8b

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerController.kt
antonicg: Sure, done in a34545b8b

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/videoplayer/BpkVideoPlayerError.kt
antonicg: Sure, changed by a regular exception and then the client can cast to the playback one if they consider necessary 👌  a34545b8b

henrik-sky (approved):
Thanks for your contribution!

PR #2772, opened 2026-07-31
Add publish-backpack-local skill

henrik-sky:
## Summary
- Adds a Claude Code skill that automates publishing backpack-android to Maven local as a SNAPSHOT and wiring up skyscanner-app to consume it, for testing unreleased changes before a real release
- Derives the SNAPSHOT version from the latest git tag (bumping patch), publishes via `./gradlew publishToMavenLocal`, then updates skyscanner-app's version catalog and adds `mavenLocal()` to its repository resolution
- Repo paths are derived at runtime (sibling directory lookup) rather than hardcoded, and version-tag filtering handles both the legacy `Backpack@X.Y.Z` and current unprefixed `X.Y.Z` tag schemes

## Test plan
- [x] Ran the skill end-to-end locally: derived version, published to Maven local, verified both files in skyscanner-app were updated correctly
- [ ] Reviewer sanity-check of SKILL.md instructions


.claude/skills/publish-backpack-local/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+allowed-tools: Bash(git rev-parse*), Bash(git fetch*), Bash(git tag*), Bash(./gradlew *), Bash(find *), Bash(test *), Read, Edit
+---
+
+# Publish Backpack to Maven Local
+
+Round-trips local backpack-android changes into the skyscanner-app for manual testing, without cutting a real release.
+
+## 0) Locate the two repos
+
+Assume `backpack-android` and `skyscanner-app` are sibling directories (same parent folder). Derive both paths from the current working directory — do not hardcode a user-specific path.
```
LokmaneKrizou: instead of assuming, its worth actually checking first and stop and ask when in doubt ( avoid eating tokens on trivial tasks)

.claude/skills/publish-backpack-local/SKILL.md
```diff
@@ -0,0 +1,95 @@
+---
+name: publish-backpack-local
+description: Publish backpack-android to Maven local as a SNAPSHOT and wire skyscanner-app to consume it. Use when the user wants to test unreleased backpack-android changes inside the skyscanner-app before a real release.
+allowed-tools: Bash(git rev-parse*), Bash(git fetch*), Bash(git tag*), Bash(./gradlew *), Bash(find *), Bash(test *), Read, Edit
+---
```
LokmaneKrizou: haven't run the skill to check yet, but if it does what it promised every time, then worth switching it to an agent instead and set model to haiku to save cost instead of allowing agents with expensive models to run simple task.

.claude/skills/publish-backpack-local/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+BACKPACK_DIR=$(git rev-parse --show-toplevel)
+APP_DIR="$(dirname "$BACKPACK_DIR")/skyscanner-app"
+test -d "$APP_DIR" || echo "NOT FOUND: $APP_DIR"
+```
+
+If invoked from somewhere other than inside `backpack-android` (e.g. `pwd` isn't that repo), find it by name under the common parent instead, then apply the same sibling logic. If `skyscanner-app` still can't be located, stop and ask the user for its path rather than guessing.
+
+Use `$BACKPACK_DIR` and `$APP_DIR` in place of any absolute path in the steps below.
+
+## 1) Derive the SNAPSHOT version
```
LokmaneKrizou: this step could be converted to a script, it would be more deterministic, link the script here and allow agent/skill to offer instructions of what to do when script fails to find it, either by allowing AI to investigate and get the version, when still in doubt ask user to provide the version

.claude/skills/publish-backpack-local/SKILL.md
henrik-sky: Done

.claude/skills/publish-backpack-local/SKILL.md
henrik-sky: Hardcoded version so we don't need a script or too wide bash permissions

.claude/skills/publish-backpack-local/SKILL.md
LokmaneKrizou: The skill's entire body is "delegate to the agent." It adds no procedure, and duplicates the agent's description almost word-for-word. On the face of it, dead weight — the agent's own description already makes it auto-delegatable from natural language.

@henrik-sky , want me to remove it?

.claude/agents/publish-backpack-local.md
```diff
@@ -0,0 +1,73 @@
...
+
+If `skyscanner-app` can't be located, stop and ask the user for its path.
+
+Use `$BACKPACK_DIR` and `$APP_DIR` in place of any absolute path in the steps below.
+
+## 1) Publish to Maven local
+
+```bash
+cd "$BACKPACK_DIR"
+./gradlew publishToMavenLocal -Pversion="99.9.9.SNAPSHOT"
```
LokmaneKrizou: The agent works as expected — published all three artifacts, POMs correctly
cross-reference backpack-common at the same version, and skyscanner-app was
wired up correctly (mavenLocal() went into dependencyResolutionManagement,
not pluginManagement).
Two questions on the version:
1. Is 99.9.9.SNAPSHOT fixed, or can a version be passed in? Looks hardcoded —
   and I see 158f04b20 deliberately replaced the tag-derived version with it.
   No objection to hardcoding, it's simpler and can't drift between step 1
   and step 2a.
2. Should it be 99.9.9-SNAPSHOT (hyphen) rather than 99.9.9.SNAPSHOT (dot)?
   Maven/Gradle detect snapshots by the literal "-SNAPSHOT" suffix, so with a
   dot this resolves as a static release version. Republishing does overwrite
   ~/.m2 correctly (confirmed), but the app's Gradle cache has no reason to
   re-read a static version — so on the second iteration you could be testing
   stale code with a green build. Every other artifact in ~/.m2 uses the
   hyphen (83.0.0-SNAPSHOT, 84.0.0-SNAPSHOT).
   Related: step 3 says "sync/rebuild" — worth suggesting
   --refresh-dependencies, since Gradle caches changing modules for 24h by
   default even with a proper -SNAPSHOT.

.claude/agents/publish-backpack-local.md
```diff
@@ -0,0 +1,73 @@
+---
+name: publish-backpack-local
+description: Publishes backpack-android to Maven local as a fixed SNAPSHOT version and wires skyscanner-app to consume it. Use when the user wants to test unreleased backpack-android changes inside skyscanner-app before a real release.
+tools: Bash, Read, Edit
+model: haiku
+---
+
+Round-trips local backpack-android changes into the skyscanner-app for manual testing, without cutting a real release.
+
+Uses the fixed version `99.9.9-SNAPSHOT` — high enough to never collide with real releases (currently ~83.x.x). The `-SNAPSHOT` suffix is required so Gradle treats it as a changing module and re-resolves on each build.
```
LokmaneKrizou: claude check this line, specifically this part (currently ~83.x.x). we should not have this hardcoded here or state something that would change in the future.

.claude/agents/publish-backpack-local.md
```diff
@@ -0,0 +1,73 @@
...
+
+Do NOT add `mavenLocal()` to the `pluginManagement` repositories block — only the `dependencyResolutionManagement` one is needed.
+
+## 3) Report
+
+Confirm to the user:
+- That `publishToMavenLocal` succeeded with version `99.9.9-SNAPSHOT`.
+- That both files in skyscanner-app were updated (or that `mavenLocal()` was already present and skipped).
+
+Remind them to sync/rebuild skyscanner-app next, ideally with `--refresh-dependencies` since Gradle caches changing modules for 24h by default.
```
LokmaneKrizou: claude: instead of reminding, you should run that urself  as part of the  Point skyscanner-app at the snapshot steps 
report should confirm if synced and worked correctly or failed with clear reasons to help debug.

.claude/agents/publish-backpack-local.md
```diff
@@ -0,0 +1,73 @@
+---
+name: publish-backpack-local
```
LokmaneKrizou: claude update  name to `backpack-local-publisher`

LokmaneKrizou (approved):

PR #2785, opened 2026-08-13
[DON-3492] Add accessibility label to BPKBannerAlert

jaspersion05:
## Summary

Adds `accessibilityLabel` support for the custom icon in `BpkBannerAlert`, closing the gap identified by the parity audit — previously there was no way to set a custom content description for the icon separate from the banner's `alertTypeContentDescription`.

## Changes

- Added a new optional trailing parameter `iconContentDescription: String? = null` to `BpkBannerAlert` and `BpkBannerAlertImpl`, threaded through to the icon's `contentDescription`.
- The icon's content description now resolves as `iconContentDescription ?: alertTypeContentDescription`, preserving existing behaviour when the new parameter isn't supplied.
- Documented the new parameter in `docs/compose/BannerAlert/README.md` with a usage example showing a custom icon paired with its own accessibility label.
- Added `BpkBannerAlertAccessibilityTest.kt` covering both the explicit `iconContentDescription` case and the fallback-to-`alertTypeContentDescription` case, following the existing `BpkCellItemAccessibilityTest.kt` convention.

## Compatibility

Source- and binary-compatible. The new parameter is appended last and defaults to `null`. All existing call sites in this repo and in `skyscanner-app` (Origami, Wasabi) use named arguments only, so none are affected.

## Verification

- New accessibility tests added and pass locally.
- No visual/behavioural change for existing callers, so no snapshot re-recording was needed.

+ [x] `README.md`
+ [x] Tests
+ [ ] [Screenshotting code](https://github.com/Skyscanner/backpack-android/blob/main/app/src/main/java/net/skyscanner/backpack/demo)
+ [ ] Adding a component? Remember to expose it in the main header/index if applicable

### DIFF
--- a/backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/banneralert/BpkBannerAlertAccessibilityTest.kt
+++ b/backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/banneralert/BpkBannerAlertAccessibilityTest.kt
@@ -1,0 +1,72 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 - 2026 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


henrik-sky (approved):

peterInTown (changes requested):
why do we need two different param alertTypeContentDescription & iconContentDescription control the same component?

jaspersion05:
> why do we need two different param alertTypeContentDescription & iconContentDescription control the same component?

alertTypeContentDescription describes the banner's type (info/warning/etc.) and is the default label used for the icon. iconContentDescription only kicks in when you pass a custom icon that isn't just illustrating the alert type e.g. the Airline icon example in the docs, which should announce "Airline" rather than "Information". Without a separate param, custom icons would always fall back to the (often wrong) type label.

PR #2796, opened 2026-08-24
fix: center content in full-width BpkChip

mldtms:
## Summary

  Centers selectable `BpkChip` content when callers stretch the chip beyond its intrinsic width (for example, `Modifier.fillMaxWidth()` in equal-width filter rows).

  `BpkDropdownChip` and `BpkDismissibleChip` retain their existing layout behavior.

  ## Testing

  - `./gradlew :app:testOssDebugUnitTest --tests 'net.skyscanner.backpack.compose.chip.BpkChipTest' -Dhttps.protocols=TLSv1.2`

  + [ ] Component `README.md` (not needed; no public API or documentation change)
  + [x] Tests


app/screenshots/oss/debug/default/net.skyscanner.backpack.compose.chip.BpkChipTest_typeDismissFullWidth.Default_VDL2.png
LokmaneKrizou, line 1: Is this something expected to have, should there be a space between the text and the trailing icon ? or should the content centred

app/src/test/java/net/skyscanner/backpack/compose/chip/BpkChipTest.kt
```diff
@@ -78,6 +81,23 @@ class BpkChipTest(flavour: Flavor) : BpkSnapshotTest(listOfNotNull(flavour.first
         }
     }
 
+    @Test
+    fun fullWidth() = snap(width = 200.dp) {
```
LokmaneKrizou: nit: could we extract this fixed width to its own global variable

backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/chip/internal/BpkChipImpl.kt
```diff
@@ -165,7 +165,11 @@ internal fun BpkChipImpl(
 
     Row(
         verticalAlignment = Alignment.CenterVertically,
-        horizontalArrangement = Arrangement.spacedBy(BpkSpacing.Md),
+        horizontalArrangement = if (type == BpkChipType.Selectable) {
```
LokmaneKrizou, line 168: Question: why is this applied only to the Selectable type, rather than also to the rest of types (Dismiss and DropDown types)? From the images above, it looks like the same issue affects those options too on wider widths.

app/screenshots/oss/debug/default/net.skyscanner.backpack.compose.chip.BpkChipTest_typeDismissFullWidth.Default_VDL2.png
mldtms: It's default behaviour. It was like that and its in production. No code change was introduced to this behaviour. I would not change it unless we have clear request from design.

app/src/test/java/net/skyscanner/backpack/compose/chip/BpkChipTest.kt
mldtms: Addressed.

app/src/test/java/net/skyscanner/backpack/compose/chip/BpkChipTest.kt
```diff
@@ -78,6 +81,23 @@ class BpkChipTest(flavour: Flavor) : BpkSnapshotTest(listOfNotNull(flavour.first
...
+    }
+
+    @Test
+    fun fullWidthWithIcon() = snap(width = CHIP_FULL_WIDTH.dp) {
+        BpkChip(
+            text = "Chip",
+            icon = BpkIcon.Deals,
+            modifier = Modifier.fillMaxWidth(),
+        )
+    }
```
andri-ivanovS, line 99: Add a longer chip text test:
@Test
fun fullWidthWithLongText() = snap(width = CHIP_FULL_WIDTH) {
    BpkChip(
        text = "A chip label at least long enough to ellipsize",
        modifier = Modifier.fillMaxWidth(),
    )
}

andri-ivanovS (approved):

PR #2809, opened 2026-09-02
[DON-2175] Fix stale bottom sheet content

henrik-sky:
Resolves DON-2175

## Context
`BpkBottomSheet` and `BpkModalBottomSheet` move caller-provided content with Compose `movableContentOf`. The movable content was previously created inside an unkeyed `remember { ... }`, so the first content lambda could remain cached for the lifetime of the composition. If a sheet stayed composed while the caller supplied different content, the sheet could continue displaying stale UI. Passing a wrapper such as `sheetContent = { content() }` could hide the issue by changing the lambda identity, but passing the content lambda directly should also update correctly.

## Change
- Key the remembered movable content with `content` in both the persistent and modal bottom-sheet implementations.
- Compose now recreates the movable content slot when the supplied content lambda changes, while retaining the existing movable-content behaviour when it remains stable.
- No public API changes are required.

## Regression coverage
Added instrumentation tests for both paths:
- persistent bottom sheet content changes from the old value to the new value while the sheet remains open
- modal bottom sheet content changes with a header present

Each test verifies that the new content is rendered and the stale content is no longer present.

## Verification
- Android Studio MCP project build passed
- Android Studio MCP instrumentation tests passed on the connected device
- detekt commit hook passed

### DIFF
--- a/backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/bottomsheet/BottomSheetContentTest.kt
+++ b/backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/bottomsheet/BottomSheetContentTest.kt
@@ -1,0 +1,85 @@
+/**
+ * Backpack for Android - Skyscanner's Design System
+ *
+ * Copyright 2018 - 2026 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


backpack-compose/src/androidTest/kotlin/net/skyscanner/backpack/compose/bottomsheet/BottomSheetContentTest.kt
```diff
@@ -0,0 +1,85 @@
...
+    @Test
+    fun bottomSheet_updatesWhenContentLambdaChanges() {
+        val showFirstContent = mutableStateOf(true)
+
+        composeTestRule.setContent {
+            BpkTheme {
+                BpkBottomSheet(
+                    state = rememberBpkBottomSheetState(BpkBottomSheetValue.Expanded),
+                    sheetContent = if (showFirstContent.value) {
+                        { BpkText(text = "First content") }
```
LokmaneKrizou: nit: extract these hardcoded texts to global variables

LokmaneKrizou (approved):

PR #2820, opened 2026-09-14
[DON-3826] Fully suppress TalkBack state announcement on BpkSwitch/BpkSaveButton

andri-ivanovS:
## Pull Request Summary
_Relates to [DON-3826](https://skyscanner.atlassian.net/browse/DON-3826)_

Follow-up to the `announceState` option added for `BpkSwitch`/`BpkSaveButton` (`BpkCardButton`): TalkBack was still announcing the default "checked"/"not checked" state even when `announceState = false`, because an empty `stateDescription` is treated by TalkBack as unset and it falls back to synthesizing its own announcement from the node's checkable/checked flags. Setting a zero-width space instead of an empty string gives TalkBack a non-empty value that renders as nothing, which fully suppresses the announcement.

## Changes Made
- `Modifier.suppressToggleableStateAnnouncement` in `backpack-compose/.../utils/Modifier.kt`: changed `stateDescription = ""` to `stateDescription = "​"`. This modifier is shared by `BpkSwitch` (interactive) and `BpkCardButton`/`BpkSaveButton`.
- `BpkSwitch.kt`: the read-only branch (`onCheckedChange == null`), which sets `stateDescription` directly rather than through the shared modifier, was updated the same way for consistency.

## Tests
Added semantics assertion tests (non-screenshot) that assert on `SemanticsProperties.StateDescription` directly, covering both fix sites:
- `BpkSwitchTest.announceStateFalseSuppressesStateDescription` — interactive switch (`onCheckedChange` set).
- `BpkSwitchTest.announceStateFalseSuppressesStateDescriptionWhenReadOnly` — read-only switch (`onCheckedChange = null`).
- `BpkCardButtonTest.announceStateFalseSuppressesStateDescription` — `BpkSaveButton`, run across all size/style flavours.

I verified these tests actually catch the regression: reverted the fix locally, confirmed all 8 (parameterized) cases failed, then restored the fix and confirmed they pass.

Also ran the full existing Roborazzi suite in verify mode (`BpkSwitchTest` 11/11, `BpkCardButtonTest` 24/24) against the committed reference screenshots — no visual diffs, confirming the fix has no rendering side effects.

## Additional Notes
Root cause confirmed by reading AndroidX Compose's `AndroidComposeViewAccessibilityDelegateCompat`: `info.isCheckable`/`info.isChecked` are set independently of `stateDescription` whenever `ToggleableState` is present, and TalkBack (not Compose) synthesizes "checked"/"not checked" from those flags when it sees an empty `stateDescription`. Compose's own source acknowledges this as a TalkBack quirk. The zero-width space is the standard mitigation for this specific platform behaviour, not a workaround unique to this codebase.

[DON-3826]: https://skyscanner.atlassian.net/browse/DON-3826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henrik-sky (approved):
I have my reservations but it does work

PR #2827, opened 2026-09-17
Fix duplicate keyboard focus stop on BpkSwitch and BpkCheckbox

mldtms:
Description:
## What

External keyboard Tab navigation stopped twice on the same `BpkSwitch`
(and, as it turns out, `BpkCheckbox`) instead of once.

**Root cause:** Material3's `Switch`/`TriStateCheckbox` attach their own
`toggleable()`/`triStateToggleable()` (and thus their own focus target)
whenever `onCheckedChange`/`onClick` is non-null. `BpkSwitch`/`BpkCheckbox`
already wrap their content in a `Row` with its own `toggleable()`/
`triStateToggleable()`, so the control ended up with two overlapping
focusable regions — one Tab stop too many.

**Fix:** stop forwarding `onCheckedChange`/`onClick` into the inner
Material3 composable. Toggling is still driven entirely by the outer
`Row`; `interactionSource` stays shared so the thumb/checkbox press
visuals keep animating.

`BpkCheckbox` wasn't part of the original report, but the same pattern
was confirmed in Material3's source (`TriStateCheckbox` behaves
identically to `Switch` here), so it's fixed in this PR too rather than
filing a separate ticket.

## Changes

- `BpkSwitchImpl` / `BpkCheckboxImpl`: no longer forward
  `onCheckedChange` / `onClick` to the inner Material3 component.
- Added `BpkSwitchAccessibilityTest` / `BpkCheckboxAccessibilityTest`
  (instrumented) asserting a single Tab stop and that Enter still
  toggles.
- Added a switch showcase example with an info icon in the label
  (`SwitchWithIconExample`), mirroring the existing checkbox one.
- `Switch`/`Checkbox` README: documented that the component is a single
  Tab stop for external keyboards.

## Testing

- New instrumented tests pass for both components; verified they
  actually catch the regression by reverting the fix locally and
  confirming the test fails as expected.
- Manually confirmed on an emulator with TalkBack's focus highlight
  tracking hardware Tab: one rectangle per switch/checkbox row, moving
  cleanly to the next row on each Tab press.

+ [x] Component `README.md`
+ [x] Tests


henrik-sky (approved):
Nice one!

